PR #7, opened 2016-05-06
Codegen must die

### LINKED ISSUE #4: Eliminate code-gen for the forallN guts

trws:
Basically what it says on the tin.  This PR contains a rework of all remaining codegen-based segments of nested-loop raja to use template metaprogramming instead.  Initial results show that, as before, this version compiles slightly slower and runs slightly faster than the generated version, presumably because more of it is guaranteed at compile time.  Cleanup still remains in a few places, but functionally this is good to go.

review @ajkunen?

fixes #4


include/RAJA/LegacyCompatibility.hxx
```diff
@@ -0,0 +1,215 @@
...
+
+static_assert(__cplusplus >= 201103L, "C++ standards below 2011 are not supported");
+
+#include <type_traits>
+#include <utility>
+#include <iostream>
+#include <functional>
+#include <cstdint>
+
+namespace VarOps {
```
ajkunen: Do we want this namespace to be RAJA::VarOps?

include/RAJA/forallN_generic.hxx
```diff
@@ -46,153 +46,65 @@
 #ifndef RAJA_forallN_generic_HXX__
 #define RAJA_forallN_generic_HXX__
 
+#include <RAJA/LegacyCompatibility.hxx>
 #include "forallN_generic_lf.hxx"
```
ajkunen: The reason there are XXX.hxx and XXX_lf.hxx files was because the XXX.hxx was code-gen'ed, and the XXX_lf.hxx wasn't (at some point lf stoof for "lambda-free", but that doesn't really matter).

Do you want to merge forallN_generic_lf.hxx into this file?
Same for forallN_permute_lf.hxx

include/RAJA/forallN_generic.hxx
```diff
@@ -46,153 +46,65 @@
...
- *
- * Provides index typing, and initial nested policy unwrapping
- */
-template<typename POLICY, typename IdxI=Index_type, typename TI, typename BODY>
+template<typename POLICY, typename ...Indices, typename...ExecPolicies, typename BODY, typename... Ts>
 RAJA_INLINE
-void forallN(TI const &is_i, BODY const &body){
+void forallN_impl_extract(RAJA::ExecList<ExecPolicies...>, BODY const &body, const Ts ... args){
+  static_assert(sizeof...(ExecPolicies) == sizeof...(args),
+                "The number of execution policies and arguments does not match");
```
ajkunen: awesome!

include/RAJA/LegacyCompatibility.hxx
tgamblin: On this note: John Mellor-Crummey was asking whether we could somehow tag the top-level entry points for RAJA so he can do better collapsing of nested/inlined calls in HPCToolkit.  Basically he wants to show users a profile that only includes, e.g. RAJA:foraall but not all the plumbing underneath. 

I think you could do this by putting all the app entry points in the RAJA namespace and moving internal calls out to other namespaces. Then tools would have nice labels and could pick what to show. Thoughts?

ajkunen:
Also, could you remove the /gen directory?  That is totally unnecessary with these changes!

If you make the changes I mentioned, I whole heatedly approve!

include/RAJA/LegacyCompatibility.hxx
davidbeckingsale: We could move the internals to RAJA::impl or something like that. We should run it past @rhornung67 and @keasler

ajkunen: @davidbeckingsale I like that idea.  We should have a RAJA:impl where we can hide all of the internal guts that the user isn't supposed to call directly

rhornung67: Todd,

Without thinking too hard, I believe that would work.

—Rich

rhornung67:
“whole-heatedly”?!?  What’s wrong with you man?

include/RAJA/LegacyCompatibility.hxx
rhornung67: Works for me.  —Rich

trws: I'm good with that.  I should really re-arrange some of that anyway since it's now both legacy compatibility code and variadic helper classes/functions in one place.

include/RAJA/forallN_generic.hxx
trws: Definitely.  This is some of the cleanup I mentioned in one of the comments.

ajkunen:
Looks like it doesn't build with Intel 17

trws:
Or some nvccs as it turns out, running more complete tests and cleaning some of that up.

include/RAJA/LegacyCompatibility.hxx
tgamblin: Minor nitpick: `RAJA::internal` isn't very descriptive.  But for internal namespaces you've got some leeway because you can change them later.

davidbeckingsale: How about `RAJA::magic`?

rhornung67: I thought we were going with RAJA::impl?

davidbeckingsale: We are, I was being facetious.

DavidPoliakoff: I feel that impl has a certain association in c++11, I think it's being used correctly here, so I'd just use it, personally. Though if we're going with facetious suggestions, I suggest raja::openmp::kokkos::cuda::openacc::chapel::agency::fiddling_with_magnets

rhornung67: I would further suggest for clarity that we remove all the vowels from our namespace labels…

DavidPoliakoff: snds gd t m, mrgng

ajkunen: I agree, that might speed up compile times

keasler:
Is this pull request still alive?

trws:
Yes, but not done.  It needs a class to replace the arrays in the layout with something that will be copied to the device when copied by  value.


trws:
Ok, this should now pass for all ForallN constructs.  It may warrant some further re-organization or cleanup, but it seems to be functional.  @ajkunen, any chance you could beat on it with some kripke cuda policies I might not have thought of?

include/RAJA/foralln/Layout.hxx
```diff
@@ -0,0 +1,138 @@
...
+  typedef IdxLin IndexLinear;
+  typedef std::tuple<IDs...> IDtuple;
+  typedef VarOps::make_index_sequence<sizeof...(IDs)> IndexRange;
+
+  static constexpr size_t n_dims = sizeof...(IDs);
+  static constexpr size_t limit = std::numeric_limits<Index_type>::max();
+
+  // const char *index_types[sizeof...(IDs)];
+
+  Index_type perms[sizeof...(IDs)];
```
ajkunen: @trws I thought you were going to implement this without the arrays... Does this work with CUDA?

trws: I was, and I did, the result of that is in IndexArray.hxx.  After that I found it was still breaking, and that the problem was actually one of the ForallN wrapper functors capturing the lambda by reference.  After fixing the functor, it works with either, and this is more efficient for accesses to dynamic offsets.  Admittedly there aren't a lot of those, but why not.

In summary, yes it does work with CUDA.

keasler:
Is it possible to quantify the compile-time difference?  As we saw in the email from Rob this morning, compile-time for other frameworks may be substantially longer than pre-framework builds, and it would be beneficial if we can minimize that issue with RAJA.  Clearly if the compiler is broken, like the current Intel compiler, we can't necessarily eliminate the problem, but we should still shoot for minimizing the impact.

keasler:
@ajkunen Tom might be waiting for another approval from you for this branch since he fixed the issues you had previously raised.

DavidPoliakoff:
@keasler: comparing against develop, with its tests enabled, the times are within the noise of one another given a make -j on a node of rzhas using gcc (and clang as well).

The noise was high, times ranging from 8 seconds to 10, and with times that low we'd want a bigger compilation task to really test this, but we can't say one is more expensive than the other based on the data available. I might look at putting together a compiler stresser to make the compile time on the order of minutes for better benchmarking

trws:
It depends on how you look at it @keasler.  Each instantiation is probably longer, but because there's so much less code to parse it's overall slightly faster to compile in my tests so far.  If someone uses a precompiled header approach for RAJA, it will always be slower, but by how much is a question.

I would actually recommend we pull the ForallN include out of RAJA.hxx by default and have people include it explicitly for just this reason.  It will make all of the non-forallN code compile faster just by not having to parse all the foralln code.

keasler:
@DavidPoliakoff "I might look at putting together a compiler stresser to make the compile time on the order of minutes for better benchmarking"  If you can ever find time for this, it would be a great extra dimension of performance testing that frameworks don't usually have -- worthy of mentioning in all our testing write-ups.

keasler:
@trws I had previously added guards for the forallN stuff -- did I miss a few spots somewhere?

trws:
@keasler, it is guarded, but it can be included at CMake time, which would mean all including compilation units would get it as opposed to just those files that want it.

rhornung67:
The fact that gcc and clang are reasonably fast compiles without much variation across differently-coded versions of the abstractions isn’t really surprising, I think.  This needs to be checked for Intel as well, which is the compiler we use in production. Also, the tests and proxy examples in the RAJA repo are not really representative of what real apps will encounter.

rhornung67:
It looks like it’s (the forallN stuff) guarded in the RAJA include directories and is off by default in the configuration.  It gets turned on explicitly in the Kripke CMakeLists.txt file. @davidbeckingsale this will solved when the proxy apps are pulled out into their own repo, correct?

trws:
@rhornung67, icpc is hard to quantify, because so much of the implementation of kripke is guarded off to prevent ICC from either dying or generating incorrect code.  I ran some quick tests, with the current guards, they're within half a second of each other between 6.2 and 6.5 seconds.  That said, it's really not worth making the comparison until ICC can handle building at least one of them properly...

ajkunen:
I approve!

This doesn't seem to change any compiler support:  what was broken is still broken, and nothing new is broken... alas nothing has been fixed either :)

trws:
The branch should be clean again, thanks for the heads-up on that @ajkunen, it was just a whitespace conflict in one of the cmake files as it turns out.

ajkunen:
Hmm,  I approved this, but the approval requirement seem steep: do we really need 2 people, PLUS either Rich or Jeff to approve?

DavidPoliakoff:
Pinging @keasler and @rhornung67 on that question, @ajkunen , we've heard that we have both too much and too little vetting of PR's recently, we should probably make our policy clear to ourselves before trying to express it clearly to pullapprove :)

davidbeckingsale:
@ajkunen - It's 2 total, with one of the two being either Rich or Jeff.

rhornung67:
I’m fine with a one or two person approval requirement for most pull requests (not the PR submitter).  I think it makes more sense to have the PR submitter identify those who they want to scrutinize the PR; I.e., choose reviewers based on the scope of the PR.  I assume pullapprove supports that. For example, I think it’s wise to include David B. for reviews of Cmake-related things — Jeff and I don’t necessarily need to be involved with that. Having as many eyeballs looking at major design changes and API issues is best.

trws:
The requirement is currently as @davidbeckingsale says, two approvals at least one of which must be @rhornung67 or @keasler.  

PullApprove does allow for some flexibility on reviewing requirements based on labels or other features of a given commit.  For example, I could require all commits that modify a CMakeLists.txt file require @davidbeckingsale's approval, or have additional reviewing requirements added for a given label, or when targeting a specific branch.

All of that said, there has to be a baseline for non-special PRs.  There's no way for the person putting up the PR to select arbitrary reviewers other than to tie each reviewer to a label or some such, in which case it wouldn't require any reviews by default.  I'm guessing that's not what we want.

That said, if we can express what we do want in terms of:

1.  What *all* PRs require
2.  What special conditions require special review
  1.  What groups should review the special condition versions

I'll be happy to set it up.

Here's a more complex example from their docs that always requires approval from both admins, review from backend_devs when labeled or targeting staging branch, and requires a design review when modifying .jpg or .png files:

```yaml
approve_by_comment: true
approve_regex: '^:\+1:'
reject_regex: '^No'
reviewers:
    -
        name: admins
        required: 2
        members:
            - a_user
            - anotheruser
    -
        name: backend_devs
        required: 1
        members:
            - backender_one
        conditions:
            labels:
                - backend-review
            branches:
                - staging
    -
        name: designers
        required: 1
        teams:
            - design
        conditions:
            files:
                - "*.png"
                - "*.jpg"
```

PR #17, opened 2016-05-17
Add BoxSegment as a replacement for RangeSegment

davidbeckingsale:
This replaces #16, which contains the doubled commits.


include/RAJA/RAJA.hxx
```diff
@@ -87,6 +87,9 @@
 
 #include "reducers.hxx"
 
+#if defined(RAJA_USE_BOXSEGMENT)
+#include "BoxSegment.hxx"
+#endif
 #include "RangeSegment.hxx"
```
davidbeckingsale: I think that this line should be removed, since RangeSegment no longer exists.

davidbeckingsale:
@keasler Here is the new version of the PR. If you have additional changes to make, please checkout this branch locally and push more commits here.

PR #32, opened 2016-05-25
Allow negative indices in RangeSegment

davidbeckingsale:
Uses INT_MIN rather than -1 to represent out-of-bounds indices when
launching the CUDA forallN

### DIFF
--- a/include/RAJA/exec-cuda/forallN_cuda.hxx
+++ b/include/RAJA/exec-cuda/forallN_cuda.hxx
@@ -59,6 +59,8 @@
 #include<RAJA/config.hxx>
 #include<RAJA/int_datatypes.hxx>
 
+#include <climits>
+
 namespace RAJA {
 
 
@@ -136,7 +138,7 @@ struct CudaThreadBlock {
 
     int idx = begin + view(blockIdx) * threads_per_block + view(threadIdx);
     if(idx >= end){
-      idx = -1;
+      idx = INT_MIN;
     }
     return idx;
   }
@@ -194,7 +196,7 @@ struct CudaThread {
     
     int idx = begin + view(threadIdx);
     if(idx >= end){
-      idx = -1;
+      idx = INT_MIN;
     }
     return idx;
   }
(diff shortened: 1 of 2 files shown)


davidbeckingsale:
@ajkunen @trws let me know what you think about this!

keasler:
There was a period in time when the Intel compiler would SIMDize loop using an unsigned int loop-index-variable, but would not SIMDize a signed loop index.  I understand this behavior may no longer be the case, but do we really need negative indices?  I like to keep things out that could prohibit optimizations from occurring in the arcane world of compiler optimizations.  If it is a must have feature that put it in.  If it is syntactic sugar, I suggest not going there.

davidbeckingsale:
They should fix that. I am using signed loop indices because the SAMRAI co-ordinate system allows for it. I could change it all to be 0-based, but I think that make it easier to introduce bugs.  Note that negative indices in RangeSegments are already supported in our CPU backends, as well as the regular CUDA one. This just allows forallN in CUDA to also operate correctly.

trws:
I'd be for allowing it.  It may not come up frequently, but I've run into a few odd places where it was actually reasonable, and don't really see a reason to disallow it.  To your point Jeff, maybe it would be worth having a warning to that effect? I think we could add one with a pragma if it's worth making the user aware.


keasler:
Ok.  Sounds like it is necessary for SAMRAI, so got for it.

keasler:
Also -- no reason to have RAJA warn, although it does bring up and interesting question of whether we should add a layer to RAJA that is capable of reporting anomalies.  There is something like that for Vista where the user registers a callback function that is passed an error level and a string by vista.  If the callback is not registered by the users, the messages don't appear.  Also, when the user registers a callback, they provide a mask value indicating which error messages are enabled.

ajkunen:
I’m all for negative indices..   I didn’t like the “-1” that I was using, but it worked in all of my use cases.

I posthumously approve of the INT_MIN approach.

-Adam

PR #33, opened 2016-05-30
adding a clang-format file to help style code

trws:
This encodes a format that is as similar as I could seem to get clang-format to go to the style used in RAJA overall.  The only real difference is that it is more likely to pull template parameters and function arguments onto a single line when they fit in the 80-column limit than the manual style.

I suggest folks give this a try, and see if it's close enough to be made a standard tool for styling.  If there are any particular complaints, it's entirely possible they can be fixed.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -17,7 +17,7 @@ PenaltyBreakString : 10
 
 # These improve formatting results but require clang 3.6/7 or higher
 BreakBeforeBinaryOperators : NonAssignment
-AlignAfterOpenBracket: Align
+AlignAfterOpenBracket: true
 BinPackArguments : false
 AlignOperands : true
 AlwaysBreakTemplateDeclarations : true
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,5 @@
 BasedOnStyle : google
-IndentWidth : 3
+IndentWidth : 2
 BreakBeforeBraces : Linux
 UseTab: Never
 AllowShortIfStatementsOnASingleLine : true
@@ -21,3 +21,4 @@ BinPackArguments : false
 AlignOperands : true
 AlwaysBreakTemplateDeclarations : true
 Cpp11BracedListStyle : true
+
--- a/.clang-format
+++ b/.clang-format
@@ -1,6 +1,7 @@
 BasedOnStyle : google
 IndentWidth : 2
-BreakBeforeBraces : Linux
+# BreakBeforeBraces : Linux
+BreakBeforeBraces : Attach
 UseTab: Never
 AllowShortIfStatementsOnASingleLine : true
 ConstructorInitializerAllOnOneLineOrOnePerLine : true
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,23 @@
+BasedOnStyle : google
+IndentWidth : 3
+BreakBeforeBraces : Linux
+UseTab: Never
+AllowShortIfStatementsOnASingleLine : true
+ConstructorInitializerAllOnOneLineOrOnePerLine : true
+AllowShortFunctionsOnASingleLine : true
+AllowShortLoopsOnASingleLine : false
+BinPackParameters : false
+AllowAllParametersOfDeclarationOnNextLine : false
+AlignTrailingComments : true
+ColumnLimit : 80
+PenaltyBreakBeforeFirstCallParameter : 100
+PenaltyReturnTypeOnItsOwnLine : 65000
+PenaltyBreakString : 10
+
+# These improve formatting results but require clang 3.6/7 or higher
+BreakBeforeBinaryOperators : NonAssignment
+AlignAfterOpenBracket: Align
+BinPackArguments : false
+AlignOperands : true
+AlwaysBreakTemplateDeclarations : true
+Cpp11BracedListStyle : true
--- a/scripts/clang-format-all.py
+++ b/scripts/clang-format-all.py
@@ -1,5 +1,14 @@
 #! /usr/bin/env python
 
+##
+## Copyright (c) 2016, Lawrence Livermore National Security, LLC.
+## 
+## Produced at the Lawrence Livermore National Laboratory.
+## 
+## All rights reserved.
+## 
+## For release details and restrictions, please see raja/README-license.txt
+##
 import os, sys, argparse, glob, time
 import fnmatch
 
--- a/scripts/clang-format-all.py
+++ b/scripts/clang-format-all.py
@@ -1,0 +1,73 @@
+#! /usr/bin/env python
+
+import os, sys, argparse, glob, time
+import fnmatch
+
+from subprocess import call
+
+#if using pypar run this command as mpirun -np 16 --hostfile hostfile.txt -mca btl tcp,sm,self python  script.py
+
+class ClangFormat :
+  cmd = "clang-format"  # assumes you've done a use clang-3.8.0 or equivalent which has clang-format built/installed
+  pypar_avail = False
+  pypar = None
+    
+  def __init__(self):
+    try:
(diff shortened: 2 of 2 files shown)


.clang-format
```diff
@@ -0,0 +1,23 @@
+BasedOnStyle : google
+IndentWidth : 3
```
davidbeckingsale: I don't like this line....

rhornung67: Do you object to "3" or standardizing indentation?  Personally, I prefer 2 instead.

rhornung67:
Tom, what command do we run to invoke the formatting? Are there options for making the changes in place or doing the changes in a separate file so we can do a diff?

.clang-format
davidbeckingsale: The 3, I would be happier with 2 also

jonesholger:
-i will allow in-place edits while otherwise output goes to stdout

play with it in vim with 

map <C-K> :pyf <path-to-this-file>/clang-format.py<cr>     
imap <C-K> <c-o>:pyf <path-to-this-file>/clang-format.py<cr>

trws:
Thanks Holger. I also recommend using it with the autoformat plugin, as it will let you do sub-file formats.

I did 3 just because that seemed to be how the project was formatted already. Given my choice I would do something closer to a google style usually, which is two spaces and slightly different argument packing rules, but for a first cut thought I'd start with a match.


jonesholger:
I noticed a mix of 2 vs 3 in the project code; so I also like the idea to enforce it (to 2)

davidbeckingsale:
How would we switch to a new format (e.g. the Google style and 2-space indents) without making a weird uber commit where one user changes all of RAJA?

jonesholger:
Hmm, I was thinking cmake rewrite rules with capture to temp and then rename back to original. Doesn't sound bulletproof though

trws:
Either we make it a requirement in the test system, and only check changed files, or just do one big commit.  It's a bit nasty, but it will show up as almost 100% whitespace changes.

DavidPoliakoff:
I'd like to emphatically come out against "only check changed files," I think that creates some really perverse incentives ("sure I *should* fix this problem in the_problem.c, but if I put it in some_random_properly_styled_file.c I don't have to fix formatting"). I think an automated pass (crediting Rich or Jeff) before we have too much external collaboration would be a good idea, just have the problem solved

rhornung67:
Agreed.

davidbeckingsale:
I agree, let's get Rich or Jeff to do a big commit formatting all files, and then add checks to ensure any new commits have the correct style.

If we are going to do this big change, let's settle on a good style (rather than just a "matching" one). I think we should use 2 space, and I'm happy to use a more Google-y style for the other stuff.

rhornung67:
Also, we should coordinate with Adam to make his python scripts for generated forallN files generate code that conforms to the style conventions. We don’t want the style checker to be changing those — they won’t follow style conventions when regenerated.

davidbeckingsale:
@rhornung67 I believe that the code generation will be removed eventually (pending some compiler stuff being worked out #7 )

rhornung67:
But, until it is removed…

trws:
I'm not planning to let that sit much longer.  Either way, I would rather put that in a separate "generated" directory or similar that is excluded from checking than try and do all the work to adjust the generator.  Unless someone has a strong reason to do otherwise I don't see it as worthwhile to spend the extra time.


keasler:
This kind of thing should not be decided by me, so I have no comment on this branch -- whatever you guys decide.

davidbeckingsale:
@trws do you want to update the file with your preferred style (including 2-space indent) and we can take a look at that?

Something like what's used for flux would be fine I think.

trws:
Sure, I'll put that up once I can get back on my laptop.  Probably mostly fluxish, but without the space before all parentheses (including function calls for some reason).


davidbeckingsale:
Sounds good to me! We could then perhaps commit all the format changes in the RAJA developer meeting on Friday.

trws:
Ok, updated version.  This one is a closer match to what I normally do. You'll notice much of it is the same, but it goes to a 2-space indent, and while it allows single-line functions it no longer allows single line loops.  The difference in most places is pretty minimal, this is the complete diff for BaseSegment for example:

```diff
$ diff -w <(clang-format include/RAJA/BaseSegment.hxx) include/RAJA/BaseSegment.hxx
60,61c60,62
< namespace RAJA
< {
---
> 
> namespace RAJA {
> 
71a73
> 
75c77,79
<   explicit BaseSegment(SegmentType type) : m_type(type), m_private(0) { ; }
---
>    explicit BaseSegment(SegmentType type)
>       : m_type(type), m_private(0) { ; }
> 
83a88
> 
87a93
> 
91a98
> 
98a106
> 
141a150
> 
```

The biggest difference is probably in mathematical and boolean logic expressions and template parameter lists. We don't have many of the former, but complex boolean expressions will be wrapped such that their operator precedence is reflected in their indentation.

Template parameter lists will end up on one line unless they exceed the line length limit, so that's a change too:
```diff
165,166c181,184 // new
< template <typename EXEC_POLICY_T, typename LOOP_BODY>
< RAJA_INLINE void forall_Icount(const RangeSegment& iseg,
--- // old
> template <typename EXEC_POLICY_T,
>           typename LOOP_BODY>
> RAJA_INLINE
> void forall_Icount(const RangeSegment& iseg,
```

Anyway, I think this is reasonable, but as before let me know if there's something that should change.  One option that I didn't hit on was whether there should be spaces inside parens or not.  The codebase is about half and half it seems, so I left it at my default which is no extra spaces in parens.

DavidPoliakoff:
I like that style, has my :+1:

rhornung67:
Tom, this looks fine if others are happy with it.  I have no religious attachment to code formatting conventions.

jonesholger:
I can put together a quick Python script to do the whole works - unless someone else has already done so

trws:
That sounds great @jonesholger, go for it.  It wouldn't hurt to set up a git hook folks can use either, but don't feel obliged.

jonesholger:
clang-format-all.py processed the RAJA tree pretty fast (<10sec), I need to add the -style switch to explicitly pick up .clang-format file, although file diffs seem to have the correct format without the switch

trws:
It searches all directories above the working directory for a .clang-format by default, so as long as it runs from the directory of the file, or at least in the tree, it should use it.


jonesholger:
For the git hook, I can investigate LLVM's git-clang-format as a pre-commit hook, but this might require some project discipline to avoid edge cases. It can also be used standalone as git clang-format which runs clang-format over the changes staged for commit.

You could also setup the pre-commit hook to abort the commit if you "forgot" to run git clang-format. So it's more of a reminder in this case.

davidbeckingsale:
I think that we should set up a check in Travis that fails if code doesn't follow the format.
That way we can ensure PRs are only merged that have the same style.

For the git pre-commit hook, I like the idea of having it just abort the commit rather than magically style. This is effectively a local form of the Travis check.

Ultimately it could be best to just make everyone use the necessary vim/emacs/x plugin to auto-format as they develop.

trws:
Thanks to @jonesholger for letting me know this was requiring clang-format 3.9.  I forgot that 3.7 only supported the boolean version of the alignment parameter, so this is now compatible with any clang-format 3.7 and higher.

If folks are good with this, shall we just get it merged and do the re-format in a subsequent PR?

davidbeckingsale:
I'm happy for it to be merged.

jonesholger:
@trws Merge sounds good.  I'll put a git-format wrapper around git-clang-format for the hxx extension, and also provide the pre-commit hook reminder in another PR.

Issue #34, opened 2016-05-31
OpenMP policies are using the wrong values


trws:
At the least, the segment iteration policies use `omp_get_max_threads()` where it should be using `omp_get_num_threads()`.  This works on some implementations, but it will break on others.

See here: https://github.com/LLNL/RAJA/blob/develop/include/RAJA/exec-openmp/forall_openmp.hxx#L970

davidbeckingsale:
It seems like there are a lot of cases where `omp_get_max_threads()` is used to get the number of threads. I have switched fixed the case you pointed at out in #36

PR #36, opened 2016-06-02
Use omp_get_num_threads for segment iteration

davidbeckingsale:


davidbeckingsale:
This was pointed out by @trws in #34 

@keasler @rhornung67 we are still using `omp_get_max_threads` in the reduction implementations. I don't necessarily think this a problem, since we want to support running with different numbers of threads during the application and I could see the case where the number of threads changes before performing the final reduction.

trws:
Because of the way it's iterating, it may skip components if num_threads and max_threads are different. The "+=numThreads" increment is really the worry.


keasler:
Tom, thanks for clarifying.

davidbeckingsale:
@trws do you agree with my assessment of the use of `omp_get_max_threads` in the reductions, for example: https://github.com/LLNL/RAJA/blob/develop/include/RAJA/exec-openmp/reduce_openmp.hxx#L104

trws:
Yes @davidbeckingsale, in the reduction, that's the correct value to use, since it's the number of reduction cells to iterate over.

Really, the canonical way to do the segment iteration would be to use an `omp for` with a stride of 1.  The only reason this is a problem the way it is is that it may accidentally skip segments if the current number of threads is less than max_threads, which is actually a frequent condition unless one is very careful about setting environment variables.  The default value for max_threads on intel, and I think gomp but I'm not sure, is basically `INT_MAX`.

davidbeckingsale:
Everyone okay with merging this for now, and dealing with the segment iteration in a better fashion at a later date?

keasler:
@trws This is another helpful clarifying comment.  Basically we were using max_threads so we could get the value of OMP_NUM_THREADS.  In my environment on  the LLNL machines, OMP_NUM_THREADS is set to a default value  I had not stopped to think that it could be INT_MAX, and that information helps in rethinking a better solution.

keasler:
I approve the change in the two spots you have it, David. You can merge that file.

trws:
I'm good with this as long as @keasler and @rhornung67 are.  In general, if `OMP_NUM_THREADS` is desired, we should use `omp_get_num_threads()` to read the `num_threads` ICV.  As long as `OMP_NUM_THREADS` is set, it usually also sets the default `OMP_MAX_THREADS` or `max_threads` ICV, but it's best not to count on those being the same, especially since the `num_threads` ICV can change dynamically at the creation of each data environment.

rhornung67:
I’m OK with this decision.  I will add an issue for the segment iteration so we don’t forget.

trws:
Ironically enough this issue, in another spot, just bit me debugging iterator-based indexsets on my laptop.  Using Intel OpenMP I'm getting a max_threads value of 708,608...

```
(lldb) n
Process 7017 stopped
* thread #1: tid = 0x135544, 0x0000000100022c0f lulesh-RAJA-parallel.exe`RAJA::buildLockFreeBlockIndexset(iset=0x0000000100200660, fastDim=45, midDim=45, slowDim=45) + 47 at LockFreeIndexSetBuilders.cxx:87, queue = 'com.apple.main-thread', stop reason = step over
Return value: (int) $0 = 8

    frame #0: 0x0000000100022c0f lulesh-RAJA-parallel.exe`RAJA::buildLockFreeBlockIndexset(iset=0x0000000100200660, fastDim=45, midDim=45, slowDim=45) + 47 at LockFreeIndexSetBuilders.cxx:87
   84  	{
   85  	   int numThreads = getMaxOMPThreadsCPU();
   86
-> 87  	   // printf("Lock-free created\n") ;
   88
   89  	   if ((midDim | slowDim) == 0)  /* 1d mesh */
   90  	   {
(lldb) p numThreads
(int) $1 = 708608
```

davidbeckingsale:
If it can return values that high do we need to instead use our own value? Something like `getMaxSupportedOMPThreads()` where we pick the max supported value at build time?

trws:
Maybe limit it to some multiple of [std::thread::hardware_concurrency](http://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency) by default?  In general, omp_get_max_threads() inside a parallel region should be ok (it was limit threads I was thinking of before, but max_threads in this context is not really safe either) but where we use what is something we'll have to keep an eye on.

PR #38, opened 2016-06-04
[DNM] Policy performance test

trws:
This implements iterators for range and list segments, adds policy implementations that take advantage of this, removes the otherwise more specific overloads, and adjusts IndexSet's implementation to call the correct overloads.  I haven't had time to do a thorough performance eval yet, but for clang-omp 3.5 on a mac, kripke performance does not change (or if it changes the change is within the error of the measurements).  More to come, but if anyone wants to play with it, let me know how it comes out.


trws:
This PR is dead, need to re-create due to some branching issues.

PR #47, opened 2016-06-09
Commit to modify broken Layout.hxx constexpr

DavidPoliakoff:
Hey, we have a constexpr variable being subtracted from, named linear, in Layout.hxx. I fixed both the script that generated the broken code and the generated code itself.


DavidPoliakoff:
Passes. Merge?

ajkunen:
I approve this change

PR #48, opened 2016-06-09
Formatting

jonesholger:
Ran a formatting pass on develop

I did notice that I had to do two passes, which was odd. The second pass seemed to only affect right-hand comment alignment. 

Anyway this should be good to go.


DavidPoliakoff:
remember, from here on out *anything* wrong with the code can be considered to be Holger's fault. It's right there in the git blame logs. Seriously though, thanks for pushing this through

ajkunen:
Yeah, that Holger guy really made a mess of this forallN stuff :)

jonesholger:
Uggh, Actually mess is right -- I think we need to standardize on a clang version as 3.8 is closer to what we want.  So, I'll push those changes and you can take a look

jonesholger:
Never mind -- reverting back to clang-3.7 since clang-3.8 formatting doesn't build (sigh)..

PR #61, opened 2016-06-20
Fix banner message concerning LULESH domain size.

keasler:


keasler:
Travis died trying to do a sudo command -- I don't think this bug is mine.

DavidPoliakoff:
Yeah, looks like a bug in our travis :/

DavidPoliakoff:
If only we had something to patch over the problems with apt, some other spackage manager we know about at LLNL. Do you know of any,  @davidbeckingsale ?

keasler:
Since this branch has an utterly trivial change (the variable name used in printf statements), I am going to go ahead and merge it.  I also deprecated a force-inline flag in the host-config for the Intel 17 compiler that was missed.

DavidPoliakoff:
I'll retroactively approve it, just so the git gods don't strike us down :)

Issue #63, opened 2016-06-22
Interface abstraction hardening tracking issue


trws:
This is meant to be a tracking issue for the general discussion around establishing the interface contracts RAJA presents to users, and requires internally.  @rhornung67, @keasler, @ajkunen, @davidbeckingsale, @DavidPoliakoff and myself have had a number of discussions on these topics, and I would be grateful for any clarifications or extensions you all would like to add on these topics.  Not all of these things will necessarily be done, but they're things we've talked about enough that I think they deserve to at least be considered, and this will help us keep track of them.

- Define the interface elements required of:
  - [ ] Iterables:
    - [ ] Segments: PR #39 makes a first pass at this, allowing any random access iterable to be passed as an Iterable, this may be a first step
    - [ ] IndexSets (if different from segments)
    - [ ] others: are forward iterators sufficient for sequential policies? This might be useful for implementing IndexSet style generators.
  - [ ] Execution policies
  - [ ] Backend implementations (like cilk/openmp/...)
- [ ] Clarify what is the job of the execution policy and what is the job of the iterable, these sometimes get conflated a bit and it may help for clarity of the interface to normalize it where possible.
- [ ] Define what interfaces are user-facing and what interfaces are "internal."  This is important for defining how things can be extended, and may allow us greater internal flexibility while maintaining user-facing API compatibility.
- [ ] Explore the possibility of normalizing the interfaces used for forall and forallN policies (highly related to the required interface of an execution policy, but not necessarily the same)
- [ ] Investigate what will be required to support:
  - [ ] Plugins, such as CHAI, profiling etc.
  - [ ] Asynchronous regions, including dependencies between them
  - [ ] Generalized, and possibly more efficient, reductions (option note, if the lambda is copied by value into each thread or task context, this may have some nice and interesting possibilities)
  - NOTE: the two above may be approachable as a common item, we've been calling one option for this a "context," somewhat in the style of CHAI's context mechanism but more generalized
- [ ] Wherever possible establish formal contracts for all of the above, in the style of those discussed in c interfaces and implementations or the flux RFC process, with what is an error checked at compile time, error checked at runtime, and what is an unchecked error.  This is *hard* and much more like working on a standard, but the more we keep it in mind the easier it would be to do later.

willkill07:
[...]

PR #68, opened 2016-06-22
3d generic reduction

jonesholger:
CUDA Reductions now support 3D block and grid dimensions
Nested GPU Unit Test exercises these reductions using the forallN construct


davidbeckingsale:
This looks good!

Can you confirm what the block size should be for the reduction in the forallN case. Are we assuming that BLOCK_SIZE == blockDim.x * blockDim.y * blockDim.z?

jonesholger:
Yes, BLOCK_SIZE should be the worst case size encountered and is == blockDim.x * blockDim,y * blockDim.z

BLOCK_SIZE is used to setup shared memory for the reductions.

One snag is that the BLOCK_SIZE used to setup the reductions may not be related to the launch parameters or to the variables used to calculate the launch parameters.  The Reduction classes don't have enough visibility to control this more tightly. For example we could for example setup dynamic shared memory as as a parameter at kernel launch, but then you would need to know how many reduction variables are participating.

davidbeckingsale:
That makes sense. Thanks @jonesholger

@trws do you think it would be appropriate to add some kind of assert/warning to the reduction constructor that mentions this requirement?

I think we need to get the RAJA logging layer implemented....

davidbeckingsale:
I approve this PR

trws:
I think an assertion would be good, as the static assert for power of two size is, but as of now we don't have a way to get both pieces of information in one reasonable place to perform the check. If the reduction were created by taking the policy as a parameter, we could do it, or if we get to some kind of context interface, but for now I think documentation is probably the best we can do.


keasler:
Won't this implementation leave 'holes'?  At a higher level, this masking is being used to screen what gets executed by the reduction code:

__global__ void forall_box2d_cuda_kernel(LOOP_BODY loop_body,
                                       Index_type corner,
                                       Index_type xExtent, Index_type yExtent, Index_type zExtent,
                                       Index_type xStride, Index_type yStride, Index_type zStride)
{
   Index_type xIdx = blockDim.x * blockIdx.x + threadIdx.x ;
   Index_type yIdx = blockDim.y * blockIdx.y + threadIdx.y ;
   Index_type zIdx = blockDim.z * blockIdx.z + threadIdx.z ;

   if ((xIdx < xExtent) && (yIdx < yExtent) && (zIdx < zExtent)) {
      Index_type ii = corner + xIdx*xStride + yIdx*yStride + zidx*zStride ;
      loop_body(ii);
   }
}

That means that some of the block would never get initialized, even if it is being used.

rhornung67:
I’m putting together a list of items we want to look into and one of them is cleaner coordination between an execution policy and a reduction policy — along the lines of what we discussed briefly at the last meeting.

trws:
Sounds like a good idea. Feel free to edit the issue in #63 to change or add items to look into by the way.


jonesholger:
The initialize shared memory section of the kernel could leave holes iff the number of threads in the block specified at kernel launch is less than BLOCK_SIZE/2. Is that what you're thinking?  Other than that there should be no holes.  We can add logic to cover the aforementioned, but you've already noticed in my talk that modulus operators will slow down kernels ;)

BTW Tom E. just sent me a "it works" note.

Jeff I'm not that familiar with your Box2D kernel, can we get some more tests setup.

jonesholger:
Should we merge this in and setup another branch for the corner, err Box cases ;)

jonesholger:
The Box case definitely limits visibility to how many threads are participating.  Without providing the reductions extent information they would be forced to figure it out on their own by filling in a mask prior to actually getting started- and slowing things down.  We already have a precedent for selecting reductions based on policy, so we could do the same for the box case -- and not leave this PR in limbo

trws:
That sounds like a pretty good idea to me.  Something else that occurred to me, as a stopgap measure to make this all more robust while we investigate more generic options, is we could require the cuda policies to all write their grid and block dimensions into a known place and copy the lambda before executing the kernel to give the reductions the information they need.

Something like this perhaps?

```cuda
thread_local int dimx,dimy,dimz,blockx,blocky,blockz;
class CudaReducerOfSomeKind {
  CudaReducerOfSomeKind(const CudaReducerOfSomeKind& r) {
    copy(r);
    get_dim_sizes();
    allocate_space();
  }
};

void forall(cuda_exec_policy<BLOCK_SIZE>, ..., BODY body) {
    blocky=blockz=dimy=dimz=0;
    blockx = BLOCK_SIZE;
    using Body_t = std::remove_reference<BODY>;
    Body_t b = std::move(body); // Copy/move/whatever
    // catch the reduction copy/move construction and initialize sizes and data, check that they match
    cuda_kernel<<<dimx, blockx>>>(..., b);
}
```

jonesholger:
I like where this is headed, but Reduction classes already have access to kernel launch provided block and grid dimensions, but I see where you're going with this. What I don't have is the subset of threads (sub-block) participating, it could be every other warp or every 4th thread for that matter. Technically the Box construct could be asking for a single thread to participate in 1x1x1 Box  to complete the calculation anywhere within a block size of say 32x32x1 within a huge grid of 2147483647x65535x65535 ( a lot of idle threads).  Foolish of me to call this PR _3d-generic-reduction_ where it's not even close.

jonesholger:
Philosophically, I have the accelerator model in my head as I'm old school. Try to launch what you want "all" threads to do to avoid divergence. Maybe stack up the variants using a "mapper" then asynchronously launch them. Policies could do some of this work. Kernels should be  as lightweight as possible.

trws:
Right, that makes sense @jonesholger.  There are a number of good ways to deal with that generically, but somehow I'm guessing that none of them will be palatable.

From your chai work, does the copy constructor for captures get called on the device when entering a kernel?  More importantly, does the destructor for them get called at the end?

jonesholger:
Yes, even in my pre-CVHAI work we had for the reductions

```
 //
  // Destructor executes on both host and device.
  // Destruction on host releases the unique id for others to use.
  //
  __host__ __device__ ~ReduceSum<cuda_reduce<BLOCK_SIZE>, T>() {
    if (!m_is_copy) {
#if defined(__CUDA_ARCH__)
#else
      releaseCudaReductionId(m_myID);
#endif
      // OK to perform cudaFree of cudaMalloc vars if needed...
    }    
  }

```

trws:
That opens some possibilities...

What would you think of having the actual operations update shared memory locations, or some thread specific location, then perform the actual block-wide reduction in the destructor?  That way you're guaranteed they all participate.

trws:
As a side note, that's actually how OpenMP reductions work on the GPU.  Not literally with a destructor, but with that kind of post-body reduction model.

jonesholger:
That actually reminds me of the very earliest experimentation when I started working with RAJA - fun times. But, we constructed the classes so the values live across multiple kernel calls - I think your model is one kernel and one destructor

trws:
Not necessarily.  It would be one destructor invocation per kernel launch, but it should be possible for it to work across multiple kernel invocations so long as the initialization in the next kernel doesn't blow away the partial results from the previous one.  At least, I haven't thought of a reason it would have to be, there may very well be one you've run into that hasn't crossed my mind.

jonesholger:
Yep, I should try this out.

jonesholger:
@trws I can see my way through to support the Box construct

A generic reducer needs to be done in three phases with respect to participating threads

1:  Initialize shared memory (all threads)
2: Load shared memory (subset of threads)
3: Perform folding calculation (all threads)

and for the code

1:  copy ctor initialize shared memory
2:  += or reduction operator loads shared memory based on extent logic
3: dtor runs the folding calculation

which according to my initial tests seem do-able.

---------------------------------------

Over the weekend I was thinking I needed a lot more scaffolding say by using the rajaCTX

ctor registers precondition/postcondition routines with rajaCTX (device side)

Kernels are adjusted to
{
   call registered precondition chain (all threads)
   call lambda body within extent logic (some threads)
   call registered post condition chain (all threads)
}

But bottom line I don't need to use this scaffolding until the rajaCTX is a go.

jonesholger:
Hey could I get @keasler and @davidbeckingsale to Review this and PR #71. Besides I'd like to see ReviewNinja's full process lest I merge without it.

davidbeckingsale:
Taking in to account the things we have already discussed, this looks okay for now.

include/RAJA/exec-cuda/reduce_cuda.hxx
```diff
@@ -931,104 +955,111 @@ class ReduceMinLoc<cuda_reduce<BLOCK_SIZE>, T> {
...
       lastBlock = false;
-      sd[threadIdx.x] = RAJA_MINLOC(sd[threadIdx.x], sd[threadIdx.x + 1]);
-      m_blockdata[m_blockoffset + blockIdx.x + 1] =
-          RAJA_MINLOC(sd[threadIdx.x],
-                      m_blockdata[m_blockoffset + blockIdx.x + 1]);
+      sd[threadId] = RAJA_MINLOC(sd[threadId], sd[threadId + 1]);
+      m_blockdata[m_blockoffset + blockId + 1] =
+          RAJA_MINLOC(sd[threadId],
+                      m_blockdata[m_blockoffset + blockId + 1]);
+      __threadfence();
```
keasler: The description of threadfence on this page says that the volatile qualifier must be used to ensure proper operation: http://docs.nvidia.com/cuda/cuda-c-programming-guide/#axzz4Ctlbutmc.  Also, the description implies that only one thread in one block can call threadfence, but the wording is so vague that I am not sure if that is really sufficient.

include/RAJA/exec-cuda/reduce_cuda.hxx
```diff
@@ -1103,6 +1134,11 @@ class ReduceMaxLoc<cuda_reduce<BLOCK_SIZE>, T> {
 #endif
       // OK to perform cudaFree of cudaMalloc vars if needed...
     }
+//    else{
```
keasler: Remove this comment block?

include/RAJA/exec-cuda/reduce_cuda.hxx
```diff
@@ -654,43 +663,53 @@ class ReduceSum<cuda_reduce<BLOCK_SIZE>, T> {
...
+       + gridDim.x * gridDim.y * blockIdx.z; 
+
+
+    int threadId = threadIdx.x + 
+               blockDim.x * threadIdx.y + 
+               (blockDim.x * blockDim.y) * threadIdx.z;
+
+
+    if (blockId  + threadId == 0) {
+      m_max_grid_size[0] = RAJA_MAX(gridDim.x*gridDim.y*gridDim.z, m_max_grid_size[0]);
```
keasler: Can we add an "after-the-fact" assertion in the destructor or casting method to verify that m_max_grid_size is actually smaller than the amount of memory we allocated for the sum reduction?  This would be better than the current state of never knowing if an internal buffer has overflowed.

include/RAJA/exec-cuda/reduce_cuda.hxx
jonesholger: Its threadId == 0 calling threadfence. So it's only one thread here

I can put volatile on m_blockdata -- which as you imply causes the compiler to use a store that bypasses L1.  Just for discussion sake It seems like if this were the case we would not need a threadfence to begin with. As a fence ensure that the global update is finished before I call the atomic increment and fall thru and then the "last" block has assurances the right values are stored.

Without it I've seen some rare races which will cause the unit test to fail otherwise. So the fence seems good here ;)

include/RAJA/exec-cuda/reduce_cuda.hxx
jonesholger: Yes, I got some other cleanup I can do too.

include/RAJA/exec-cuda/reduce_cuda.hxx
keasler: Right, I was just wondering if we should add the volatile type modifier  to be "fully compliant" with recommendations.

include/RAJA/exec-cuda/reduce_cuda.hxx
jonesholger: Sounds good, while I'm at it I'll change the reasonableRangeCheck on BLOCK_SIZE  to max out at 1024

include/RAJA/exec-cuda/reduce_cuda.hxx
trws: As a note on this, the L1 bypassing store isn't the real reason to use volatile there (although it is a good reason to keep in mind), it's real function is to keep the compiler from caching an intermediate value for output to that space in a register across the threadfence.  In practice, that would basically never happen the way this is written, but just to make sure the optimizer doesn't get uppity it's a good idea to mark it.  This is exactly the right place, and situation, to use a threadfence in by the way.

On a curiosity note, have you tried an atomic version of the min/maxloc reducers? It would require sm_30 I think, but the 128bit CAS should be sufficient to allow it to work the same way as the min/max reducers for everything but complex.

jonesholger: Hmm, have to admit to not seeing a 128bit CAS anywhere in the CUDA 7.5 docs, but I can dig around for the signature. Also, the compiler also insists that I define an explicit assignment operator for this volatile type .. ughh

trws: You're absolutely right.  I'm sorry, I shouldn't comment on these things 
when I'm out of it.  The CPU architectures all have it, the GPUs don't.  
I needed one for a structure in OpenCL a couple of years ago and ended 
up implementing one using the exch primitive to build a single-thread 
lock and use a pair of exchanges for write operations.  It's probably 
not worth doing that here.

davidbeckingsale:
As per discussion on #71 - is this ready to get merged?

trws:
@jonesholger, I think this looks like it's in good shape.  Do you want to get the last check @keasler mentioned before it goes in, or address that in a subsequent PR?

jonesholger:
Hi Tom, this still requires some additions, and weird testing (setting up circumstances locally that would hopefully never exist). I should be able to get this out before I leave today.  First, I'll finish up #71 which I'll resubmit.

keasler:
Can someone else approve this branch so Holger can commit it?

davidbeckingsale:
👍

davidbeckingsale:
@jonesholger you are good to merge this, just ignore the reviewninja check.

PR #70, opened 2016-06-26
[WIP] new reduction pattern in openmp reduce sum, mostly an example for now

trws:
I've been considering this for a bit, mostly after reading through these and seeing the capture behavior @jonesholger described in #68.  This version of ReduceSum works almost exactly the same way OpenMP reductions do internally, it creates a value local to each thread by copying the lambda into the thread using firstprivate.  The resulting reducer is const, so the value is mutable, the thread reduces into the value that's local to its stack, removing the need for TLS, false sharing avoidance and checking the thread number, and only folds into the common space once at destruction time.  The other change is that it never returns itself by value, which avoids the extra copy construction we currently have on every += operation on a RAJA reduction, instead returning a reference. It also can be default constructed with the default initial value in principle, and probably be turned into a cilk+-like generic monoid for CPU threading models.

Upshot of all of this, for the test code below, it's a bit over **28 times** faster than the current version:
(edit: given better compilers and better flags, the gap can swing between 35% and 30 times, but this version is always faster)

output:

```
Total=500000000
Time=7.94917
OMP_NUM_THREADS=4 ./old-reduction  27.89s user 0.67s system 324% cpu 8.792 total
Total=500000000
Time=0.278651
OMP_NUM_THREADS=4 ./new-reduction  1.79s user 0.63s system 209% cpu 1.157 total
```

code:

``` c++
#include <omp.h>
#include "RAJA/RAJA.hxx"
#include <iostream>
#include <vector>
#include <chrono>

{
    std::vector<int> v2(500000000, 1);
    int * __restrict__ v = v2.data();
    RAJA::ReduceSum<RAJA::omp_reduce, int> r(0);
    RAJA::ReduceSum<RAJA::omp_reduce, int> r1(0);
    RAJA::ReduceSum<RAJA::omp_reduce, int> r2(0);
    RAJA::ReduceSum<RAJA::omp_reduce, int> r3(0);
    RAJA::ReduceSum<RAJA::omp_reduce, int> r4(0);
    RAJA::ReduceSum<RAJA::omp_reduce, int> r5(0);
    // std::cout << "val=" << r << " addr=" << &r << std::endl;

    auto start = std::chrono::high_resolution_clock::now();
    RAJA::forall<RAJA::omp_parallel_for_exec>(RAJA::RangeSegment(0,v2.size()),[=](int i) {
            r += v[i];
            r1 += v[i];
            r2 += v[i];
            r3 += v[i];
            r4 += v[i];
            r5 += v[i];
            });
    auto stop = std::chrono::high_resolution_clock::now();

    std::cout << "Total=" << r << std::endl;
    std::cout << "Time=" << std::chrono::duration_cast<std::chrono::duration<double, std::ratio<1>>>(stop - start).count() << std::endl;   
    return 0;
}

```


review-ninja:
<a href="https://app.review.ninja/LLNL/RAJA/pull/70" target="_blank">ReviewNinja

DavidPoliakoff:
Oh this is cool, I think it may solve the issues @keasler and I saw at Intel where the reducer wasn't vectorizing because the compiler saw dependencies.... will test tomorrow, but wanted to ping Jeff because I think it'd be an interesting evolution on that development path

trws:
Chances are this version still won't vectorize, since the storage is widely distributed to avoid false sharing. That said, we could try a version that makes the actual storage contiguous, then it might.


trws:
@DavidPoliakoff it looks like you're right, it can actually vectorize at least the loads and the combining into the local value.  Instruments is telling me the fastest version isn't vectorized, but whatever it's doing is very, very fast.  The vectorized code was about 10 times slower, so... yeah.  Maybe a lack of alignment assertions on the vector? Not sure, either way, the optimizer does seem to be rather happier.

DavidPoliakoff:
Hey, screw it, vectorization is a practical thing, not an article of faith, if non-vectorizing is slower, great. The broader point (which you've addressed) is in having a construct which uses OpenMP in such a way that it makes the optimizer comfortable. My rough cut of a lesson is "avoid using omp_get_[something_thread_identifier_related] in your OpenMP classes to minimize false dependency detection," which is a lesson worth learning.

The performance gain is here is just bananas, I'm trying to learn exactly why, and think that rough cut is at least in the ballpark of what we should be looking at.

trws:
I'd say that's a good lesson to take from this.  Also, surprise for me, the `omp_get_thread_num()` is coming out as much more expensive just by itself than I would have believed without looking at a trace.  This is what Instruments gives me for the test:

```
Running Time	Self (ms)		Symbol Name
21071.0ms  100.0%	0.0	 	Main Thread  0x5a4ce9
7822.0ms   37.1%	7822.0	 	 __kmp_get_global_thread_id
5155.0ms   24.4%	5155.0	 	 __kmp_get_global_thread_id_reg
2986.0ms   14.1%	2986.0	 	 RAJA::ReduceSum<RAJA::omp_reduce, int>::operator+=(int) const
2631.0ms   12.4%	2631.0	 	 omp_get_thread_num
740.0ms    3.5%	    740.0	 	 main::$_0::operator()(int) const
727.0ms    3.4%	    727.0	 	 _platform_memset_pattern16$VARIANT$Haswell
686.0ms    3.2%	    686.0	 	 DYLD-STUB$$omp_get_thread_num
192.0ms    0.9%	    192.0	 	 .omp_microtask.
130.0ms    0.6%	    130.0	 	 _kernelrpc_mach_vm_deallocate_trap
1.0ms      0.0%	    1.0	 	 ImageLoaderMachO::parseLoadCmds(ImageLoader::LinkContext const&)
1.0ms      0.0%	    1.0	 	 stat64

```

DavidPoliakoff:
Good lord, that's bananas. What on earth is omp_get_thread_num *doing*? Oh well, file it under "haha nothing ever works and HPC frameworks are a fractal of sandcastles built out of smaller sandcastles" (an oddly large folder these days)

ajkunen:
Do you have any stats on how many times omp_get_thread_num is called in that trace below?  I’m curious what the time/call is.

If it’s once per loop iteration, like 500000000, then it’s only about 4ns per call… and that shit just adds up.

-Adam

trws:
Yeah, once per loop iteration.  I'm more surprised that the ReduceSum 
copy construction and destruction that happens on every iteration with 
the old += operator isn't a larger percentage.  Thinking on it some 
more, this shouldn't be a huge surprise, since the call has to hit the 
OpenMP library, and as a result has to be a shared library indirect 
call, but still it's more than I would have expected.

rhornung67:
@trws this looks good. interestingly, I was putzing around with pieces of this a while back, but couldn't get it to work -- making the loop body firstprivate is the part I missed. Is this guaranteed to be portable across all the compilers we care about?  Also, I don't see any issues lurking, but there are two pieces of functionality we need to support with the reducer objects. First, we need to be able to use the same reduction object in multiple, successive loops; i.e., create reducer object, run a loop reduction, get the reduced value (or not), then run another loop (possibly with intervening code between the loops) with the same reducer and access the aggregate reduction. Second, a few users have asked that we provide "reset" (or "reinit") methods to our reduction classes so they can do a reduction, access the reduced value, then reinit the reducer and run another reduction with it. Since it appears that you've removed the need for the reduction ID and the shared memory block (at least for the CPU case), maybe this is now a moot point.  Thoughts?

trws:
@rhornung67, the copy constructor behavior on firstprivate is required by the standard, so it certainly *should* be portable (famous last words, I know ><).  I will note that since I didn't add firstprivate everywhere, this is still really a toy rather than a production implementation, but at least it lets us test it out.

This version should work correctly across any number of loops, but it's always possible I made a mistake somewhere.  Any user that wants reinit can get it with just `r = ReduceSum<omp_reduce, int>(0)`, but we can certainly make a shortcut that would have the same effect.  It could also have reference counting to track the number of outstanding children that haven't synced back, or let a user know when they're using a child instead of the parent to get the value, but I left that out for simplicity and speed for now.  Probably well worth if for debugging now that I think about it though.

Also, I meant to ask, why are the default constructors inaccessible? Each reduction type has a well-known "nil" initial value (0 for sum, 1 for mul, "" for cat, etc.), is there a reason not to allow the user to fall back on those?

DavidPoliakoff:
Hey all,

I have checked this out, compiled and ran it with

gcc-4.9.0
intel-16

Both on rzhasgpu (because that's where it was). It worked on the tests, so I'm approving it. I'll reapprove if we decide to expand on it

rhornung67:
@trws I felt it was cleaner (and less error-prone) to force users to explicitly specify an initial reduction value. I always prefer making things explicit and promoting a particular usage pattern to help avoid unexpected outcomes.  If others think that is an unwise choice, we can discuss it.

DavidPoliakoff:
@rhornung67 I'll say that it's certainly unexpected, most things give the operation's identity for the real/integer sum/multiplication as a default, but I don't know that it's wrong, per se

trws:
Fair enough @rhornung67, it's certainly not wrong, it just makes it a bit more verbose.  Actually, the only thing in the interface I would really like to change is to add a `get()` method so that users can be explicit about retrieving the value rather than using a cast.

rhornung67:
@trws I agree with that addition

jonesholger:
So, I'll start another branch adding get() to the CUDA reductions.

trws:
Side thought @rhornung67, if it turns out that the firstprivate behavior isn't portable, we can get the same effect by interposing a single copy by value to a temporary in the same place.  Really, this is what firstprivate should do anyway, it just makes it explicit in the c++:

this:
```c++
Body b;
...
#pragma omp parallel firstprivate(b)
{
...
}
```

would become this, or maybe a copy by value function parameter pass:
```c++
Body b;
...
#pragma omp parallel
{
  typename std::remove_reference<decltype(b)>::type b = b;
  ...
}
```

rhornung67:
@trws Either way works for me.

davidbeckingsale:
@trws - are you planning to update this ready for a merge? I'm happy to work on updating it if you don't have time.

trws:
I have been planning to update it, but keep not getting to it.  If you 
find you have time it would certainly help.

davidbeckingsale:
Sure thing. Would you be opposed to me rolling this back to 659f89d and dealing with the CUDA reductions in another branch?

trws:
Not at all, I think there's a more complete prototype of the CUDA version somewhere else anyway.

davidbeckingsale:
@trws are you happy if we close this and open another PR for the CUDA reductions?

trws:
Yup, works for me.

PR #71, opened 2016-06-28
Feature/holger/addget

jonesholger:
Adds a **get** member function to each of the CUDA reduction classes. This is based on #68


review-ninja:
<a href="https://app.review.ninja/LLNL/RAJA/pull/71" target="_blank">ReviewNinja

DavidPoliakoff:
Sorry for being a pain, but could you add the same methods on the other execution policies? I doubt it would happen, but would hate to see somebody change from cuda to OpenMP and ask why their function no longer works. Otherwise :+1:

rhornung67:
@jonesholger Ditto on @DavidPoliakoff comment about adding the get method to the other reductions. Otherwise, I approve.

davidbeckingsale:
I agree with @rhornung67 and @DavidPoliakoff - add the gets to all reductions before this gets merged.

keasler:
Holger, I like how you implemented this to piggyback off the cast operator.  Thank you for implementing it this way.  In general, it's good if the RAJA functionality can be as transparent as possible with respect to existing serial code.  The only place I feel we really had to compromise this so far has been with the min(), max(), minloc(), and maxloc() methods.  By defining the get method as you did, people who want to stick with their old serial code can still do so.

davidbeckingsale:
As a further comment I don't think this should have been based on #68 - that's a separate feature and hence shouldn't be included in this PR.

Can you update this branch to contain only #526d25d and then update the other reductions (or let someone else do it)?

jonesholger:
@davidbeckingsale I half agree with you since #68 contains a hotfix.  But I can certainly back it out

davidbeckingsale:
Okay, my comment still stands. Let's make sure #68 is okay, merge that, then merge this 'gets' branch separately, based on `develop`.

jonesholger:
Blowing this PR away

PR #72, opened 2016-06-28
Added in NVCC flags to RAJA config file exported by CMake

DavidPoliakoff:
Hey all,

This is a much more meager solution than the full fix of the configuration discussed in this morning's meeting, this is meant as a hotfix. We now export CUDA flags as per this morning's meeting.

Reviews @davidbeckingsale @rhornung67 ? It's a fairly innocuous 3-liner


review-ninja:
<a href="https://app.review.ninja/LLNL/RAJA/pull/72" target="_blank">ReviewNinja

cmake/SetupCompilers.cmake
```diff
@@ -65,6 +65,8 @@ if (RAJA_ENABLE_CUDA)
   elseif(CMAKE_BUILD_TYPE MATCHES Debug)
     set(RAJA_NVCC_FLAGS -g; -G; -O0; -restrict; -arch compute_35; -std c++11; --expt-extended-lambda; -x cu; -ccbin ${CMAKE_CXX_COMPILER})
   endif()
+  MESSAGE("${RAJA_NVCC_FLAGS}")
```
davidbeckingsale: You don't need the message in here.

davidbeckingsale:
I believe that FindCUDA is a pain if the NVCC_FLAGS are not a CMake list - have you tested this out?

DavidPoliakoff:
Modifying to use a list construction leads to the following

```  set(RAJA_NVCC_FLAGS -O2;-restrict;-arch;compute_35;-std;c++11;--expt-extended-lambda;-x;cu;-ccbin;/usr/apps/gnu/4.9.0/bin/g++;-Xcompiler;-fopenmp)
```

Without spaces. Should I push that?

davidbeckingsale:
I think the list construction is best. We pass a list to FindCUDA here: https://github.com/LLNL/RAJA/blob/develop/cmake/SetupPackages.cmake#L59

I'm happy with this, nice work.

davidbeckingsale:
Pinging @cyrush

DavidPoliakoff:
This is also now a one-liner

davidbeckingsale:
Looks good!

davidbeckingsale:
👍

cyrush:
I usually use a string for flags case --  I don't think cmake will auto expand the list when passed to create a target via  cuda_add_library() or cuda_add_executable()

http://stackoverflow.com/questions/11594905/how-do-i-correctly-pass-cmake-list-semicolon-sep-of-flags-to-set-target-proper. We can  always change a list to a string, but thats a pain.

DavidPoliakoff:
So we have one vote for https://github.com/LLNL/RAJA/pull/72/commits/78ad2e182bf04ddbd99c4899b0d80a78babb7816
one vote for
https://github.com/LLNL/RAJA/pull/72/commits/e9b319b94a328522b9f18d28f3544c205eeaac3c
(minus a message I can remove)

I have no preference, any opinions on whether CUDA flags should be in a list or string? @trws @rhornung67 ?

davidbeckingsale:
My concern wasn't an opinion, it was a question - strings are fine as long as passing a string of flags to FindCUDA works correctly.
If you have verified that, I'm happy with strings since they will match all the other flags.

cyrush:
my q is if the work when wielding them before a call to cuda_add_exectuable() or cuda_add_library()
I know the string case will (that is what I am using currently, and what you are using in the nvcc host-config)

trws:
FWIW I've used a string with the stock FindCUDA in the past without issue, but don't see any particular reason to lower it to a string if we don't have to.  It would make it easier to filter things out, just for example, if it can be kept as a list I guess.

trws:
Oh, I also believe this should be approved, but given the recent discussions about when merging is or is not appropriate, I'm guessing we need an approve from either @rhornung67 or @keasler at least before this can be considered ready?

keasler:
For build related issues, I elect David Beckingsale as my approval proxy unless it is Beckingsale's branch, in which case I elect David Poliakoff as my proxy.

davidbeckingsale:
Okay, I'm happy to merge but we need one more NinjaStar - @trws can you star it?

trws:
Sounds like @DavidPoliakoff needs to do that, unless @keasler or @rhornung67 are comfortable with me as a reviewer in this situation.

rhornung67:
I approve.

DavidPoliakoff:
Sorry, @rhornung67 , get with the times, we no longer approve, we "give ninja stars to." Apparently. :p

davidbeckingsale:
@rhornung67 - can you add a star on ReviewNinja, or just a 👍 here...

rhornung67:
ReviewNinja doesn't work for me. I have been getting unresponsive script errors and other garbage.

trws:
@rhornung67, how about I kill ReviewNinja and we can try pullapprove? It seems much more stable, much leaner, and it's more configurable in terms of which reviewers we require, when, and how they put in an approval.  ( I didn't know it existed when we started this trial, or I probably would have gone there first)

davidbeckingsale:
Fine with me, happy to try a few different options out.

trws:
Merged since @rhornung67's approval was sufficient IMO.

davidbeckingsale:
Agree.

rhornung67:
Whatever you folks think is best.  ReviewNinja did seem flaky to me.

PR #73, opened 2016-06-28
Make existing tests use CTest

davidbeckingsale:
This is a stopgap PR until we get #50 up and running.

PR #74, opened 2016-06-29
adding pullapprove configuration

trws:
This adds a pullapprove configuration.  You can read the YAML for full details, but PRs will require two approvals from the group @ajkunen, @keasler, @rhornung67, and @trws (this is arbitrary and temporary, let me know who all should be on the list).  This system can also let us say one reviewer from a given group and one review from another to merge, so I'll be happy to set that up but I need to know what the preference is.

To use it, you can link through to pullapprove to approve or reject, or you can use a github comment to get the same effect.  I adjusted the approval regex to hit the common items, but you will need to have a line that starts with one of:

*  `:+1:` or `:thumbsup:` or the unicode equivalent thereto (:+1:)
*  `:shipit:` (:shipit:)
* Approved
* I approve

To reject, it needs a line starting with "Rejected" currently.

I would appreciate feedback and preferences from folks on how they would like this to work.  It's already enabled with a default non-yaml config so we can actually try it out on this PR.

CC: @DavidPoliakoff, @davidbeckingsale, @jonesholger

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1,0 +1,13 @@
+approve_by_comment: true
+approve_regex: '^([Aa]pproved|:shipit:|:\+1:|I [Aa]pprove)'
+reject_regex: ^Rejected
+reset_on_push: false
+author_approval: ignored
+reviewers:
+  members:
+  - ajkunen
+  - keasler
+  - rhornung67
+  - trws
+  name: pullapprove
+  required: 2
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -4,10 +4,17 @@ reject_regex: ^Rejected
 reset_on_push: false
 author_approval: ignored
 reviewers:
-  members:
-  - ajkunen
-  - keasler
-  - rhornung67
-  - trws
-  name: pullapprove
-  required: 2
+  - name: czars
+    members:
+      - keasler
+      - rhornung67
+    required: 1
+  - name: everyone
+    members:
+    - ajkunen
+    - trws
+    - davidbeckingsale
+    - DavidPoliakoff
+    - keasler
+    - rhornung67
+    required: 1


davidbeckingsale:
I approve!

This looks cool. For the number of approvals and the groups, I would suggest something like:
- two approvals, with at least one from @keasler or @rhornung67

trws:
Hmm... That's a good idea @davidbeckingsale, I'll see if I can get it to do that.

davidbeckingsale:
Thanks! One other thing, do you know whether we can get pullapprove to note approvals in existing PRs? For example, you approved #73 , but that's not showing up.

trws:
I don't think there's a way to get it to search old comments.  I'll double check, but we may just have to go back and redo it for the first time.

keasler:
[ attempt at getting "at least one review from the fearless leaders" ] 

Now let's think about this -- who is probably the last person in the world you would ask about how to set up a git commit process?

Seriously, when it comes to git and cmake decisions,   I think Rich will likely be the final arbiter.

davidbeckingsale:
Thanks for updating the config, hopefully that works! My question would be whether an approval from @keasler or @rhornung67 would count twice since they are in both groups.

trws:
Ugh, I really don't want to do this, but it looks like it has to be *in* the develop branch to work.  Not in a PR, so we might need to take a couple of passes at it and merge the file straight in until we straighten out the preferences.

trws:
Well, I guess I'll just do it.  If anyone disagrees with this decision, let me know and I'll back it out. That way we can get to refining it.

trws:
Ok.  I think we have the configuration working as @davidbeckingsale suggested.  @keasler, would you mind putting an accept comment on one of the PRs that you think is ready so we can see whether it counts you for both or not?

davidbeckingsale:
It counts twice. See #73 

We can probably just set the count of the "everyone" to 2 and it will work as expected.

trws:
Got it.  Updated to 1 and 2, seems to be giving what I'm after now.  Not 
sure why the reviewninja sigils are sticking around, I've removed that 
integration, but it must be sticky on existing PRs or something.

PR #80, opened 2016-07-05
Tell CMake to use the right XL flags, and to define RAJA_COMPILER_XXXX

tomdeakin:
Without a valid RAJA_COMPILER_XXXX, RAJA won't build and you get
RAJA_INLINE not found errors.


cmake/SetupCompilers.cmake
```diff
@@ -53,6 +53,10 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES GNU)
 elseif (CMAKE_CXX_COMPILER_ID MATCHES Intel)
   set(RAJA_COMPILER "RAJA_COMPILER_ICC")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+elseif(CMAKE_CXX_COMPILER_ID MATCHES XL)
+  set(RAJA_COMPILER "RAJA_COMPILER_XLC12")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O5")
```
davidbeckingsale: Since `-O5` is not a required flag, I think that this line is unnecessary. Stuff like this can be passed to CMake in the standard way: `cmake -DCMAKE_CXX_FLAGS="-O5"`

tomdeakin:
One possible issue is hardcoding `RAJA_COMPILER_XLC12` as this may be true of other XL versions (I was using 13)

keasler:
Hi,
It was BG/L definitely required that 'builtin' keyword, and I'm not sure about BG/Q.  Also, I have a vague memory that -O5 enabled some important optimization on the BG/L platform, but my memory may be false.

tomdeakin:
In my case I'm using Xl to target Power 8, not a BG/* system.

cmake/SetupCompilers.cmake
```diff
@@ -53,6 +53,10 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES GNU)
...
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
```
trws: This should be set in cache.  I realize the intel compiler version doesn't, but they both should.  Just add `CACHE STRING "C++ compiler flags"` at the end of the set arguments.

davidbeckingsale: I agree with @trws

keasler:
Tom D., thanks for this.  We also have to support the BG/Q environment, so once someone verifies the change here does not interfere with BG/Q, I will approve this.

rhornung67:
@tomdeakin thanks. :+1:  @keasler I don't think this matters for BG/Q since gcc and clang are really the only viable compiler options there (unless you want to write functors :-()

davidbeckingsale:
Superseded by #142, which doesn't have the extra compile flags.

PR #88, opened 2016-07-07
adding the GPU unit tests to ctest

trws:
If cuda is enabled and tests are enabled, build and run GPU tests as part of ctest.  This will help support bamboo or any other GPU testing solution we go for, also makes it a bit easier to run all of the unit tests as a group when a GPU is available.


davidbeckingsale:
👍

Looks good! Thanks @trws 😄 

[Approved with PullApprove](https://pullapprove.com/LLNL/RAJA/pull-request/88/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=LLNL/RAJA)

rhornung67:
:+1: I'll add an issue to make sequential tests for the CPU nested tests

keasler:
I think it was Tom Epperly that made the point that having separate CPU and GPU tests seemed to go against what RAJA was promising to deliver.  I agree with Tom, but I'm not sure how to restructure the test implementations to address his observation.

trws:
We can add some new tests that use all of the policies that are enabled, rather than exercising one each.   It shouldn't be all that hard to put together, but I would much rather build it on top of what @davidbeckingsale has been building in #50 and now #89.

rhornung67:
@keasler Agreed. The CUDA tests were used to work out some issues we encountered when building the reductions. It's easy enough to rework the tests to do what @trws suggested.

rhornung67:
Looks good! :+1:

PR #89, opened 2016-07-07
Added GoogleTest for future testing

davidbeckingsale:
This branch is based on #50 , but leaves the existing testing setup intact. This avoids the need to convert all existing tests to GoogleTest before merging, and gets the GoogleTest infrastructure setup ready for writing future tests.


trws:
This looks really good, but something funny is happening with the clang builds, and I think with the GTest tests also.  The clang builds are all failing on travis right now, or rather the build succeeds but the test can't find the gtest build products.  Also, the runtime for the old CPU* unit tests is showing up as immeasurably small, where on ipa they take between .04 and .7 seconds each, so I'm really not sure if they're running or not on the GCC side.

I haven't dug into the log far enough to be sure, but I have a feeling this might be an old version of CMake biting us and not honoring the dependency between the externalProjectAdd and the other parts.  Or perhaps an issue of a header dependency on something that doesn't have an explicit dependency on the RAJA target to be sure the headers exist? If it's the latter, adding a step target for download might be a good thing.

Either way, once whatever that little oddness is is worked out, I'm all for this.  The new tests look great by the way!

davidbeckingsale:
Yup 😞 I'll take a look at getting this worked out.

CMakeLists.txt
```diff
@@ -58,6 +58,7 @@ option(RAJA_ENABLE_OPENMP "Build OpenMP support" On)
 option(RAJA_ENABLE_CUDA "Build CUDA support" Off)
 option(RAJA_ENABLE_CILK "Build Cilk support" Off)
 option(RAJA_ENABLE_TESTS "Build tests and exmaple applications" On)
+option(RAJA_ENABLE_EXAMPLES "Build simple exmaple applicaitons" On)
```
rhornung67: exmaple -> example

CMakeLists.txt
```diff
@@ -88,6 +89,9 @@ add_subdirectory(src)
 install(FILES ${PROJECT_BINARY_DIR}/include/RAJA/raja-config.cmake DESTINATION share/cmake/raja)
 
 if(RAJA_ENABLE_TESTS)
-  enable_testing()
   add_subdirectory(test)
 endif()
+
+if(RAJA_ENABLE_EXAMPLES)
```
rhornung67: This is meant to support basic examples we will build to illustrate RAJA concepts, correct?

test/unit/cpu/test-indexsets.cxx
```diff
@@ -0,0 +1,32 @@
...
+
+};
+
+TEST_F(IndexSetTest, IndexSetEquality) {
+  for (unsigned ibuild = 1; ibuild < NumBuildMethods; ++ibuild) {
+    EXPECT_EQ(index_sets_[ibuild], index_sets_[1]);
+  }
+}
+
+// TODO: tests for adding other invalid types
```
rhornung67: Eventually, we want to have an IndexSet model that supports arbitrary segment types without the switch-yard crap. So we may not need this in the long run, but is OK for now.

CMakeLists.txt
davidbeckingsale: Yup, there is just the pi in there right now.

CMakeLists.txt
```diff
@@ -57,7 +57,8 @@ set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/modules" ${CMAKE_MODULE_PATH}
 option(RAJA_ENABLE_OPENMP "Build OpenMP support" On)
 option(RAJA_ENABLE_CUDA "Build CUDA support" Off)
 option(RAJA_ENABLE_CILK "Build Cilk support" Off)
-option(RAJA_ENABLE_TESTS "Build tests and exmaple applications" On)
+option(RAJA_ENABLE_TESTS "Build tests" On)
+option(RAJA_ENABLE_EXAMPLES "Build simple example applicaitons" On)
```
rhornung67: applicaitons -> applications

cmake/SetupPackages.cmake
```diff
@@ -81,3 +81,35 @@ endif ()
...
+  ExternalProject_Get_Property(googletest binary_dir)
+  add_library(gtest      UNKNOWN IMPORTED)
+  add_library(gtest_main UNKNOWN IMPORTED)
+  set_target_properties(gtest PROPERTIES
+    IMPORTED_LOCATION ${binary_dir}/libgtest.a
+  )
+  set_target_properties(gtest_main PROPERTIES
+    IMPORTED_LOCATION ${binary_dir}/libgtest_main.a
+  )
+  add_dependencies(gtest      googletest)
```
davidbeckingsale: @trws I think these dependencies should be enough to ensure the correct build order for GoogleTest as an external project, but perhaps not. I've added an explicit dependency of RAJA on GoogleTest which should fix it.

trws: You've written it correctly, what I meant is that old versions of cmake 
(for like 5 years or some crazy thing) didn't actually honor those 
dependencies and tried to build in parallel right past them.  The output 
looked a bit like what that used to do, so it made me wonder about the 
cmake version on travis.

davidbeckingsale: Oh I see what you mean. The other thing is to force a serial build I guess?

trws: Yeah, either build in serial or ensure that a sufficiently recent cmake is available.  Either way.

davidbeckingsale: We depend on 3.1, which is also what's being used on the LC machines (where this problem doesn't show up)

trws: True, though I did run into an issue like it with the raja-apps repo's dependency on raja.  Either way, it could well be something else (or something in addition to).

trws: I take that back. Looking at it now, it looks like pthreads isn't getting linked into the tests that don't actually use openmp or something.  Since gtest uses the pthread_{get,set}specific interface for thread-local state, it has to have it even for serial tests.

davidbeckingsale: Looking at it's CMake build system, GoogleTest should actually link against pthreads itself.
Is the problem with the clang 3.6 and 3.7 on Travis?

Note that CMake doesn't find an OpenMP flag for either of those compilers.

trws: Since libpthreads is an so, and libgtest is being pulled in as a static archive, it can't provide its own pthreads other than transitively through either libtool or cmake inclusion, neither of which seem to be in use.  Since there's no OpenMP flag, that's almost certainly the problem, there needs to be an add_libraries(pthread) or similar to get pthread linked in for anything that links with gtest.

davidbeckingsale:
@trws @rhornung67  I've fixed up the typo and added the necessary libraries for building GoogleTest. This should be good to go assuming the Travis build passes.

trws:
Yup, it looks good to me, and the pre-typo version passed all tests (that can pass, windows is broken pretty thoroughly right now).  I'll approve gladly assuming this set passes.

trws:
I approve.

👍

PR #91, opened 2016-07-08
[WIP] CUDA Reductions split into 3 phases to support boxsegment

jonesholger:
Since boxsegment has an arbitrary number of "participating" threads it is necessary to split the processing into phases where we can control the number of threads used in each phase. The code in this PR has equivalent performance to the unsplit case (verified by a 90 second timing test)

Initialization:  all threads
Load data: some threads
Folding: all threads

The technique borrows from @trws  PR #70 -- thanks Tom 

Initialization utilizes the device side copy constructor, and to make sure this is activate at the right time we needed to modify the kernels 

The reduction operators just load shared memory

The reduction dtors do the actual folding

```
template <typename LOOP_BODY>                                                                                                                                                                                                    
__global__ void forall_cuda_kernel(LOOP_BODY loop_body,                                                                                                                                                                          
                                   Index_type begin,                                                                                                                                                                             
                                   Index_type len) {                                                                                                                                                                             
  auto body = loop_body;         <--- do this before any extent logic to setup copy constructors                                                                                                                                                                                                
  Index_type ii = blockDim.x * blockIdx.x + threadIdx.x;                                                                                                                                                                         
  if (ii < len) {                                                                                                                                                                                                                
    body(begin + ii);                                                                                                                                                                                                            
  }                                                                                                                                                                                                                              
}       
```

Another modification is the use of "dynamic" shared memory, so addressing equivalent shared memory offsets can be done across the three device routines (phases)

I'm thinking we can perform a dynamic shared memory allocation by keeping track of reduction ID's in scope with their type sizes and BLOCK_SIZE. Right now it's one chunk just to demonstrate feasibility. 

One last thing was to modify the operator call so that it returns a reference; avoiding an extra copy.


include/RAJA/exec-cuda/forall_cuda.hxx
```diff
@@ -176,7 +182,7 @@ RAJA_INLINE void forall(cuda_exec<BLOCK_SIZE>,
 
   RAJA_FT_BEGIN;
 
-  forall_cuda_kernel<<<gridSize, BLOCK_SIZE>>>(loop_body, begin, len);
```
keasler: Is the third launch argument the number of shared memory bytes?

jonesholger: Yes,  for generality we could add some logic to check which reduction variables are in scope based on ID and also it's type size to compute it's offset, or do something easier since reduction variables are the only thing using shared memory, and use an array based on max reduce vars .. what do you think?

include/RAJA/exec-cuda/forall_cuda.hxx
```diff
@@ -88,9 +88,10 @@ template <typename LOOP_BODY>
 __global__ void forall_cuda_kernel(LOOP_BODY loop_body,
                                    Index_type begin,
                                    Index_type len) {
+  auto body = loop_body;
   Index_type ii = blockDim.x * blockIdx.x + threadIdx.x;
```
keasler: I'm not sure how we are getting away with only comparing the x dimension here.  The user block may be x=19, y=27, z = 4.

jonesholger: You're right -- those kernels need to be changed too

jonesholger:
@keasler  Ok Global ID calculation added

jonesholger:
@keasler .. this also begs the question, the kernel launches are all only 1d as compared to the forallN variants.  But at least the kernels are more generic now.

keasler:
The BoxSegemt branch actually launches 3d thread blocks, and I think that going forward some other applications are going to want to do the same in order to extract performance.

include/RAJA/exec-cuda/forall_cuda.hxx
```diff
@@ -201,7 +230,7 @@ RAJA_INLINE void forall(cuda_exec_async<BLOCK_SIZE>,
 
   RAJA_FT_BEGIN;
 
-  forall_cuda_kernel<<<gridSize, BLOCK_SIZE>>>(loop_body, begin, len);
+  forall_cuda_kernel<<<gridSize, BLOCK_SIZE,BLOCK_SIZE*sizeof(double)>>>(loop_body, begin, len);
```
keasler: The value for len needs to be adjusted.  If the grid is 1027*2035*2097 then len is equal to that product.   I think this reduction has to be implemented directly in the context of the BoxSegment branch, but I see where this is going.  I'll take a stab at it on the BoxSegment branch.

jonesholger:
This PR is near ready for approval.  @MrBurmark and I have been banging on this for a couple of weeks.The main issue we tackled is to control the number of copies (for the reductions's copy constructors) and setup dynamic shared memory based on which reduction variables are in scope. It does make some changes to forallN constructs, and has been verified to work with nested and kripke, but may not go far enough with respect to forallN robustness checks based on conversations with @trws.  Which may mean we need to add more unit tests. Otherwise, this PR is good to go. @ajkunen it may be good to have you look at this as well.

This also brings in @MrBurmark async reductions, but I have not tested this in the context of this PR. They are currently benign as they are policy tagged.

include/RAJA/exec-cuda/MemUtils_CUDA.hxx
```diff
@@ -62,37 +62,90 @@
 
 namespace RAJA
 {
+/// Maximum number of blocks that RAJA will launch
+/// threads are reused to finish all work
+#define RAJA_CUDA_MAX_NUM_BLOCKS (1024 * 16)
```
rhornung67: Is the '16' here meant to be the same as the #define at line 78? If so, can we do this in such a way that they stay consistent (e.g., if the max reduction vars is change, then this will be changed automatically). Perhaps the best way to handle this is by making these const int variables?  Also, it would help to add some comments describing why these values were chosen.

include/RAJA/exec-cuda/MemUtils_CUDA.hxx
```diff
@@ -62,37 +62,90 @@
...
 #define RAJA_CUDA_REDUCE_TALLY_LENGTH RAJA_MAX_REDUCE_VARS
 
-///
-/// Typedef defining common data type for RAJA-Cuda reduction data blocks
-/// (use this in all cases to avoid type confusion).
-///
-typedef double CudaReductionBlockDataType;
+/// Should be large enough for all types for which cuda atomics exist
+/// includes the size of the index variable for Loc reductions
+#define RAJA_CUDA_REDUCE_VAR_MAXSIZE 16
```
rhornung67: Please add a comment describing what this #define means. It's confusing with alongside "RAJA_MAX_REDUCE_VARS".

include/RAJA/exec-cuda/MemUtils_CUDA.hxx
```diff
@@ -62,37 +62,90 @@
...
+
```
rhornung67: Please add documentation comments (using Doygen delimiters -- we eventually want to generate this documentation) to all types, #defines, typedefs, etc. where they are missing.

Part of our issue discussion at the last meeting centered around documentation. We may as well start addressing this

include/RAJA/exec-cuda/forall_cuda.hxx
```diff
@@ -150,9 +176,12 @@ __global__ void forall_Icount_cuda_kernel(LOOP_BODY loop_body,
                                           Index_type length,
                                           Index_type icount)
 {
-  Index_type ii = blockDim.x * blockIdx.x + threadIdx.x;
+
+  auto body = loop_body;
```
rhornung67: Why are we doing this?

include/RAJA/exec-cuda/forall_cuda.hxx
```diff
@@ -175,18 +204,26 @@ template <size_t BLOCK_SIZE, typename LOOP_BODY>
 RAJA_INLINE void forall(cuda_exec<BLOCK_SIZE>,
                         Index_type begin,
                         Index_type end,
-                        LOOP_BODY loop_body)
+                        LOOP_BODY&& loop_body)
 {
+  beforeCudaKernelLaunch();
```
rhornung67: Have we measured the overhead of these before/after calls for kernels without reductions?

src/MemUtils_CUDA.cxx
```diff
@@ -73,16 +74,30 @@ namespace RAJA
...
 // for CUDA reduction objects are used and which are not.
 //
-static bool cuda_reduction_id_used[RAJA_MAX_REDUCE_VARS];
+static bool s_cuda_reduction_id_used[RAJA_MAX_REDUCE_VARS];
 
 //
 // Pointer to hold shared managed memory block for RAJA-Cuda reductions.
 //
-CudaReductionBlockDataType* s_cuda_reduction_mem_block = 0;
+static CudaReductionDummyBlockType* s_cuda_reduction_mem_block = 0;
```
rhornung67: Why are we making these, and others below, static? If it is to ensure they are file scope, then I believe the more common C++ practice for this is to include them in an unnamed namespace.

include/RAJA/exec-cuda/forall_cuda.hxx
jonesholger: auto body = loop_body triggers the copy constructor for all captured reduction variables, this is required to sequence the 3 phase construct properly. In the older code, a copy was being triggered by the reduction's operator return, which was returned by value. In the new scheme the operator signature is

__device__ ReduceSum<cuda_reduce<BLOCK_SIZE>, T> const &operator+=(                                                                                                                                                                                              T val) const   

which passes back a reference and does not invoke a copy.

include/RAJA/exec-cuda/MemUtils_CUDA.hxx
jonesholger: A lot of what you're requesting stems from the fact that I've weaved in Jason's work, and we still need to do some cleanup/clarification and consistency checks.

include/RAJA/exec-cuda/MemUtils_CUDA.hxx
jonesholger: This is used to setup the dummy type to handle up to the worst case reduction variable size (eg minmaxloc)

and is used a few lines later in the following
/// dummy types for use in allocating arrays and distributing array segments                                                                                                                                                                                                                             
struct CudaReductionDummyDataType {                                                                                                                                                                                                                                                                      
  unsigned char data[RAJA_CUDA_REDUCE_VAR_MAXSIZE];                                                                                                                                                                                                                                                      
};

include/RAJA/exec-cuda/forall_cuda.hxx
jonesholger: Not yet!!

src/MemUtils_CUDA.cxx
jonesholger: I agree; so I'll have to ping Jason to see why they were changed, as it could be build related.

trws:
I see there's been quite a bit of discussion here in the last day or so.  Have been playing a bit more with the thread local reducer option that I initially tried with OpenMP, and came up with a way to make it work for cuda.  It uses the same 3-phase reduction technique more or less, but doesn't require dynamic shared memory, or managed memory actually, to work. It's not optimized all the way, but it seems to be about 10% faster than the original managed memory reductions, I'm not sure about compared to this branch though.

Downside? It costs one or two registers per reduction variable.  Not sure if this is better, worse or just different, but if you'd like to take a look I incorporated a sum reduction for cuda that does this in #70 just now.

jonesholger:
@trws I like your approach especially wrt simplifying copy ctor, and moving "all" of the work into dtor, and removing the need for dynamic shared memory - very nice.  I plan to explore this further when I get back from vacation.  I didn't think the class variable m_reduced_val would be fast to accumulate into in the operator += on a thread basis. Also, thinking about other use cases where you could call += multiple times for the same index (thread) in a forall, and finish with dtor.

MrBurmark:
@trws I briefly tried doing something similar at one point.  The problem I had was that instead of adding a couple of registers per reduction variable per thread it added many more. I didn't look into it too deeply, but my guess is that the compiler wasn't optimizing out the thread local copy of the loop body. I did a small occupancy study and found that in depending on your kernel using some combination of reduction variables using registers and some using shared memory was ideal from an occupancy standpoint. So it would be good from an optimization standpoint to have both options.

trws:
That's a good point.  The downside to the version over in #70 is register pressure, for sure, and I wasn't sure how it would come out.  The per-thread usage seems to be 1-2 registers on average per reducer, but depending on overall pressure and the number of reductions that could be too much.

@jonesholger, your comment reminds me that there is a possible gotcha with my current implementation, in that if a thread copy-constructs the reducer in a branch, there could be trouble in the destructor.  It probably needs some of the same reduce-into-parent logic on the device that the OpenMP version has to be sure it would work in all cases.

include/RAJA/exec-cuda/forall_cuda.hxx
```diff
@@ -175,18 +206,26 @@ template <size_t BLOCK_SIZE, typename LOOP_BODY>
...
+
+  auto body = loop_body;
+
   Index_type len = end - begin;
-  size_t gridSize = (len + BLOCK_SIZE - 1) / BLOCK_SIZE;
+  size_t gridSize =
+      RAJA_MIN((len + BLOCK_SIZE - 1) / BLOCK_SIZE, RAJA_CUDA_MAX_NUM_BLOCKS);
```
MrBurmark: This calculation of gridSize assumes a loop in the kernel to get all work done, which is not included. This should be changed to an assert on gridSize or the loop over len should be added back into the kernels.

include/RAJA/exec-cuda/forall_cuda.hxx
```diff
@@ -709,18 +866,23 @@ template <size_t BLOCK_SIZE, typename LOOP_BODY>
...
-                                                                loop_body);
+    executeRangeList_forall_Icount<cuda_exec_async<BLOCK_SIZE>>(seg_info, body);
 
   }  // iterate over segments of index set
 
   cudaErrchk(cudaPeekAtLastError());
   cudaErrchk(cudaDeviceSynchronize());
+
+  afterCudaKernelLaunch();
 }
```
MrBurmark: These functions ultimately call back into the other cuda forall launcher functions, so they do not need to call before/afterCudaKernelLaunch or to copy the loop_body. In fact it would be better to pull those calls up into this function entirely instead of calling them for each function call. Also is there a need for these functions to be always synchronous?

jonesholger:
@MrBurmark thanks for putting in the async option (bool) as we discussed before I left on vacation. I may be able to extend my vacation now ;)

jonesholger:
@MrBurmark could you also add doxygen style docstrings/tags to the added documentation; at least for function calls

include/RAJA/exec-cuda/MemUtils_CUDA.hxx
```diff
@@ -62,57 +62,295 @@
...
+ * Used in CudaReductionDummyDataType to allocate arrays and accommodate all the Reduction types
+ * includes the size of the index variable for Loc reductions
+ */
+#define RAJA_CUDA_REDUCE_VAR_MAXSIZE 16
+
+/*!
+ * \def MACROSTR(x)
+ * Used in the static_assert macros
+ */
+#define STR(x) #x
```
rhornung67: These macros are not specific to CUDA mem utils stuff and should be moved to a central header file.

include/RAJA/exec-cuda/MemUtils_CUDA.hxx
```diff
@@ -62,57 +62,295 @@
...
+#define MACROSTR(x) STR(x)
+
+/*!
+ * \def RAJA_STRUCT_ALIGNAS
+ * abstracts the alignas keyword
+ */
+#define RAJA_STRUCT_ALIGNAS(dalign) alignas(dalign)
```
rhornung67: Should this macro also be moved to another header file?

trws: I think there's already an alignment macro in RAJA.hxx.

include/RAJA/exec-cuda/MemUtils_CUDA.hxx
```diff
@@ -62,57 +62,282 @@
...
+ */ 
+#define RAJA_CUDA_MAX_NUM_BLOCKS (1024 * 16)
 
-/// Size of reduction memory block for each reducer object (value based on
-/// rough estimate of "worst case" -- need to think more about this...
-#define RAJA_CUDA_REDUCE_BLOCK_LENGTH (1024 + 8) * 16
+/*!
+ * \def RAJA_CUDA_REDUCE_BLOCK_LENGTH
+ * Size of reduction memory block for each reducer object (value based on
+ * rough estimate of "worst case" maximum numvber of blocks)
```
rhornung67: "numvber" -> "number"

include/RAJA/exec-cuda/MemUtils_CUDA.hxx
```diff
@@ -62,57 +62,282 @@
...
-typedef double CudaReductionBlockDataType;
-
-///
-/// Struct used to hold current reduced value and corresponding index
-/// for "loc" reduction operations.
-///
-struct CudaReductionLocBlockDataType {
-  CudaReductionBlockDataType val;
+/*!
+ * \def RAJA_CUDA_REDUCE_VAR_MAXSIZE
```
rhornung67: Should document units here. I think its bytes, but am not entirely sure.

include/RAJA/exec-cuda/MemUtils_CUDA.hxx
```diff
@@ -62,57 +62,282 @@
...
+ *
+ ******************************************************************************
+ */ 
+void afterCudaKernelLaunch();
+
+/*!
+ ******************************************************************************
+ *
+ * \brief Updates host tally cache for read by reduction varaible with id.
+ *
```
rhornung67: Should comment both of these or note that there are sync and non-sync versions of these and add '///' on line 310 so second one gets documented.

Would it be better to make these templates with the reduction policy as the template argument and explicitly specialize for each case? This would help avoid calling the wrong one erroneously.

include/RAJA/exec-cuda/MemUtils_CUDA.hxx
```diff
@@ -127,37 +352,35 @@ void freeCudaReductionTallyBlock();
...
- *       value for each thread, a single slot for the global reduced value
- *       across grid blocks, and a single slot for the max grid size
+ *       value for each thread block.
  *
  ******************************************************************************
  */
-CudaReductionBlockDataType* getCudaReductionMemBlock(int id);
-///
-CudaReductionLocBlockDataType* getCudaReductionLocMemBlock(int id);
+
```
rhornung67: Remove blank line so doxygen knows that the comment applies to the method immediately after it.

include/RAJA/exec-cuda/forall_cuda.hxx
```diff
@@ -127,51 +164,74 @@ __global__ void forall_Icount_cuda_kernel(LOOP_BODY loop_body,
...
+  beforeCudaKernelLaunch();
+
+  auto body = loop_body;
+
   auto begin = std::begin(iter);
   auto end = std::end(iter);
   Index_type len = std::distance(begin, end);
 
-  size_t gridSize = (len + BLOCK_SIZE - 1) / BLOCK_SIZE;
+  size_t gridSize =
```
rhornung67: Should we make this a macro to avoid potential errors?

include/RAJA/exec-cuda/forall_cuda.hxx
```diff
@@ -127,51 +164,74 @@ __global__ void forall_Icount_cuda_kernel(LOOP_BODY loop_body,
...
-                                               std::move(begin),
-                                               len);
+  forall_cuda_kernel<<<gridSize, 
+                       BLOCK_SIZE, 
+                       getCudaSharedmemAmount(gridSize, BLOCK_SIZE)
+                       >>>(std::move(body), 
+                           std::move(begin),
+                           len);
+
   cudaErrchk(cudaPeekAtLastError());
```
rhornung67: Make this sort of thing a macro also?

include/RAJA/exec-cuda/forall_cuda.hxx
```diff
@@ -205,7 +266,9 @@ RAJA_INLINE void forall(IndexSet::ExecPolicy<seq_segit, cuda_exec<BLOCK_SIZE>>,
   }  // iterate over segments of index set
 
   cudaErrchk(cudaPeekAtLastError());
-  cudaErrchk(cudaDeviceSynchronize());
+  if (!Async) {
```
rhornung67: Would it be better (less error-prone) to bury the conditional and the line above it in macro?

include/RAJA/exec-cuda/raja_cuda.hxx
```diff
@@ -144,14 +144,22 @@ struct cuda_exec_async : public cuda_exec<BLOCK_SIZE, true> {
 ///
 ///////////////////////////////////////////////////////////////////////
 ///
-template <size_t BLOCK_SIZE>
+template <size_t BLOCK_SIZE, bool Async = false>
```
rhornung67: What do we gain by having the bool parameter here?  We explicitly define cuda_reduce_async, etc. policies below.

include/RAJA/exec-cuda/reduce_cuda.hxx
```diff
@@ -111,6 +121,38 @@ __device__ __forceinline__ double shfl_xor(double var, int laneMask)
 #define double_to_ull(x) __double_as_longlong(x)
 #endif
 
+/*!
```
rhornung67: Should these be pulled out of the cuda reduction file and placed in raja_cuda.hxx instead?

include/RAJA/exec-cuda/reduce_cuda.hxx
```diff
@@ -111,6 +121,38 @@ __device__ __forceinline__ double shfl_xor(double var, int laneMask)
...
+ ******************************************************************************
+ *
+ * \brief Generics of atomic update methods used in reduction variables.
+ *
+ * The generic version just wraps the nvidia cuda atomics.
+ * Specializations implement other more atomics using atomic CAS.
```
rhornung67: remove "more"

include/RAJA/exec-cuda/reduce_cuda.hxx
```diff
@@ -334,151 +493,202 @@ __device__ inline void atomicAdd(double *address, double value)
...
-      sd[0] = RAJA_MIN(sd[0], sd[1]);
-      atomicMin(&(m_tallydata->tally), sd[0]);
+  /*!
+   * \brief Operator that returns reduced min value.
+   *
+   * Note: accessor only executes on host.
+   */
+  operator T()
+  {
+    if (Async) {
```
rhornung67: So the ability to do this sort of thing is why the bool template parameter was added to the reduce policy, correct?

include/RAJA/exec-cuda/reduce_cuda.hxx
```diff
@@ -830,154 +1175,221 @@ private:
...
-///
-/// Each ReduceMinLoc or ReduceMaxLoc object uses retiredBlocks as a way
-/// to complete the reduction in a single pass. Although the algorithm
-/// updates retiredBlocks via an atomicAdd(int) the actual reduction values
-/// do not use atomics and require a finishing stage performed
-/// by the last block.
-///
-__device__ __managed__ unsigned int retiredBlocks[RAJA_MAX_REDUCE_VARS];
+
+// set index first to avoid changes to val1 or val2 when writing to val_set
```
rhornung67: These macros are not specific to CUDA reductions and should probably be moved the the reducers.hxx file.

include/RAJA/foralln/Generic.hxx
```diff
@@ -55,6 +55,11 @@
 
 #include "RAJA/LegacyCompatibility.hxx"
 
+#ifdef RAJA_ENABLE_CUDA
+#include "RAJA/exec-cuda/MemUtils_CUDA.hxx"
+#else
```
rhornung67: Is the 'else' clause needed for something?

include/RAJA/exec-cuda/reduce_cuda.hxx
MrBurmark: These macros unfortunately are specific to the cuda version. The cuda version uses struct of arrays in shared memory to avoid bank conflicts so it can't use the generic max/minloc macros.

include/RAJA/exec-cuda/reduce_cuda.hxx
MrBurmark: Exactly, now we one implementation can do both policies.

include/RAJA/exec-cuda/reduce_cuda.hxx
MrBurmark: Do we plan to provide atomics based on execution policy?

jonesholger:
Tested under CUDA 7.5/8.0 including sample apps. So, good to go. Still need someone besides Rich to approve.

willkill07:
:shipit: Looks good to me. I approve.

include/RAJA/exec-cuda/raja_cuda.hxx
trws: I realize this is old, but I made that change when I was doing the rest of the iterator rework.  It uses a bool parameter so that only a single host-side function is required to implement sync and async kernel launches, the single function just checks the bool template parameter to see if it should include a sync or not.

PR #93, opened 2016-07-13
hotfix for compilation issues with nvcc and the new test integration

trws:
This is a set of hotfixes for build issues with the gtest integration that only showed up on odd environments.  Since a build with enable_cuda forces all including files to include cuda constructs, they only build with NVCC.  We really need to fix that.  Also, the gtest library needs to be built with the same ABI as the rest of raja, so it must be built with the same compilers. This version passes all CPU and GPU tests on rzhas.

### DIFF
--- a/cmake/SetupPackages.cmake
+++ b/cmake/SetupPackages.cmake
@@ -91,6 +91,11 @@ if (RAJA_ENABLE_TESTS)
       googletest
       GIT_REPOSITORY https://github.com/google/googletest.git
       GIT_TAG release-1.7.0
+      CMAKE_ARGS                
+          -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+          -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+          -DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}
+          -DCMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}
       INSTALL_COMMAND ""
       LOG_DOWNLOAD ON
       LOG_CONFIGURE ON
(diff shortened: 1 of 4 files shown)


davidbeckingsale:
👍 

Thanks for fixing this up! I agree about the CUDA stuff, will make an issue to track it.

cmake/SetupPackages.cmake
```diff
@@ -91,6 +91,11 @@ if (RAJA_ENABLE_TESTS)
       googletest
       GIT_REPOSITORY https://github.com/google/googletest.git
       GIT_TAG release-1.7.0
+      CMAKE_ARGS                
+          -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
```
davidbeckingsale: If a C or CXX compiler is not explicitly specified in the `CMAKE_XXX_COMPILER` variables, does this still work?

I was doing something similar yesterday and it just tried to pass an empty string. We might need to check the variable and build up the `CMAKE_ARGS` list depending on what is set.

trws: It should pull whatever the value determined for the cache is, since it's evaluated at runtime (or it's supposed to be) but it's worth testing it.

davidbeckingsale: It's working fine if the compilers are unspecified.

davidbeckingsale:
@rhornung67 can you approve this for merging?

jonesholger:
@davidbeckingsale -- I think he's on vacation

trws:
So... Any chance of an approve / merge @keasler?


davidbeckingsale:
@trws - since this change only really impacts the unit testing, I'm happy to merge it now. We can always back it out at a later date if necessary.

PR #97, opened 2016-07-15
Fixes atomicAdd(double.. name clash; use global namespace for atomicA…

jonesholger:
Easy fix really -- just qualified the use of atomicAdd(unsigned int*, unsigned  int)  with :: (global namespace) . Cleaned things up and moved our CAS based atomics back into RAJA namespace.

### DIFF
--- a/include/RAJA/exec-cuda/reduce_cuda.hxx
+++ b/include/RAJA/exec-cuda/reduce_cuda.hxx
@@ -68,7 +68,24 @@
 
 #include "RAJA/exec-cuda/raja_cudaerrchk.hxx"
 
-// The following atomic functions need to be outside of the RAJA namespace
+
+namespace RAJA {
+
+
+/*!
+ ******************************************************************************
+ *
+ * \brief Method to shuffle 32b registers in sum reduction.
+ *
+ ******************************************************************************
+ */
+__device__ __forceinline__ double shfl_xor(double var, int laneMask) {
+  int lo = __shfl_xor(__double2loint(var), laneMask);
+  int hi = __shfl_xor(__double2hiint(var), laneMask);
+  return __hiloint2double(hi, lo);
+}
+
+
 //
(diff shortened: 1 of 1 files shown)


include/RAJA/exec-cuda/reduce_cuda.hxx
```diff
@@ -1032,7 +1037,7 @@ class ReduceMinLoc<cuda_reduce<BLOCK_SIZE>, T> {
           RAJA_MINLOC(sd[threadId],
                       m_blockdata[m_blockoffset + blockId + 1]);
       __threadfence();
-      int oldBlockCount = atomicAdd(&retiredBlocks[m_myID], (int)1);
+      unsigned int oldBlockCount = ::atomicAdd((unsigned int*)&retiredBlocks[m_myID], (unsigned int)1);
```
keasler: Holger, what about for the older versions of NVCC?  Don't we still need our implementation of atomicAdd for those?

jonesholger: This doesn't do anything to replace our CAS variants, the :: qualifier on this atomicAdd is to bring in the CUDA default one for unsigned int -- which I should have done in the first place.   Also, this was tested with 7.5 too.

jonesholger: I realize it might be confusing, but the atomicAdd for retired blocks which is unsigned int caused the compiler to complain that it could not disambiguate for atomicAdd(double *, double) -- go figure. But, this global namespace qualifier was all it needed.

keasler:
@trws Tom, can you approve this very small branch if it looks ok to you?  I was most worried about backward compatibility issues that I believe Holger has checked.

trws:
Huh, I would have thought that would take more code.  Nice fix @jonesholger.

👍

PR #99, opened 2016-07-25
Feature/trws/bgq testing

trws:
This PR injects a TEST_DRIVER into each unit test's command, so we can launch tests onto the compute nodes on BGQ.  Because of the current state of cmake on BGQ, that 3.1 is available but has a broken ctest, the minimum required version needed to be decremented.  With this PR in, the gcc 4.8.4 version is auto-tested, and passes, on BGQ running all unit tests on the compute nodes.

@davidbeckingsale, is the 3.1 requirement hard for some feature?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@
 # 
 ###############################################################################
 
-cmake_minimum_required (VERSION 3.1)
+cmake_minimum_required (VERSION 3.0)
 
 project(RAJA LANGUAGES CXX C)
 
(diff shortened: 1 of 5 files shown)


davidbeckingsale:
This looks fine to me @trws 

I don't think 3.1 support was for anything in particular. Using newer versions is always good for things like `CXX_STANDARD` though.

DavidPoliakoff:
A painful tension is that we need 3.1 to work on BG/Q, but need the 3.6.whatever release candidate to make CMake detect Intel compiler flags correctly. We may need to get a good CMake for that

davidbeckingsale:
👍 - since this is clearly working fine with CMake 3.0

PR #123, opened 2016-09-06
Setup compiler-specific defines in CMake, remove macro switch

davidbeckingsale:
This branch moves the if-else macro checks in `config.hxx` to the CMake compiler setup file.
This will make it easier to handle unknown compilers and specify default defines.

Additionally, in future we could write compiler checks to test for specific align/simd and use whichever ones work, regardless of compiler.

### DIFF
--- a/cmake/SetupCompilers.cmake
+++ b/cmake/SetupCompilers.cmake
@@ -48,16 +48,41 @@ set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0" CACHE STRING "")
 
 if (CMAKE_CXX_COMPILER_ID MATCHES Clang)
   set(RAJA_COMPILER "RAJA_COMPILER_CLANG")
+
+  set(RAJA_ALIGNED_ATTR "alignas(N)")
+  set(RAJA_INLINE "inline  __attribute__((always_inline))")
+  set(RAJA_SIMD "")
+  set(RAJA_ALIGN_DATA "")
+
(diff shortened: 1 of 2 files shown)


trws:
I feel kinda dirty saying this, but what about builds where RAJA is installed as a package, or otherwise pre-compiled, and the user uses a different compiler to build their code against the installed version?  Are we planning to make that an unsupported configuration?

davidbeckingsale:
Good question 😕  in that case we should stick with macros, but then I would argue we should be using internally defined macros to detect compilers (e.g. `__NVCC__`)

trws:
Another question, do we actually want any of these to be part of the user API, or are they just internal? If they're internal, it doesn't matter, we can do it with cmake and just not install the resulting configuration header.

davidbeckingsale:
These are all internal, but they "leak" to the user application because they are used in headers.

trws:
Yuck... very true, that also means the headers will only work if they're detected for the compiler currently in use. Crud.

davidbeckingsale:
I have another branch that uses the complier-specific defines to check for specific compilers. Do you think that's better than `RAJA_COMPILER_X`?

trws:
For anything that persists after install, it probably is unless RAJA_COMPILER_X can be determined from the compiler processing the header.  I don't really like either solution, but it's a reality I guess.

davidbeckingsale:
Closing this, will make a branch that uses compiler-specific defines to enable features. This branch would have hard-coded the defines to the compiler used at install time.

Issue #125, opened 2016-09-09
Enabling NESTED yields unexpected compiler warnings


tepperly:
This report is based on the develop branch. I noticed that CMakeLists.txt has 
`option(RAJA_ENABLE_NESTED "Enable nested loop support" Off)`
Because I am using the nested capabilities, I thought I should build RAJA as follows:
cmake   -DCMAKE_C_COMPILER=`which gcc` -DCMAKE_CXX_COMPILER=`which g++` -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=`cd ../../external_install ; pwd` -DBUILD_SHARED_LIBS=On -DRAJA_ENABLE_CUDA=On -DRAJA_ENABLE_NESTED=On `cd .. ; pwd`

Enabling nested yields these compiler messages:
build-chaos-gnu-release/include/RAJA/config.hxx:86:0: warning: "RAJA_ENABLE_NESTED" redefined
 #define RAJA_ENABLE_NESTED
 ^

davidbeckingsale:
Thanks for the report :) Enable nested gets forcibly defined so that the tests build, it can be ignored for now, and won't impact your application builds (these redefinitions are only passed on the command line when building the tests)

PR #126, opened 2016-09-09
Add guards for RAJA_ENABLE_NESTED to tests

davidbeckingsale:
This address compiler warnings mentioned in #125 
Tests requiring nested functionality are only build when `-DRAJA_ENABLE_NESTED=On`

### DIFF
--- a/test/unit/cpu/CMakeLists.txt
+++ b/test/unit/cpu/CMakeLists.txt
@@ -41,7 +41,6 @@
 #
 ###############################################################################
 
-add_definitions(-DRAJA_ENABLE_NESTED)
 
 if (RAJA_ENABLE_CUDA)
   cuda_add_library(bis buildIndexSet.cxx)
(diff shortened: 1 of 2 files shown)


davidbeckingsale:
@trws @DavidPoliakoff can I get an approval?

DavidPoliakoff:
Approved . Nice catch

trws:
Since the user can use add that define, and use the constructs from the headers, even without setting that CMAKE define, do we really want to not run the tests?  It might be fine, but the default should be that RAJA_ENABLE_NESTED is on if that's the case.

davidbeckingsale:
Sure, we could revert it, and check `RAJA_ENABLE_NESTED` before we re-define it.

davidbeckingsale:
I reverted the previous changes and instead guarded the definitions.
@trws how's this?

trws:
Works for me.

LGTM

DavidPoliakoff:
Updates look fine to me, and pullapprove says it's good to go, as does Travis. Shiny merge button?

davidbeckingsale:
I'll just wait for Travis to pass, then I'll hit the button.

PR #141, opened 2016-10-27
Remove customized FindCUDA, use CUDA_SOURCE_PROPERTY_FORMAT

davidbeckingsale:


davidbeckingsale:
Just FYI, Travis builds are failing because it has an older CMake. We now require >= 3.3

davidbeckingsale:
@DavidPoliakoff @rhornung67 @DavidPoliakoff @jonesholger do any of you guys want to test this? Provided we use CMake > 3.3.0, this means we don't need the custom copy of FindCUDA anymore. (Note that CMake 3.4 in installed on LC machines)

davidbeckingsale:
Bumping this! @DavidPoliakoff @trws @rhornung67 @jonesholger

DavidPoliakoff:
My only real qualm is that right now we work with the default (3.1.2) CMake on our systems, and this would bump us up to needing a newer CMake. Might want to ask @MrBurmark if it'll bring up any drama for the teams he works with, might also really want feedback from @rhornung67 , I don't know when bumping a version is going to cause us grief. If either of those guys comment and say it's okay, I'll also approve, this seems technically sound

trws:
What cmake can we get on BGQ? If it's too low...

davidbeckingsale:
Good point - I'll check. Since this is only impacts CUDA we could 
probably just special case the version for BlueGene if necessary.

trws:
That might be the way to go, worth seeing what it would take to handle 
both before we commit either way.

rhornung67:
Regarding CMake versions on our platforms, here's our status with the Toolkit (FWIW). Latest installed cmake on BG/Q systems is 3.1.2. On CHAOS systems, latest is 3.4.1. There is an issue with Intel compilers with 3.1.2 that we have apparently figured out how to work around. 3.4 has some issues with findPython that we haven't figured out how to fix. 

Ultimately, we would like to use 3.6 if these issues are fixed, but we haven't tried it yet. So we build our own CMake 3.3.1 as a TPL dependency.  Unfortunately, we have not been able to build any version of CMake successfully on BG/Q. There's some weird Spack stuff going on that we hope may be fixed in the new release which is imminent.

At least one production code project here (that is an important RAJA user) uses cmake 2.8 by default.

davidbeckingsale:
One thing to note is that with an older CMake version, the `cuda_source_property_format` thing just doesn't work. So the CMake configuration can be used with an older version if CUDA is not needed.

DavidPoliakoff:
:+1: , that's good enough for me, just don't go building any BGQGPU machines

davidbeckingsale:
@rhornung67 as discussed this morning, shall we get this merged?

rhornung67:
:+1:

PR #146, opened 2016-11-04
Correct typo in ReduceMaxLoc copy constructor

davidbeckingsale:

PR #148, opened 2016-11-04
New OpenMP form for all reduction operations

davidbeckingsale:
This takes @trws code from #70 and extends it to all the other reduction types. I measured performance with the following test code:

```
  std::cout << "Testing ReduceSum...";

  std::vector<int> v2(500000000, 1);
  int * __restrict__ v = v2.data();
  RAJA::ReduceSum<RAJA::omp_reduce, int> r(0);
  RAJA::ReduceSum<RAJA::omp_reduce, int> r1(0);
  RAJA::ReduceSum<RAJA::omp_reduce, int> r2(0);
  RAJA::ReduceSum<RAJA::omp_reduce, int> r3(0);
  RAJA::ReduceSum<RAJA::omp_reduce, int> r4(0);
  RAJA::ReduceSum<RAJA::omp_reduce, int> r5(0);
  // std::cout << "val=" << r << " addr=" << &r << std::endl;

  auto start = std::chrono::high_resolution_clock::now();
  RAJA::forall<RAJA::omp_parallel_for_exec>(RAJA::RangeSegment(0,v2.size()),[=](int i) {
    r += v[i];
    r1 += v[i];
    r2 += v[i];
    r3 += v[i];
    r4 += v[i];
    r5 += v[i];
  });
  auto stop = std::chrono::high_resolution_clock::now();

  std::cout << "done in " << std::chrono::duration_cast<std::chrono::duration<double, std::ratio<1>>>(stop - start).count() << "s" << std::endl;
```

extended for all the other reduction operators. Results with the new pattern are:

```
Testing ReduceSum...done in 0.0824701s
Testing ReduceMax...done in 0.168889s
Testing ReduceMin...done in 0.168747s
Testing ReduceMaxLoc...done in 0.148311s
Testing ReduceMinLoc...done in 0.340467s
```

With reductions use the old pattern:

```
Testing ReduceSum...done in 1.32834s
Testing ReduceMax...done in 1.20044s
Testing ReduceMin...done in 1.27122s
Testing ReduceMaxLoc...done in 134.829s
Testing ReduceMinLoc...done in 1.08649s
```

I think the huge time for the ReduceMaxLoc is due to the compiler-generated copy constructor being called (fixed by #146).

Please sanity check this and let me know what you think!


rhornung67:
If it's faster, who cares if it's correct?!?  I'll need to look at it later.

davidbeckingsale:
It passes all the tests....

rhornung67:
:+1:

trws:
As I think I mentioned earlier, this looks great overall.  The only things to think about:

1.  set up a reproducible-ordered version (we already talked about this, it's a little annoying but will reuse everything that's here)
2. Most of these don't actually need `critical`, they should probably be either `pragma omp atomic` with the appropriate call or operation or implemented with C++11/RAJA atomics

Of these, 2 should be pretty quick and easy, and might make that performance difference even wider, especially because using a bare critical causes all instances to use a single global mutex (really not a fan of that part of the OpenMP spec, but there you have it ><).

davidbeckingsale:
Sounds good. We can talk about 1) next week. 

For 2), I can see that atomic would drop in for Sum, but I'm not sure for the others - what did you have in mind?

trws:
OpenMP atomic should work for everything but maxloc and minloc, it 
supports max and min natively, but maxloc and minloc are harder.  For 
those the best one can do is a loop over updates.  It would have to be 
something like an atomic min followed by an atomic CAS with a loop 
around if the CAS failed, or have a preprocessor check for a 
double-width atomic CAS or LL-SC and update both the min or max and the 
location as a unit.  Either would be faster than using a lock, but also 
rather more complicated.  Alternatively they could be set up to, on copy 
construction, reserve themselves a space in the ultimate parent and do 
the reduction only when the value is requested.  It would waste some 
memory but reduce contention and probably end up being simpler if not 
also faster... hmm...

davidbeckingsale:
@trws GCC is complaining about my use of `RAJA_MIN` with `atomic`:

```
/home/travis/build/LLNL/RAJA/include/RAJA/exec-openmp/reduce_openmp.hxx:115:47: error: invalid form of ‘#pragma omp atomic’ before ‘;’ token
       parent->val = RAJA_MIN(parent->val, val);
```

My reading of the OpenMP standard led me to believe it was okay (and it is in Clang). Thoughts?

trws:
It may be that it just isn't matching what GCC expects, what's inside that macro, is it using "max" the function, a ternary operator, ..?




On November 8, 2016 at 8:50:43 AM PST, David Beckingsale <notifications@github.com> wrote:

@trws<https://github.com/trws> GCC is complaining about my use of RAJA_MIN with atomic:

/home/travis/build/LLNL/RAJA/include/RAJA/exec-openmp/reduce_openmp.hxx:115:47: error: invalid form of '#pragma omp atomic' before ';' token
       parent->val = RAJA_MIN(parent->val, val);


My reading of the OpenMP standard led me to believe it was okay (and it is in Clang). Thoughts?

-

davidbeckingsale:
It's a ternary operator.

trws:
Ugh, apparently OpenMP only defines atomic min/max for Fortran.  You should probably just use the atomic compare and swap primitives, either the intrinsic or the c++11 standard ones, it would be basically this (pseudocode):

auto old = val;
auto tmp = val;
while (old < rhs && (old = atomic_cas(&val, val, min(old,rhs)) != tmp) {
  tmp = old;
}

I think there are actually some nice template functions for this in the CUDA reductions stuff, they would just need to have the function names changed.




On November 8, 2016 at 9:26:24 AM PST, David Beckingsale <notifications@github.com> wrote:

It's a ternary operator.

-

davidbeckingsale:
I got rid of the atomics on this branch, so it's ready to go with the new form using OpenMP `critical`. The version with atomics is on [`feature/reduction-reform-atomic`](https://github.com/LLNL/RAJA/tree/feature/reduction-reform-atomic).

trws:
I'm good with this at this point, only thing that needs to happen soon is get a reproducible version available.  Should we take as is, make this in parallel to the old, or wait for that?

I'm for either option 1 or option 2 personally, @rhornung67?

DavidPoliakoff:
:+1: , and I'd say make it in parallel to the old

davidbeckingsale:
My code never gets the same answer so I don't mind if it's not reproducible.

If we make it parallel then we need to decide:

- [ ] which version get's the `omp_reduce` policy?

I would propose that the new version gets the default policy. For a project whose goal is to enable code to run on multiple highly-parallel platforms it seems weird to make a reproducible reduction the default.

DavidPoliakoff:
Right there with Beckingsale, default is not reproducible, choose to enable that behavior

davidbeckingsale:
https://en.wikipedia.org/wiki/Entropy 😄

DavidPoliakoff:
As an institution of physics, we are required to always increase entropy

trws:
👍

davidbeckingsale:
Just to confirm, I'm going to add the old reductions back with the policy `omp_reduce_reproducible` - unless anyone has a better policy name? 😄

rhornung67:
:+1:   And to be safe, let's all keep obeying the laws of thermodynamics....

keasler:
omp_reduce_ordered

DavidPoliakoff:
"And to be safe, let's all keep obeying the laws of thermodynamics...."

Some people just have no sense of joy

rhornung67:
dS/dt >= 0, where S = k_B ln(Omega)...

davidbeckingsale:
I've added the old reducers with the policy name `omp_reduce_ordered`. Pending tests passing, I'll merge this branch and start a new release branch.

keasler:
I've been thinking... how about having omp_reduce_ordered and omp_reduce_unordered policies, and having the default omp_reduce policy defined as omp_reduce_unordered?  And also do the same for CUDA?  That way, you could have a cmake flag for an ordered vs an unordered reduction build that would simply change the definition of the xxx_reduce symbol.

DavidPoliakoff:
I'm with @keasler on that one. Good catch

PR #150, opened 2016-11-04
Add warnings to one gcc compiler in Travis configuration

davidbeckingsale:
This will fail if code is generating warnings.


rhornung67:
:+1: Thanks!

DavidPoliakoff:
👍  neet

rhornung67:
Wait. How does the Travis log file help me? I don't see any warning/errors, only that the compiler 'cannot build a simple test program'.

DavidPoliakoff:
Odds are he's still wrestling with CMake nastiness

davidbeckingsale:
Correct! There's an error in my Travis config. If I build locally I see 
things like:

     error: comparison between signed and unsigned integer expressions 
[-Werror=sign-compare]

rhornung67:
Sorry. I'm used to thinking that pull requests are ready to review. But that doesn't make sense in  this case.

davidbeckingsale:
No problem, I should have made a note of that. Travis needs the pull request to trigger the build, which is why I created it.

davidbeckingsale:
I'm going to commit both the new .travis.yml and fixes for all the warnings before merging, otherwise our travis builds for develop will start failing.

DavidPoliakoff:
In which Beckingsale gets tricked into fixing everything wrong in raja

davidbeckingsale:
Can you spot what's wrong with my .travis.yml - I'm obviously missing something obvious :(

DavidPoliakoff:
Looks like extra quotes on cmake line. More when not on bus

DavidPoliakoff:
Be sure to remove the echo before commit. Small thing, but would recommend

davidbeckingsale:
@trws do you have a second to take a look at my many attempts to pass extra flags to CMake? 😄  I based it off your coverage branch, but it seems like quoting is going awry and CMake actually thinks the flags are part of the compiler command.

trws:
What you have in the last one looks like it should work, if that doesn't 
you may just have to break down and use `export` in the before_install 
section to get it in there, or set CMAKE_CXX_FLAGS_ENV and have that on 
the cmake line or similar.  Seems like something that should just work, 
but looks like it's a persistent problem for travis users.

davidbeckingsale:
I've tweaked the CMake config to enable warnings as errors when `-DRAJA_ENABLE_WARNINGS=On` is passed to CMake, and enabled this for all CMake builds. There are now plenty of warnings (errors) throughout RAJA 😄

rhornung67:
@davidbeckingsale I just pushed a bunch of commits so that the code (and proxy apps) compiles cleanly for the most. There still a bit more work to do when -Wextra is also used. I'll try to finish that soon. I'm pursuing work therapy in an attempt to forget about the election results...

davidbeckingsale:
This is looking good - thanks for adding the extra host-config too.

include/RAJA/internal/defines.hxx
```diff
@@ -24,19 +24,66 @@
...
+ *  double myVar = ...
+ *  RAJA_DEBUG_VAR(myVar); 
+ *  
+ *                       
+ *  cassert(myVar > 0)  // variable used in an assertion
+ *
+ * \endcode
+ *******************************************************************************
+ */
+#define ATK_DEBUG_VAR(_x)   static_cast<void>(_x)
```
davidbeckingsale: This needs renaming 😄

rhornung67: Yeah, I know. I'm working on more changes....

PR #163, opened 2016-12-13
Remove 'breathe' from Sphinx requirements

davidbeckingsale:


rhornung67:
So are we planning to generate source code documentation in sphinx via breathe? That may be OK, but it's more limited (navigation, etc.) than doxygen.

ajkunen:
We are starting to use doxygen only in Ardra… doxygen 1.8.0 introduced Markdown support, making it unnecessary to use doxygen+sphinx

-Adam

davidbeckingsale:
I didn't think we were, but you had 'breathe' as a requirement in the sphinx config, so I fixed it this way. Since we aren't generating doxygen you could just instead remove the breathe requirement entirely.

davidbeckingsale:
I'm documentation tool agnostic, as long as it uses a sane format.

davidbeckingsale:
I have removed the 'breathe' requirement since we don't have doxygen anyway. We can always document classes using Sphinx. There's a good example of that here: http://ceres-solver.org/guide.html

PR #167, opened 2016-12-15
Use CMake to build Sphinx documentation

davidbeckingsale:
This adds a CMake target, `docs`, which can be used to build the Sphinx documentation as part of an install. This also removes the Makefile that built the sphinx and forced it up to the `gh-pages` branch.


davidbeckingsale:
Locally, `make docs`. ReadTheDocs is setup to automatically build whatever is in the docs folder, so this PR doesn't impact that at all.

PR #169, opened 2016-12-16
Atomic support

MrBurmark:
RAJA::atomic is a class that acts for the most part like std::atomic but can be captured in lambdas and used inside RAJA loops.

Known issues:
- only supports built-in numeric types, no pointers or use-defined types
- policy names are weird (cpu_nonatomic)
- cuda version uses cuda reduction variable backend for data motion


trws:
Why are there policies for this? I'm trying to imagine what the `nonatomic` policies would be useful for, in particular, and it's not coming clear.  Do you have a usecase for them?

MrBurmark:
Non_atomic gives you the option to turn atomics on and off depending on how you are compiling your code. So we would turn them off when we make a sequential only version of our code.

davidbeckingsale:
Closing for now, can revisit the idea of atomics in future.

PR #170, opened 2016-12-16
Remove pullapprove configuration

davidbeckingsale:


davidbeckingsale:
I think you can assign reviewers, but I'm not sure if we can require particular reviewers yet. I think for now we just insist on at least one reviewer, and all the tests (travis, appveyor) still passing.

PR #173, opened 2017-01-05
Use < and > in Loc reductions for consistency

davidbeckingsale:


davidbeckingsale:
CUDA reductions are unchanged since they already used < and > instead of <= and >= in their Loc reductions.

PR #179, opened 2017-01-20
Feature/cordery1/openmp4

mcordery:
Created off of develop branch using pointers in reductions. Trying to maintain how that works.
Created OpenMP4.5 versions of omp_parallel_exec and omp_reduce. parallel exec works for pi.cxx. ReduceSum tested in concept but still has link issues in RAJA.


include/RAJA/exec-openmp/target_reduce_openmp.hxx
```diff
@@ -0,0 +1,498 @@
...
+  //
+  explicit ReduceSum(T init_val, T initializer = 0)
+    : parent(NULL), val(init_val), custom_init(initializer)
+  {
+	hostid = omp_get_initial_device();
+	devid  = omp_get_default_device();
+	dev_val = new T;
+	is_mapped = false;
+	my_lev = 0;
+        dev_val = (T *)omp_target_alloc( sizeof(T), devid );
```
trws: I'm guessing you mean dev_val = new T not to be there, but one of them needs to go.  For this to get picked up correctly, it might be better to leave the dev_val = new T part and use `#pragma omp target enter data to(dev_val[0:1])` on it, that way it should be found correctly in the region which the omp_target_alloc value may not be.

include/RAJA/exec-openmp/target_reduce_openmp.hxx
```diff
@@ -0,0 +1,498 @@
...
+  //
+  // Destruction releases the shared memory block chunk for reduction id
+  // and id itself for others to use.
+  //
+  ~ReduceSum<omp_target_reduce, T>()
+  {
+
+   if( omp_in_parallel() && !omp_is_initial_device() )
+    { 
+#pragma omp critical
```
trws: Since the current foreach just does target + parallel for, this should work, but it will break as soon as teams and distribute go in.  The `atomic` directives are guaranteed to work for types up to 64bit though, so that might be a good solution.

include/RAJA/exec-openmp/target_reduce_openmp.hxx
```diff
@@ -0,0 +1,498 @@
...
+	{
+	  *dev_val += val;
+	}  
+
+    }
+    if( my_lev==1 && !is_mapped && omp_is_initial_device() )
```
trws: I might be misunderstanding, but I think this only brings the value back when the first value copy-constructed from the initial instance is destructed.  It would probably be good to bring it back on a get() or operatorT() as well.

mcordery:
Tom,
I haven't run into any problems with it really but I pinged the clang compiler folks and they didn't mention anything other than if you have an if on a combined construct the if is evaluated on the host. If it's a separate parallel region it can be evaluated on the device. I haven't tested this myself.
 
----- Original message -----From: Tom Scogland <notifications@github.com>To: LLNL/RAJA <RAJA@noreply.github.com>Cc: Matthew Cordery <cordery1@llnl.gov>, Author <author@noreply.github.com>Subject: Re: [LLNL/RAJA] Feature/cordery1/openmp4 (#179)Date: Sat, Jan 21, 2017 3:10 PM 
@trws commented on this pull request.
Thanks for getting the ball rolling on this Matt, some comments on the reduction for now. Do you happen to know what the status of if on combined constructs is with the IBM clang compiler? While I was reading through the changes it occurred to me that I'm not sure how well that's supported yet.
In include/RAJA/exec-openmp/target_reduce_openmp.hxx:
> +{+  using my_type = ReduceSum<omp_target_reduce, T>;++public:+  //+  // Constructor takes default value (default ctor is disabled).+  //+  explicit ReduceSum(T init_val, T initializer = 0)+    : parent(NULL), val(init_val), custom_init(initializer)+  {+ hostid = omp_get_initial_device();+ devid  = omp_get_default_device();+ dev_val = new T;+ is_mapped = false;+ my_lev = 0;+        dev_val = (T *)omp_target_alloc( sizeof(T), devid );
I'm guessing you mean dev_val = new T not to be there, but one of them needs to go. For this to get picked up correctly, it might be better to leave the dev_val = new T part and use #pragma omp target enter data to(dev_val[0:1]) on it, that way it should be found correctly in the region which the omp_target_alloc value may not be.
In include/RAJA/exec-openmp/target_reduce_openmp.hxx:
> +    dev_val(other.dev_val),+    custom_init(other.custom_init),+    is_mapped(other.is_mapped), my_lev(other.my_lev+1), hostid(other.hostid), devid(other.devid)+  {+  }++  //+  // Destruction releases the shared memory block chunk for reduction id+  // and id itself for others to use.+  //+  ~ReduceSum<omp_target_reduce, T>()+  {++   if( omp_in_parallel() && !omp_is_initial_device() )+    {+#pragma omp critical
Since the current foreach just does target + parallel for, this should work, but it will break as soon as teams and distribute go in. The atomic directives are guaranteed to work for types up to 64bit though, so that might be a good solution.
In include/RAJA/exec-openmp/target_reduce_openmp.hxx:
> +  //+  // Destruction releases the shared memory block chunk for reduction id+  // and id itself for others to use.+  //+  ~ReduceSum<omp_target_reduce, T>()+  {++   if( omp_in_parallel() && !omp_is_initial_device() )+    {+#pragma omp critical+ {+  *dev_val += val;+ }  ++    }+    if( my_lev==1 && !is_mapped && omp_is_initial_device() )
I might be misunderstanding, but I think this only brings the value back when the first value copy-constructed from the initial instance is destructed. It would probably be good to bring it back on a get() or operatorT() as well.

mcordery:
Given the issues with the clang compiler that need to be fixed, this isn't so much a merge request as it is a chance to review where it's at.  I was kind of out of it last week as I got pulled into something by my new boss and then on Friday we lost our machine over the weekend so I couldn't push this. Anyway, I don't have anything until Wed when I was going to Jeff's talk but I can easily come down at any time.

davidbeckingsale:
Superseded by #255

PR #180, opened 2017-01-24
Logging and Errors on the gpu

MrBurmark:
Initial attempt at supporting printf like logging on the gpu.
The RAJA::Logger<policy> object can be captured and used in a RAJA forall.
That logger object's constructor takes a handler function of the form void(*)(int udata, const char* msg).
That logger can then be used in a forall with the log and error methods. Those methods create logs that consist of an int userdata and message formatted using sprintf. Calls to log are expected to not be fatal and calls to error are expected to be fatal. The handler function for errors should abort execution, if not RAJA will call exit(1) after it returns.
Programs may register a cleanup function that will be called in the event of a fatal error. Cleanup callbacks may be added with RAJA::add_cleanup and removed with RAJA::remove_cleanup. The cleanup functions will be called before the handler function to the error is called.

On the cpu the handler function is called immediately at the call to log or error.
On the gpu the method RAJA::check_logs will check if messages are available and if so synchronize the device and then retrieve the messages. If using asynchronous gpu execution policies check_logs may not see the messages immediately, synchronizing the device and then calling check_logs will ensure all messages are handled. The buffer used for messages on the device is in pinned memory


davidbeckingsale:
Closing, this is kind of application-specific, and the (performance) cost of supporting this is high.

PR #184, opened 2017-02-08
Cuda Scan Asynch

MrBurmark:
Synchronize the device or not based on the execution policy.


MrBurmark:
@davidbeckingsale I think it makes sense to use the same cuda policy that is used elsewhere for consistency. Also block size may be ignored, but if we change the implementation later (using cub for example) we might end up using it.

davidbeckingsale:
This PR is included in #210

PR #188, opened 2017-02-08
[WIP/DNM] Kernel Streaming Model

DavidPoliakoff:
As an incentive for other people to come up with clever stream models, I have decided to put up the first stream model I thought of and see where it goes. I put in a very light wrapper around Intel TBB task groups, as I think they are simple and will be useful to people

TODO

- [x] Add a cudaExecuteOnStream policy seeming thing
- [x]  Make this use that
- [ ] Add functionality for asynchronous copies?
- [ ] Tests
- [ ] Docs


DavidPoliakoff:
First statement for anybody reviewing: we need a way to make sure to do a cuda____Synchronize on a wait if the last thing run on a stream before a wait is a CUDA function

rhornung67:
@DavidPoliakoff I believe the OpenMP/TBB runtime coordination will be a non-issue for Intel compilers since they use the same runtime (we tried this a couple of years ago and it worked fine). If an application pulls in TBB as a TPL, then it may be an issue. However, exploring this combination is a likely some way off for our apps teams.

DavidPoliakoff:
@rhornung67 nifty, I'll operate under that assumption then. I think I may need to change the API significantly (I think perhaps there should be an execution policy passed to the "run" function, so that the context we want to tie in to our asynchronous behaviors has something to be related to, also it'll allow for cudaStreamSynchronizes where appropriate), I'll probably try to get it to that point and then ask for more serious feedback.

DavidPoliakoff:
@trws @davidbeckingsale, mind giving the interface a look? The code sample is in

https://github.com/DavidPoliakoff/RAJA/blob/bff74fb72eba67024db365ecd94c276b75e0feb9/test/unit/cpu/dnm_stream_mockup.cxx

Work remains, I think we need a dedicated function to copy things (notice how the copy is running in RAJA::seq_exec, that *has* to happen and is unintuitive), I just want a sanity check on the code

DavidPoliakoff:
Current errors:

```
CUDAassert: driver shutting down /g/g0/dzpolia/src/RAJA/include/RAJA/exec-cuda/forall_cuda.hxx 262
CUDAassert: OS call failed or operation not supported on this OS /g/g0/dzpolia/src/RAJA/include/RAJA/exec-cuda/forall_cuda.hxx 262
```

Oof. Work continues, but this is a sane place to push, so I have

trws:
That interface really doesn't look too bad.  It's a little funky, but 
actually gives a good picture of what things will look like if we end up 
passing a context around.

DavidPoliakoff:
Also @davidbeckingsale this has the runtime policy thing done. Lookup wouldn't work in the way we discussed, but this is similar

DavidPoliakoff:
Hey folks, I've been redirected to other problems, but this is actually in a functional state, the example works. Notably, this version removes TBB from the code in favor of std::futures, which means that it'll work anywhere we have a C++11 implementation (anywhere RAJA can run). Anyway, I might clean up the build system and docs at home and polish this off, but I think the implementation of StreamPool could use a once-over.

Known issues

- [ ] Only forall implemented (need forall_Icount and such)
- [ ] Need CPU-only implementation (maybe different class, maybe ifdef'ing out the CUDA behaviors)
- [ ] Test reorg (maybe not testing the GPU functionality in the CPU unit test directory on something that isn't a unit test)

I'd also be curious of @rhornung67 's opinion of the runtime policy stuff (https://github.com/DavidPoliakoff/RAJA/blob/feature/tbbstreams/include/RAJA/forall_context.hxx), this is how I've implemented an ability for an outer construct like streams to communicate with the inner layers (like the cpu_stream_exec I implemented).

Anyway, I'm redirecting till COB, but it's ready for a test drive if people want, just build as normal on a GPU-enabled node

DavidPoliakoff:
Updated thoughts on this:

The mechanisms are mostly fine. As per @davidbeckingsale and I's conversation, a StreamPool should probably be a pool of objects called Streams, which applications could grab (so the stream-per-domain model could be functional).

The problem with this is that if we want different streams associated with different devices, we need to make sure that a CUDA kernel is never run on a stream that is not associated with the active device (set with cudaSetDevice). This requires some coordination among the streams, either with a StreamPool (hahaoops) or heavyhanded mechanisms (a cudaSetDevice per CUDA kernel launch, and cudaSetDevice is expensive).

Anyway, just wanting all the information to be out there.

davidbeckingsale:
Hey @DavidPoliakoff, I'm closing this for now just to clean up the PR list.

PR #193, opened 2017-02-15
Add OffsetLayout

davidbeckingsale:
OffsetLayout is a Layout struct that can be used with the view class to index data whose natural co-ordinates begin with some offset (i.e. not from 0^n for n-dimensional data).

The OffsetLayout is constructed with two initializer lists, one for the upper and and one for the lower co-ordinates. The implementation uses the variadic pattern as the regular Layout, so there is not a hard limit on the maximum dimension.


test/unit/test-layout.cxx
```diff
@@ -63,9 +63,9 @@ TEST(LayoutTest, 2D_IJ)
   ASSERT_EQ(0, layout(-1, -2));
 
   /*
-   * (0, -2) should be the second element.
+   * (0, -2) should be the second element??
    */
-  ASSERT_EQ(3, layout(0, -2));
+  ASSERT_EQ(1, layout(0, -2));
```
davidbeckingsale: I think that this should be the second element, since I'm using PERM_IJ. 

@trws am I misunderstanding how the permutations work?

willkill07: PERM_IJ is a no-op (no reordering occurs with the unpacking/swizzle of the permutation). For row-major access pattern (the default) it would be the second index.

davidbeckingsale: Makes sense, thanks! I think I was getting swizzled myself 😄

ajkunen:
This is really good.

DavidPoliakoff:
First time I've actually looked at the code. I really like this. Have we done any performance benchmarking yet? If performance is comparable, I'd love to at least look at maybe putting this in CHAI.

Issue #199, opened 2017-02-22
Different Segment types in forallN


MrBurmark:
Currently implementing coloring in forallN is not possible as the only segment types accepted are range segments.

davidbeckingsale:
Fixed by #206

PR #206, opened 2017-02-27
foralln: experimental non-rangesegment support

trws:
The only thing currently tested is RangeStrideSegment which appears to
work without issues.  Other iterables should also work, but may cause
issues if their iterators' call operators have non-integer return types.

This may address #199, but was specifically targeted at RangeStrideSegment support for the time being.


davidbeckingsale:
I think what @MrBurmark was getting at is supporting `IndexSet`s in `forallN`.

trws:
I got that impression, but that's worked for quite some time, you just can't have an indexset inside a cuda policy.

include/RAJA/exec-cuda/forallN_cuda.hxx
```diff
@@ -139,31 +139,34 @@ struct CudaIndexPair : public POL {
...
+      : distance(std::distance(std::begin(i), std::end(i)))
   {
     setDims(dims);
   }
 
   __device__ inline int operator()(void)
   {
-    int idx = begin + view(blockIdx) * threads_per_block + view(threadIdx);
-    if (idx >= end) {
+    int idx = 0 + view(blockIdx) * threads_per_block + view(threadIdx);
```
davidbeckingsale: Does this still work if begin != 0?

trws: It's been turned a bit inside-out.  Instead of storing begin and end, which may be meaningless, it stores the number of iterator results between begin and end and gives each thread one of those.  Those are then passed to `iterator::operator[]` by the new CudaIterableWrapper class, which should return whatever iterator value is at that offset from begin.

davidbeckingsale:
It would be good to add a quick exmaple of the IndexSet thing. Does it just work like:

```
typedef NestedPolicy< ExecList< IndexSet::ExecPolicy< seq_segit, seq_exec > > > policy;
forallN< policy >(iset, [=] (int i) { .. });
```

assuming `iset` is a valid IndexSet?

trws:
I actually just tested it.  Currently it requires adding RAJA_HOST_DEVICE to each of the call operators on ForallN_PeelOuter to work, and doing so causes an immense number of warnings but produces correct results.  It may be a matter of some refactoring required or some such thing.

trws:
FYI: I would appreciate some cuda testing from folks on this one.  It works for me with a couple different configurations, but some of this stuff just isn't covered by CI right now.

davidbeckingsale:
@MrBurmark does this address the issue you raised?

MrBurmark:
Yes, it does.

PR #207, opened 2017-02-28
Add build matrix item without OpenMP

davidbeckingsale:


davidbeckingsale:
Realized this was stupid because clang already builds without OpenMP.

PR #208, opened 2017-02-28
add abort and throw option in place of exit

trws:


trws:
So, are folks good with having exceptions by default unless an environment variable is specified?   This one is simple enough that I'd like to just get it in, but it's a notable enough change I'd rather not do it without being sure.  @rhornung67?

DavidPoliakoff:
I like it, but wait on Rich, agree it's a bit of a thing (exceptions give some people hives)

trws:
Thanks @rhornung67.

PR #214, opened 2017-03-15
Extract lambda argument types for automatic index-type deduction

willkill07:
Useful for `forallN`, and could also be used to make IndexSet argument types match lambda.


willkill07:
Converted to a PR with `extract_arg_t` residing under  a new `type_traits` header.

Most basic usage within the context of RAJA:

```cpp
#include "RAJA/type_traits.hxx"

template <typename Body,
          typename IndexType = RAJA::extract_arg_t<0, Body>>
void forall(IndexType begin, IndexType end, Body body) {
  for(IndexType i = begin; i < end; ++ i){
    body(i);
  }
}
```

include/RAJA/type_traits.hxx
```diff
@@ -0,0 +1,105 @@
...
+// IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#include <tuple>
+#include <type_traits>
+
+namespace RAJA
+{
```
davidbeckingsale: Do we want these kinds of functions in a `detail` namespace?

willkill07: Ideally, yes? Unless someone can think of a reasonable use case that we'd allow the programmer to use such a construct.

willkill07: I moved the entire header within `RAJA::detail`. I don't think we want to give users access to `extract_arg` since it's use is most likely limited to `forallN` internals

willkill07:
I just need a sanity check here. If we wanted to adopt this internally, I know I can edit the templates for `ForallN_BindFirstArg_HostDevice`, `ForallN_BindFirstArg_Host`, and within `ForallN_PeelOuter`. Is this all that needs changed to handle deduction of Index types or am I missing something else?

paging @ajkunen or @trws (who have far more experience with forallN internals than me)

trws:
That sounds right, that and the actual frontend functions.

willkill07:
I'm currently working on integrating the extraction where necessary, but I'm running into some special template hell. `forall` creates various function objects that store the underlying lambda, but these objects also store permutation information, which is critical when peeling in non-default permutation sequence.

Example:

```cpp
RAJA::ForallN_PeelOuter<
  RAJA::ForallN_Executor<
    RAJA::ForallN_PolicyPair<RAJA::simd_exec, RAJA::RangeSegment>>,
    RAJA::ForallN_Permute_Functor_impl<
      VarOps::integer_sequence<0, 1>,
      VarOps::integer_sequence<1, 0>,
      RAJA::ForallN_IndexTypeConverter<(lambda at
      /Users/wkillian/Desktop/RAJA/test/unit/cpu/main-nested.cxx:66:31) &, int, int>>>
```

Do you think the best approach to solving this would be to add a typedef with each function object to propagate the argument types up (and swizzle as needed with the `integer_sequence`s? Or what do you think would be the best solution?

PR #194, opened 2017-03-17
Inclusive scan init value argument confusing

willkill07:
Inclusive scans take a value argument that is not used. This is confusing and not found in other scan APIs.


willkill07:
I butchered this during conversion and wasn't on the correct branch when creating a PR. Keeping for the sake of history and a reminder to not be dumb.

PR #217, opened 2017-03-17
Add MultiPolicy wrapper policy for dynamic policy selection

trws:
This code still needs work, but it's enough to discuss the interface/implementation.  MultiPolicy stores a functor, or lambda or function pointer or whatever, which takes an iterable and based on that and whatever other data it stores internally decides which policy to invoke, and selects it by returning an integer offset into a list of policies in the parameter pack passed to MultiPolicy.  I'm not sure I love this interface, but it works and at least gives us a place to start.  The big questions are:

* What should the MultiPolicy selector take to choose a policy
* How does the selector signal the selected policy

The answer could even be that it takes instances of all of the available policies and directly invokes one, but I'm trying to avoid that if possible.  Perhaps ironically the MultiPolicy currently can't handle a nested MultiPolicy, because it only works with types rather than instances, but that should be easily fixable.


trws:
One other note here, as much for me to remember as anything, it may be possible to make this conditionally work with a default copy constructor given a selector that captures no data.  It should be possible to detect this by attempting to convert the selector to a function pointer as SFINAE, I'm just not sure how to remove the delete qualifier on the default constructor.

include/RAJA/MultiPolicy.hxx
```diff
@@ -0,0 +1,135 @@
...
+{
+  return MultiPolicy<Selector, Policies...>(s);
+}
+
+namespace detail
+{
+
+template <size_t index, size_t size, typename first, typename... rest>
+struct policy_invoker {
+  using policy = first;
```
willkill07: I would add a `static_assert` here in case the user attempts to call the multi-policy again. This would clean up the ~awful~ *awesome* error message that would most likely be emitted.

test/unit/multipolicy.cxx
```diff
@@ -0,0 +1,21 @@
+#include "gtest/gtest.h"
+#include "RAJA/RAJA.hxx"
+
+TEST(multipolicy, basic)
+{
+  auto mp = RAJA::make_multi_policy<RAJA::seq_exec,RAJA::omp_parallel_for_exec>(
+      [] (const RAJA::RangeSegment& r) {
```
willkill07: What are your thoughts on a type-aware dispatch pending the `Iterable` type? For example, a list of lambdas which are specialized for `RangeStrideSegment` or `RangeSegment`. It would be similar to the `std::visitor` pattern (observe) found with `std::variant`

```cpp
auto mp = RAJA::make_multi_policy<RAJA::seq_exec, RAJA::omp_parallel_for_exec>(
            RAJA::specialized_multi_policy(
              [] (const RAJA::RangeSegment &r) {
                 //......
              },
              [] (const RAJA::RangeStrideSegment &r) {
                 //......
              }));
```

`specialized_multi_policy` could extract the type arguments of each lambda/callable and invoke the correct one based on the invocation of `forall`

trws:
Thanks @willkill07, I like the idea of having a policy adapter like that to use lambdas. The current version can take a functor that does that, and supports polymorphic lambdas, but having a way to build the functor from lambdas would be cool.


---
Sent from Boxer | http://getboxer.com<http://bxr.io/PBI3C>

On March 17, 2017 at 8:39:31 PM CDT, Will Killian <notifications@github.com> wrote:

Overall, I really like this. Returning the index is probably the easiest way of choosing a policy to invoke.

In include/RAJA/MultiPolicy.hxx<https://github.com/LLNL/RAJA/pull/217#discussion_r106769028>:

> +    return s(i);
+  }
+};
+
+template <typename ...Policies, typename Selector>
+auto make_multi_policy(Selector s) -> MultiPolicy<Selector, Policies...>
+{
+  return MultiPolicy<Selector, Policies...>(s);
+}
+
+namespace detail
+{
+
+template <size_t index, size_t size, typename first, typename... rest>
+struct policy_invoker {
+  using policy = first;


I would add a static_assert here in case the user attempts to call the multi-policy again. This would clean up the awful awesome error message that would most likely be emitted.

In test/unit/multipolicy.cxx<https://github.com/LLNL/RAJA/pull/217#discussion_r106769489>:

> @@ -0,0 +1,21 @@
+#include "gtest/gtest.h"
+#include "RAJA/RAJA.hxx"
+
+TEST(multipolicy, basic)
+{
+  auto mp = RAJA::make_multi_policy<RAJA::seq_exec,RAJA::omp_parallel_for_exec>(
+      [] (const RAJA::RangeSegment& r) {


What are your thoughts on a type-aware dispatch pending the Iterable type? For example, a list of lambdas which are specialized for RangeStrideSegment or RangeSegment. It would be similar to the std::visitor pattern (observe) found with std::variant

auto mp = RAJA::make_multi_policy<RAJA::seq_exec, RAJA::omp_parallel_for_exec>(
            RAJA::specialized_multi_policy(
              [] (const RAJA::RangeSegment &r) {
                 //......
              },
              [] (const RAJA::RangeStrideSegment &r) {
                 //......
              }));

specialized_multi_policy could extract the type arguments of each lambda/callable and invoke the correct one based on the invocation of forall

DavidPoliakoff:
Let me know when this gets merged, I want to reimplement part streams in terms of this.

Other than -Werror silliness, looks good to me.

davidbeckingsale:
I like this, don't really see any problems with it right now. It would be cool to provide some default selectors (like the size-based one for OpenMP/Sequential).

trws:
I agree with that. Aside from one that takes a size, any ideas for others?


---
Sent from Boxer | http://getboxer.com<http://bxr.io/PBI3C>

On March 21, 2017 at 11:29:06 AM CDT, David Beckingsale <notifications@github.com> wrote:

I like this, don't really see any problems with it right now. It would be cool to provide some default selectors (like the size-based one for OpenMP/Sequential).

davidbeckingsale:
If we could pass another object to the selector function, I could see something like domain id being used to determine which cuda stream to launch on, or whether or not to use GPU.

Have you considered this use case?

DavidPoliakoff:
How expensive are the multipolicy objects? I wouldn't assume they'd be that bad. Could domain dispatch be done by having each domain have a selector which captures domain id?

davidbeckingsale:
I guess, but I kind of imagined each process having multiple domains, e.g.:

```
for (int d = 0; d < num_domains; d++) {
    Domain d = domains[d];
 
    forall(mp, d.getRangeSegment(), [=](int i) {
        // body
    });
}
```

DavidPoliakoff:
I guess, but I kind of imagined each process having multiple domains, e.g.:

`
for (int d = 0; d < num_domains; d++) {
    Domain d = domains[d];
 
    forall(d.mp, d.getRangeSegment(), [=](int i) {
        // body
    });
}
`

?

davidbeckingsale:
Yeah, that works. I had things laid out differently. Hard to code in your head in a meeting...

DavidPoliakoff:
Note that I'm not dumping on the idea of passing things into the selector, even if your problem is solvable without the capability it's entirely possible somebody else will have a use case

trws:
I thought about it @davidbeckingsale, but couldn't think of a good way to support that in terms of the interface.  A couple of options that come to mind are users adding such data into a wrapper segment type that just extends from the normal segment but carries their other data, or adding a method to the MultiPolicy that binds arguments to the selector.  Maybe something like...

```c++
auto make_multi_policy_and_bind(Selector s, Args... args) {
	auto real_s = [=](auto iterable) {
		return s(iterable, args...);
	};
	return make_multi_policy(real_s);
}
```

Clearly C++14 and so-forth, but there's no reason we can't make something like
that work.  Would it cover what you want to do?

davidbeckingsale:
I like the look of that :)

trws:
Now that by-value policies are supported in here I'm substantially happier with this.  @davidbeckingsale, @ajkunen, review?

include/RAJA/MultiPolicy.hxx
```diff
@@ -0,0 +1,166 @@
...
+{
+  return detail::make_multi_policy(
+      VarOps::make_index_sequence<sizeof... (Policies)>{}, s, policies);
+}
+
+template <typename... Policies,
+          typename Selector,
+          typename Iterable,
+          typename Body>
+RAJA_INLINE void forall(MultiPolicy<Selector, Policies...> p,
```
davidbeckingsale: Can you add Doxygen for this before we merge it?

trws:
That's a good point.  Yes, I'll add some Dox for it.

davidbeckingsale:
Great, thanks!

trws:
Assuming this passes the CI, these docs work for you @davidbeckingsale?

davidbeckingsale:
The docs look good, but do you think it's worth having a comment on the specialization? It doesn't need to say much, apart from maybe offer some description of what it can be used for, with links back to the builder function.

rhornung67:
@davidbeckingsale thanks for requesting documentation for new features.  I have been refraining from similar comments to avoid being branded an ogre. I figured we would have a documentation party at some point...

davidbeckingsale:
Thanks @trws!

davidbeckingsale:
@trws ready for this to be merged?

trws:
Sure, ready whenever at this point.

davidbeckingsale:
Great, thanks for implementing this.

PR #221, opened 2017-03-22
More header refactoring

davidbeckingsale:
Following this layout:

1.  `pattern`
    - This would contain our execution patterns like forall, forallN, reduction.
    - The generic headers for each pattern would live here, so if someone just wanted forall, they would just `#include RAJA/pattern/Forall.hxx`.
2. `policy`
    - This would contain the specific implementations of each pattern, grouped by programming model.
    - Let's say the user from before wants the openmp forall, they would `#include RAJA/policy/OpenMP.hxx` at some point after the forall include.
3. `util`
    - This could be split up further at a later point, but to start with would contain the Layout and View classes and stuff like that.
    - We could still have a detail directory, and an impl or internal directory for stuff the user shouldn't include directly.
4. `index`
     - Contains stuff related to index sets and segments.


davidbeckingsale:
Still lots of changes to make since the auto-refactoring has probably missed some of the new paths. Any comments on this design are appreciated.

The main thing I'm working on now is ensuring that each header is self-contained, so that we aren't quite so dependent on order of includes like we are now.

davidbeckingsale:
One other thing I'm interested in, is do we want to do have policies included first, or patterns? 

Currently, the generic pattern has to be included before the policies, but it's actually the reverse with scan (which has implementations inside a detail namespace).

trws:
What are the current blockers here?  As it turns out, my forallN rework for MultiPolicy is going to need the split between outer forall and impl::forall to work, to break a reference cycle actually.  Is there anything I can do to help get this through?

davidbeckingsale:
@trws - I just need to make sure it all builds with Cilk and CUDA, then it's ready to go. I can get this done today.

The other thing was the file naming - I had converted _everything_ to CamelCase, but I'm not sure if that's the best approach going forward.

trws:
That sounds great!

As to CamelCase, it's not my favorite but if it's good with everyone 
else I'm not going to argue, consistency would be a benefit regardless.

davidbeckingsale:
Okay, I think it's pretty much done - feel free to take a poke at it and let me know if you run into any problems.

trws:
Why a header refactor would cause this I'm entirely unsure, but the travis CPU reduction errors with OpenMP reproduce locally for me.  Any ideas?

davidbeckingsale:
Yup, same - it looks like it's only the ones that use the openmp segment 
execution policy. I'm trying to figure it out now.

rhornung67:
@davidbeckingsale @trws I think camel case file names is a good choice for files containing class definitions and implementations if that style matches the style of the class name, which it does in RAJA. For other things, like forall methods, simd, etc. I think camel case is confusing because it doesn't match the form of the name for methods or structs inside the files.  Thoughts?

include/RAJA/RAJA.hxx
```diff
@@ -63,39 +63,60 @@
...
 //
-#include "RAJA/IndexValue.hxx"
+#include "RAJA/index/IndexValue.hxx"
 
 //
 // Generic iteration templates require specializations defined
 // in the files included below.
 //
-#include "RAJA/forall.hxx"
+#include "RAJA/pattern/Forall.hxx"
```
rhornung67: As in my earlier comment, I think the form of this file name and the policy stuff above should match the style of the forall methods and policies, etc.

include/RAJA/exec-hcc/raja_hcc.hxx
```diff
@@ -0,0 +1,107 @@
+/*******************************************************************************
```
rhornung67: Are we sure this copyright works with the RAJA copyright?

include/RAJA/RAJA.hpp
```diff
@@ -94,29 +94,29 @@
...
 
 //
 // Generic iteration templates for perfectly nested loops
 //
-#include "RAJA/pattern/ForallN.hxx"
+#include "RAJA/pattern/forallN.hpp"
 
 
-#include "RAJA/pattern/Reduce.hxx"
+#include "RAJA/pattern/reduce.hpp"
```
rhornung67: We have 'Operators' and 'reduce'. Should the former be changed for consistency?

include/RAJA/index/RangeSegment.hpp
```diff
@@ -55,8 +55,8 @@
 
 #include "RAJA/config.hxx"
 
-#include "BaseSegment.hxx"
-#include "RAJA/internal/Iterators.hxx"
+#include "BaseSegment.hpp"
+#include "RAJA/internal/Iterators.hpp"
```
rhornung67: Same here. Should this be 'iterators'?

davidbeckingsale:
I'll get to all your comments :)

trws:
I realize I'm a bit out of step on the conversation here, but having some headers lower snake and  some camel based on whether the header contains a struct or class of the same name strikes me as unintuitive.  I'll readily admit that this may be personal preference, but for me it makes the naming look inconsistent and brings up questions for files like `reduce.hpp`.  Should that be `reduce`, because there's no actual Reduce class, or `Reduce.hpp`, because of the `Reduce.*` classes in the file?  Either way, this whole thing is a big step forward, just my 2c.

davidbeckingsale:
Yeah, it's a pain. That's why I went all CamelCase but it was kind of annoying.

trws:
Not a fan of all lower_snake_case?

Either way, the commit I just put up should address the multipolicy issue.  The forall in the multipolicy file needed to be in the impl namespace and needed a bit of extra care to resolve the recursion.

davidbeckingsale:
Yeah, I was missing the value-based policy thing too. Thanks! I'll bring up the naming issue at the meeting today and see if we can decide on something.

davidbeckingsale:
@rhornung67 shall I merge this?

PR #224, opened 2017-03-27
Remove memcpyAsync in cuda reduce async.

MrBurmark:
Use pinned memory and a copy kernel to stream the results back to host memory through the pcie bus. This is to avoid the use of memcpyAsync to pageable memory which may actually be asynchronous on the host on future platforms.
In this implementation reducer.get waits on a boolean flag set in the copy kernel to avoid synchronizing the device.
Currently the copy kernel is compiled into the library with MemUtils_CUDA. It may be better to pull the call to the copy kernel into a header to preempt possible linking errors down the line.

### DIFF
--- a/src/MemUtils_CUDA.cxx
+++ b/src/MemUtils_CUDA.cxx
@@ -130,6 +130,11 @@ namespace
    */
   CudaReductionDummyTallyType* s_cuda_reduction_tally_block_host = 0;
 
+  /*!
+   * \brief Pointer to pinned mem flag used to indicate completion of data copy to host.
+   */
+  volatile bool* s_cuda_reduction_done = 0;
+
   //
   /////////////////////////////////////////////////////////////////////////////
   //
@@ -337,8 +342,12 @@ void freeCudaReductionMemBlock()
 void getCudaReductionTallyBlock(int id, void** host_tally, void** device_tally)
 {
   if (s_cuda_reduction_tally_block_host == 0) {
-    s_cuda_reduction_tally_block_host = 
-        new CudaReductionDummyTallyType[RAJA_CUDA_REDUCE_TALLY_LENGTH];
+    cudaErrchk(cudaMallocHost((void**)&s_cuda_reduction_tally_block_host,
+                              sizeof(CudaReductionDummyTallyType) *
+                                RAJA_CUDA_REDUCE_TALLY_LENGTH + sizeof(bool)));
+
+    s_cuda_reduction_done = reinterpret_cast<volatile bool*>(
+          s_cuda_reduction_tally_block_host + RAJA_CUDA_REDUCE_TALLY_LENGTH);
 
     cudaErrchk(cudaMalloc((void**)&s_cuda_reduction_tally_block_device,
                           sizeof(CudaReductionDummyTallyType) *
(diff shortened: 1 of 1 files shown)


trws:
The point of using async was to avoid having the synchronous copy to avoid blocking all streams right? Is it overkill to keep the memcpyAsync and use a stream synchronize rather than device synchronize?

MrBurmark:
I tried that, but unfortunately the performance results were the same. We only use one stream and don't launch anything in other streams. It appears that the runtime sees that only one stream is active and synchronizes UM at streamSynchronize.
Using cudaEvents has the same problem and adds more overhead on the host. Though if you launch an empty kernel after recording the event and before synchronizing the event you can avoid the UM synchronization.

davidbeckingsale:
I agree with @trws here.

davidbeckingsale:
This is the kind of performance issue we should report to NVIDIA. I'm not keen on merging a pull request when it specific to one application use case and something that should really be handled by the programming model runtime (not RAJA).

trws:
Right, it would be because all of the UM is in the global domain so it 
syncs whenever you synchronize with anything at all.  God I hate UM... 
Anyway, how does it come out if you just use UM for the tally buffer? It 
has to have been synchronized for any of this to be valid anyway, so it 
shouldn't add any extra delay.

MrBurmark:
We used to have the tally buffer in UM, but writing to it required a synchronization during construction of the reduction variables. That would also add page fault overhead when it's read.

trws:
Do you have a benchmark you're using to test this, or is it straight in 
the app? I feel like there should be a way around this, but would rather 
poke at it than try and send everything through comments.  It might be 
worthwhile to work out some of the corner cases of UM sync anyway now 
that I think about it.

MrBurmark:
I have a little benchmark that tests these different patterns directly. I have been using nvvp to look at performance.

trws:
Any chance you could throw it my way? I'm leaning toward either taking this version, or saying to go with what it was but with an attribute check to ensure it gets the right behavior with a fallback on a sync memcopy.


---
Sent from Boxer | http://getboxer.com<http://bxr.io/PBI3C>

On March 30, 2017 at 8:21:41 AM PDT, Jason Burmark <notifications@github.com> wrote:

I have a little benchmark that tests these different patterns directly. I have been using nvvp to look at performance.

trws:
Having run the microbenchmark, and poked at this some, I'm not seeing huge performance implications on this either way.  What kind of differences are you seeing that motivate doing this over just using a synchronous memcpy or UM?

MrBurmark:
Indeed the performance implications are not huge.
The main concern is in places where we run multiple small kernels with reductions. Basically in places where launch overhead becomes a large concern.

trws:
That being the case, I personally would rather just have it be a trivially correct cudaMemcpy and only do this if and when there's a clear performance reason to do so.  It's vastly more complicated, and as it stands I'm not sure it's guaranteed to be correct on PowerPC as written.  It should always work on x86 and derivatives, but this is another potential portability issue because of coherence rules on the host access to the flag value.

@rhornung67, @davidbeckingsale, what do you think?

davidbeckingsale:
I don't think we should be implementing these kinds of hacks for very specific use cases - this should be handled in application code, if at all. I'm closing this PR, and we can discuss this more at the RAJA meeting.

Issue #223, opened 2017-03-27
ForallN with omp collapse nowait policies only works with range segment.


MrBurmark:
Using ForallN with policy omp_collapse_nowait_exec does not compile with range stride segments.

davidbeckingsale:
Fixed (partially) by #227 .
Will make a new internal issue for complete support.

PR #227, opened 2017-03-28
Add collapse for RangeStrideSegment

davidbeckingsale:
Addresses #223


davidbeckingsale:
@MrBurmark does this add the functionality you need (from #223)?

trws:
This looks just fine, but is there a reason to have it be specialized on each of the segments? Could it just be a partially specialized on the omp collapse policy while leaving the segment type un-resolved to work for arbitrary segments?

davidbeckingsale:
Yeah, I will give that a try, thanks! I wasn't sure if the partial specialization would be enough.

trws:
I'm not sure I remember all of the ramifications, but IIRC it was only restricted to RangeSegment because we didn't have a generic option available when those classes were written.  Hopefully it should work pretty easily, but if not I'm good with taking this while we look for a way to generalize.

willkill07:
This could be part of the necessary `static_assert()` or `std::enable_if`

```cpp
template <typename SegmentT>
struct is_range_segment {
  static constexpr const bool value =
      std::is_same<RAJA::RangeSegment, SegmentT>::value
      || std::is_same<RAJA::RangeStrideSegment, SegmentT>::value;
  constexpr operator bool() const {
    return value;
  }
};
```

Also, to extract a "default" increment, maybe template specialization would work best?

```cpp
template <typename SegmentT>
Index_type getIncrement(const SegmentT &) {
  return 1;
}

template<>
Index_type getIncrement<RangeStrideSegment>(const RangeStrideSegment &rss) {
  return rss. getStride();
}
```

Code is untested, but should compile/function correctly.

trws:
I’m not sure it would even need that honestly, it should be able to 
decompose into three loops over a range from 0 to 
std::distance(std::begin(r), std::end(r)) and pass the offset 
dereference of the iterator, like body(x[i], (yay for random access 
iterators!).

This would actually allow us greater control if we lose performance 
though, I like the design Will.

davidbeckingsale:
Just implemented this and threw an internal compiler error in gcc-6 😢

trws:
Dang... well, for now we should probably take this as is.  Maybe keep that version in a branch and link from an issue so we can revisit?

Sent from my iPad

On Apr 19, 2017, at 10:30 AM, David Beckingsale <notifications@github.com<mailto:notifications@github.com>> wrote:


Just implemented this and threw an internal compiler error in gcc-6 😢

trws:
@davidbeckingsale this looks good, quick rebase or update before merge?

davidbeckingsale:
Feel free to get this merged, and open an issue for the partial specialization version.

PR #230, opened 2017-04-10
Feature/docker

DavidPoliakoff:
Hey folks, adding a template Dockerfile to the repo.

You can build this pretty easily and start working with RAJA anywhere Docker exists. CUDA programs will only run successfully if the host docker runs on is configured specially, but those programs can compile and link, and all other programs should work.

Zero urgency on this PR, I just use this Dockerfile anyway, thought I'd share it

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM nvidia/cuda:8.0-devel-ubuntu16.04
-MAINTAINER David Poliakoff <poliakoff1@llnl.gov>
+MAINTAINER RAJA Development Team <raja-dev@llnl.gov>
 
 RUN apt-get update -y
 RUN apt-get install -y git cmake gdb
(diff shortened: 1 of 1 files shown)


davidbeckingsale:
This is pretty cool. I would suggest changing the maintainer to raja-dev if we are going to merge this?

DavidPoliakoff:
Fixed, thanks for that

davidbeckingsale:
I'm happy with this, but we should probably get more of a consensus before merging.

DavidPoliakoff:
Sure thing, it's zero urgency.

@trws : I'd guess you're the most likely to have used Docker in the past, anything you see that's iffy?
@rhornung67 : you okay with having another file in the root directory? That's pretty standard practice for people shipping Docker containers, but about half the people downloading the repo might say "what on earth is this?"

trws:
It looks just fine to me.

PR #237, opened 2017-04-27
[WIP/DNM] Introducing transform_reduce

willkill07:
Usage:

`RAJA::transform_reduce<ExecPol>(Iterable, reducers... , TransformFn, ReducerFn)`

This is a flexible paradigm that has the ability to be more efficient than `forall` + `ReduceX` members. The API is still definable, but the current implementation is most likely the most flexible option currently available.

We could "tuple-fy" the reducer variables and arguments for the reduce function which could help with defining RAJA helpers for aggregate reductions with a type list.

I'll vet out some examples but I met with @trws at Salishan and have been in active discussion with him on this. Below is a `transform_reduce` pi example

```cpp
int main() {
  int N = 10000000;
  double pi = 0.0;
  double count = 0.0;
  RAJA::transform_reduce<RAJA::omp_parallel_for_exec>(
    RAJA::RangeSegment{0, N}, pi, count,
    [N](RAJA::Index_type i) {
      double x = (static_cast<double>(i) + 0.5) / N;
      return std::make_tuple(4.0 / (1.0 + x * x), x);
    },
    [](double& pi, double& count, const double p, const double c) {
      pi += p;
      count += c;
    });
  printf("%0.16lf\n", pi);
  printf("%0.16lf\n", count);
  return 0;
}
```


include/RAJA/internal/type_traits.hxx
```diff
@@ -97,11 +97,55 @@ struct function_traits<R (T::*)(Args...) const> {
...
+using extract_arg =
+    std::tuple_element<I, typename function_traits<Fn>::argument_types>;
 
 template <size_t I, typename Fn>
 using extract_arg_t = typename extract_arg<I, Fn>::type;
 
+template <typename Exec>
+struct is_exec_policy
+  : public std::integral_constant<
+      bool, std::is_base_of<RAJA::PolicyBase, Exec>::value> {
```
davidbeckingsale: I'm not sure all the policies extend this, do they?

willkill07: Hmm... After looking they don't, but they either should or the idea of PolicyBase should go away

willkill07:
I'm going to close this because I don't believe this approach fits the needs in of our users and for RAJA's flexibility in the long term.

PR #240, opened 2017-05-01
[Bugfix] Repair additional headers

willkill07:
In addition to resolving the last `hxx` headers, this PR switches all instances of relative includes (e.g. within the same directory) to absolute path includes.

RFC: there is `"cycles.h"` defined under fault_tolerance which no longer exists -- what should we do about this?

### DIFF
--- a/include/RAJA/index/IndexSet.hpp
+++ b/include/RAJA/index/IndexSet.hpp
@@ -55,7 +55,7 @@
 
 #include "RAJA/config.hpp"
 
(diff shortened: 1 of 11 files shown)


willkill07:
Just noticed there are a few places with inconsistencies and macOS doesn't care about case -- will need to issue another commit to resolve.

willkill07:
Once travis goes passes I believe this is good to go :shipit:

PR #241, opened 2017-05-02
[enhancement] Fix operator limits by removing `std::numeric_limits`

willkill07:
`numeric_limits` is great with host-only code, but breaks with GPU code. I hand-rolled limits for all integral types and floating-point types.

* Unsigned Integral types rely on the compiler masking appropriately
* Signed integral types use some bit hack magic -- I tested with numeric_limits before committing
* Floating-point types rely on `cfloat` and `{FLT,DBL,LDBL}_MAX`

I also threw in a fix for a massive warning dump with CUDA 8 on my arch box -- `operator()` within `ForallNPolicy` was not marked `RAJA_SUPPRESS_HD_WARN` (but is now). If this is intentional, I can revert it back.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,6 +76,8 @@ include(cmake/SetupPackages.cmake)
 include(cmake/SetupRajaConfig.cmake)
 # Macros for building executables and libraries
 include (cmake/RAJAMacros.cmake)
+# Sanity check for compiler compatibility
+include (cmake/CompilerCompatibility.cmake)
 
 include_directories(${PROJECT_BINARY_DIR}/include/RAJA)
 include_directories(${PROJECT_BINARY_DIR}/include)
--- a/cmake/CompilerCompatibility.cmake
+++ b/cmake/CompilerCompatibility.cmake
@@ -1,7 +1,9 @@
 include(CheckCXXSourceCompiles)
 
 set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
-set(CMAKE_REQUIRED_FLAGS ${CMAKE_CXX_FLAGS})
+if (NOT MSVC)
+  set(CMAKE_REQUIRED_FLAGS "-std=c++11")
+endif()
 
 CHECK_CXX_SOURCE_COMPILES(
 "#include <type_traits>
--- a/cmake/CompilerCompatibility.cmake
+++ b/cmake/CompilerCompatibility.cmake
@@ -1,0 +1,66 @@
+include(CheckCXXSourceCompiles)
+
+set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+set(CMAKE_REQUIRED_FLAGS ${CMAKE_CXX_FLAGS})
+
+CHECK_CXX_SOURCE_COMPILES(
+"#include <type_traits>
+#include <limits>
+
+template <typename T>
+struct signed_limits {
+  static constexpr T min()
+  {
+    return static_cast<T>(1llu << ((8llu * sizeof(T)) - 1llu));
+  }
+  static constexpr T max()
+  {
+    return static_cast<T>(~(1llu << ((8llu * sizeof(T)) - 1llu)));
+  }
+};
+
+template <typename T>
+struct unsigned_limits {
+  static constexpr T min()
+  {
+    return static_cast<T>(0);
+  }
+  static constexpr T max()
+  {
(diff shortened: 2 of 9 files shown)


willkill07:
Also, do we want to consider adding `--expt-relaxed-constexpr` to CUDA 8 targets? This supposedly gets rid of the aforementioned warnings.

davidbeckingsale:
@willkill07 I don't think we should use the `--expt-relaxed-constexpr` flag if these warnings can be addressed with pragmas. Since this stuff is predominantly in headers, the flag would work for our build, but someone using RAJA without this flag would still get the warning dump.

trws:
So, it's a bit of a PITA, but if the bit twiddling is a worry there's always the option of using SFINAE combined with stdint to get everything from the implementation constants...  Alternatively, we could have a static_assert under an ifdef so it only compiles in host code that compares this with std::numeric_limits if it's enough of a worry.

davidbeckingsale:
That would be good, just some kind of insurance policy against anything weird happening 😄 

@trws what's you opinion on the host-device warnings?

trws:
Still thinking about it.  I prefer solving the warnings wherever practical, and it's making me wonder what the cause is there.

willkill07:
> where was this tested, and where might it break?

I tested on `cab`, my laptop, `rzhasgpu`. Basically x86-64. I did not test under POWER.

The exact check i did was a bunch of `static_assert`s comparing `limit<T>::min()` to `std::numeric_limits<T>::max()`

I'll throw in a macro guard compile-time check for the integral types.

willkill07:
The only two places where the bit hackery would fail is:

1. On architectures with non-power-of-two sized integral types
2. The compiler cannot use `unsigned long long` to perform compile-time constant folding.

davidbeckingsale:
Makes sense, and the asserts look good. Thinking about it some more, we could probably do the check limits stuff as part of the CMake step...

willkill07:
> we could probably do the check limits stuff as part of the CMake step...

Done! Unit test added for compile-time and run-time for all integral types.

davidbeckingsale:
Whoops, I guess I wasn't clear - I meant using the CheckCXXSourceCompiles thing, so we know at "configure" time if the limits are working: https://cmake.org/cmake/help/v3.0/module/CheckCXXSourceCompiles.html

willkill07:
@davidbeckingsale none of the existing files makes sense for me to add to within the `cmake` directory. Should I add a new file which does compiler compatibility checks? Perhaps name is `CompilerCompatibility.cmake`?

davidbeckingsale:
Yeah, that works for me. Thanks!

willkill07:
The irony that the compile-time check doesn't work on travis...

willkill07:
Once travis finishes this should be good to go

willkill07:
Tests pass on my machine with CUDA 8

willkill07:
@trws changes have been made accordingly. If you or someone else could test on manta/ray that would be great -- I only have access to cross compilers for POWER.

DavidPoliakoff:
Confirmed it builds. Any specific tests?

trws:
The unit tests for ForallN almost all exercise that in some way, so it should be gtg.

willkill07:
@DavidPoliakoff I just noticed that the test results you ran were CPU only. Any chance you could also run GPU tests?

DavidPoliakoff:
@willkill07 probably not for a while, in meetings for a bit, but will when time allows

willkill07:
friendly ping @DavidPoliakoff -- btw, I'll be requesting an account on manta/ray when I arrive in ~3 weeks (so I don't need you to do my bidding)

davidbeckingsale:
@willkill07 I'm testing this right now - I had tried earlier which was when I uncovered the bugs fixed by #244 😄

willkill07:
I'll work on this late tomorrow or on Saturday. Thanks for letting me know.

davidbeckingsale:
Yeah no worries, I'll try another compiler tonight to see if it was just a gcc thing.

willkill07:
It cropped up on another test machine of mine -- I'm missing something somewhere :)

willkill07:
@davidbeckingsale I got the tests to pass by removing the extra `RAJA_SUPPRESS_WARN` I added to a few `operator()`. Behavior should be identical to what's on develop. I tested on an x86 CPU + Maxwell GPU. A run on manta/ray would be nice for verification purposes.

davidbeckingsale:
@willkill07 I'll give this a shot today.

davidbeckingsale:
Tests passing, this looks good.

PR #247, opened 2017-05-17
Updated OpenMP4.5 implementation 

mcordery:
This version is based on the latest version of the development branch and works with the IBM LLVM compiler.  Configure is

`cmake -DOpenMP_C_FLAGS="-fopenmp=libomp" \
 -DOpenMP_CXX_FLAGS="-fopenmp=libomp -fopenmp-targets=nvptx64-vidia-gpu -fopenmp-implicit-declare-target" \
 -DCMAKE_BUILD_TYPE=Release \
 -DRAJA_ENABLE_CUDA=Off -DRAJA_ENABLE_OPENMP=On -DRAJA_ENABLE_TARGET_OPENMP=On  \
 -DCMAKE_C_COMPILER=mpiclang-gpu -DCMAKE_CXX_COMPILER=mpiclang++-gpu \
 -DCMAKE_INSTALL_PREFIX=/g/g0/cordery1/Projects/RAJA/raja_install ../`

Note the use of an additional macro, `RAJA_ENABLE_TARGET_OPENMP`. I've tested this with the pi.cxx code using all of the reduction types and they all work. I bumped up the size of the pi test to 512^3 and ran some performance tests with a variety of team sizes. A team size of 256 appeared to work best for ReduceSum:

with critical
16 0.018638
32 0.0118481
64 0.00861594
128 0.00701526
256 0.00668969
512 0.00684503

If I looked at adding in a scoped critical or an omp atomic update neither of those actually helped, even with replacing the call to omp_get_team_num with a version that set a team_id during the copy construct and thus was persistent. 

critical with scope
256 0.00682466

with atomic update
256 0.00708789

with atomic update and tid created in copy construct
256 0.00710864

Compare to using the standard OpenMP omp_reduce policy on P8

thread_num time
160 0.287098
80 0.119766
40 0.178442
20 0.221733

Best time is at about 80 threads. GPU version seems to be about 17x faster.


cmake/SetupCompilers.cmake
```diff
@@ -59,6 +59,9 @@ if (CMAKE_CXX_COMPILER_ID MATCHES GNU)
 elseif (CMAKE_CXX_COMPILER_ID MATCHES AppleClang)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
 elseif (CMAKE_CXX_COMPILER_ID MATCHES Clang)
+  set(CMAKE_CXX_STANDARD 11) # C++11...
+  set(CMAKE_CXX_STANDARD_REQUIRED ON) #...is required...
+  set(CMAKE_CXX_EXTENSIONS OFF)
```
willkill07: Does it fail with these not added? I agree this is The CMake Way ™ but it should be equivalent to `"-std=c++11"`

examples/pi.cxx
```diff
@@ -6,20 +6,31 @@
 
 int main(int RAJA_UNUSED_ARG(argc), char** RAJA_UNUSED_ARG(argv[]))
 {
-  typedef RAJA::seq_reduce reduce_policy;
-  typedef RAJA::seq_exec execute_policy;
+  typedef RAJA::omp_target_reduce reduce_policy;
```
willkill07: There should be an entirely new unit test for target openmp instead of modifying the existing one

include/RAJA/policy/openmp/target_forall.hpp
```diff
@@ -0,0 +1,122 @@
...
+template <typename Iterable, typename Func>
+RAJA_INLINE void forall(const omp_target_parallel_for_exec&,
+                        Iterable&& iter,
+                        Func && loop_body)
+{
+    using body_type = typename std::remove_reference<decltype(loop_body)>::type;
+    auto begin = std::begin(iter);
+    auto end = std::end(iter);
+    auto distance = std::distance(begin, end);
+    int nteams=16;
```
willkill07: Do you want a way to specify the number of teams as part of the execution tag? This is how the CUDA execution policy works and what I'm proposing for the OpenACC execution policy.

cmake/SetupCompilers.cmake
mcordery: I noticed that cmake was doing it's standard sort of thing of adding in additional flags, in particular without doing these sorts of things it ends up adding in -std=gnu++14. This kills that. Just trying to keep odd things from happening, as I've noticed before when it was adding in -std=gnu++11 in addition to -std=c++11

examples/pi.cxx
mcordery: This is true, i was simply modifying the easiest test for me to access to work out funcitonality first. I didn't expect that this would be merged in.

include/RAJA/policy/openmp/target_forall.hpp
mcordery: I think it would be better to have means of passing the number of teams to the forall, either as a tag in the policy or as a parameter.  I didn't know what the plan going forward on such things was or if it had been thought about. I can take take a look at what you've done with CUDA.

cmake/SetupCompilers.cmake
willkill07: >  it was adding in -std=gnu++11 in addition to -std=c++11

I was just wondering about this -- good to know that CMake doesn't do the right thing. In this case, we should probably do it the CMake Way ™. The only problem is some compilers don't have configurations for CMake or CMake doesn't know how to feature test them. We'll probably need to vet that out in our Cmake helpers.

include/RAJA/policy/openmp/target_forall.hpp
willkill07: If `omp_target_parallel_for_exec` had a unsigned integral template value added, the following changes should work:

```cpp
template <size_t TeamSize, typename Iterable, typename Func>
 RAJA_INLINE void forall(const omp_target_parallel_for_exec<TeamSize>&,
                         Iterable&& iter,
                         Func && loop_body)
```

no runtime check is necessary and a reasonable default could be set with the definition of the target tag.

cmake/SetupCompilers.cmake
trws: Note that the default is that we set CMAKE_CXX_STANDARD to 14, because it automatically decays to 11 if that's all that can be found and otherwise gives us some better performance by properly honoring constexpr in places we can't use it in 11.  The extensions haven't hurt us, so we generally just haven't mucked with them.  In terms of making it required, we wanted it to decay from 14-11, so we took that off but have a static_assert that fires in the first header if we don't get at least 11.

willkill07: @trws would adding `set(CMAKE_CXX_EXTENSIONS OFF)` to our default at least prevent `-std=gnu++1[14]` from being emitted? That way, if the compiler supports C++14 it will use it but still decay to C++11?

trws: It would, the results would be limited to c++11 and c++14, no gnu extensions either way.

willkill07:
@mcordery I took a stab at isolating the example (done), adding teams as a template parameter (currently broken), and restructuring the reducers to eliminate code duplication (can confirm it compiles).

I know that the teams template parameter isn't working yet (I need to model the original OpenMP outer + inner policies) , but it should in a couple hours. 

It will need tested, but I don't currently have access to a Power8 machine (e.g. Manta, Ray) to test on. I pushed to the `features/killian4/openmp4-enh-reducers` branch if you want to check it out.

willkill07:
Update: it _should_ work. I decoupled the reducers quite a bit, which should allow for a more modular implementation for other execution policies. 

Quick reference to code is here: https://github.com/LLNL/RAJA/blob/features/killian4/openmp4-enh-reducers/include/RAJA/policy/openmp/target_reduce.hpp

mcordery:
Will, that looks very nice. If I had to add something it would be a check to make sure that Teams matches a value returned from omp_get_num_teams() when you're on on the device. This would just avoid the possible problem that a user would call forall with
teams = n1 but define reducesum, say, with teams = n2. It would also avoid the possible issue of the number of teams used by forall
not necessarily matching the number of teams specified in the forall num_teams clause. As per the OpenMP spec
 
                The number of teams created is implementation defined, but is less than or equal to the value                         18  specified in the num_teams clause. A thread may obtain the number of teams by a call to the                         19  omp_get_num_teams routine.         
In the former case, one could just throw an error. In the latter case, it's a bit tricker. You could throw an error but that doesn't seem like a great idea. A better option might be to internally use a team number that equals omp_get_num_teams, but then you'd have to pick that up while you were on the device and communicate that back to the host for the reduction across teams that works there. I'd ping Tom about this though.
----- Original message -----From: Will Killian <notifications@github.com>To: LLNL/RAJA <RAJA@noreply.github.com>Cc: Matthew Cordery <cordery1@llnl.gov>, Mention <mention@noreply.github.com>Subject: Re: [LLNL/RAJA] Updated OpenMP4.5 implementation (#247)Date: Fri, May 19, 2017 8:01 PM 
Update: it should work. I decoupled the reducers quite a bit, which should allow for a more modular implementation for other execution policies.
Quick reference to code is here: https://github.com/LLNL/RAJA/blob/features/killian4/openmp4-enh-reducers/include/RAJA/policy/openmp/target_reduce.hpp

mcordery:
So I tried mucking about with this today on rzmanta using the latest compiler.
One thing is that there were double negatives in the following (!!host, !!device, !!omp_target...)
    if (!host && !device        && !omp_target_memcpy(reinterpret_cast<void *>(device),                              reinterpret_cast<void *>(host),                              Teams * sizeof(T),                              0,                              0,                              info.deviceID,                              info.hostID)) {      printf("Unable to copy memory from host to device\n");      exit(1);    }  }
  void deviceToHost(Offload_Info &info)  {    if (!host && !device        && !omp_target_memcpy(reinterpret_cast<void *>(host),                              reinterpret_cast<void *>(device),                              Teams * sizeof(T),                              0,                              0,                              info.deviceID,                              info.hostID)) {      printf("Unable to copy memory from device to host\n");      exit(1);    }  }
 
I took those out and tried running. I'm getting a 0 value using ReduceSum with 128 teams
in the pi test code. So far, it seems that when you hit this line in ~TargetReduce
 
        Reducer::apply(val.device[tid], val.value);
 
 nothing seems to happen and it kills off anything that might happen after that (e.g. a print statement)
 
My guess is that it doesn't like the inclusion of static in the apply() templates.
----- Original message -----From: Will Killian <notifications@github.com>To: LLNL/RAJA <RAJA@noreply.github.com>Cc: Matthew Cordery <cordery1@llnl.gov>, Mention <mention@noreply.github.com>Subject: Re: [LLNL/RAJA] Updated OpenMP4.5 implementation (#247)Date: Fri, May 19, 2017 8:01 PM 
Update: it should work. I decoupled the reducers quite a bit, which should allow for a more modular implementation for other execution policies.
Quick reference to code is here: https://github.com/LLNL/RAJA/blob/features/killian4/openmp4-enh-reducers/include/RAJA/policy/openmp/target_reduce.hpp

willkill07:
@mcordery good catch. I found a bug with the transfer in my implementation that [wc]ould cause it to fail (swapped device and host IDs).

r.e. the decision to not continue because of the apply... that doesn't make a whole lot of sense from a translation unit perspective. I'm committing a version to this branch where it instantiates an instance and calls `operator()`

willkill07:
Whoops, I meant to say I updated my branch to handle the logic error and potentially resolve the invocation of the reduce function.

mcordery:
Will, what machine did you run your tests on? When I run the raja-pi-offload tests it complains at setup that I can't copy from host to device but I'm not seeing what's wrong (if anything, yet)
 
----- Original message -----From: Will Killian <notifications@github.com>To: LLNL/RAJA <RAJA@noreply.github.com>Cc: Matthew Cordery <cordery1@llnl.gov>, Mention <mention@noreply.github.com>Subject: Re: [LLNL/RAJA] Updated OpenMP4.5 implementation (#247)Date: Tue, May 23, 2017 6:28 AM 
Whoops, I meant to say I updated my branch to handle the logic error and potentially resolve the invocation of the reduce function.

willkill07:
Right now I face an unfortunate consequence of not having access to rzmanta. I should be able to test/revisit next week when I arrive on site for the summer.

mcordery:
No worries, I was just wondering if you'd successfully tested it. If so, what compiler and version did you use (if you recall). It looks fine and the arguments and types all seem ok so I'm wondering if there's something else wrong.
 
----- Original message -----From: Will Killian <notifications@github.com>To: LLNL/RAJA <RAJA@noreply.github.com>Cc: Matthew Cordery <cordery1@llnl.gov>, Mention <mention@noreply.github.com>Subject: Re: [LLNL/RAJA] Updated OpenMP4.5 implementation (#247)Date: Wed, May 24, 2017 5:52 PM 
Right now I face an unfortunate consequence of not having access to rzmanta. I should be able to test/revisit next week when I arrive on site for the summer.

mcordery:
Ok, so I figured out that problem. It is that omp_target_memcpy returns 0 when it's successful so having if( !omp_target_memcpy()) as a test will fail if the memcpy is successful
 
I got it compiled with clang and it runs but as with my original version it doesn't send the values back properly to this->val on the device which is why I had the ptr2this->val hack in there. I'll send this version to the clang folks though to see if we are any closer to a solution.
 
I tried compiling with XL but hit a problem at link time so I'll have to ping the compiler guys about that.
 
----- Original message -----From: Matthew Cordery/Almaden/IBMTo: reply@reply.github.comCc:Subject: Re: [LLNL/RAJA] Updated OpenMP4.5 implementation (#247)Date: Wed, May 24, 2017 5:58 PM 
No worries, I was just wondering if you'd successfully tested it. If so, what compiler and version did you use (if you recall). It looks fine and the arguments and types all seem ok so I'm wondering if there's something else wrong.
 
----- Original message -----From: Will Killian <notifications@github.com>To: LLNL/RAJA <RAJA@noreply.github.com>Cc: Matthew Cordery <cordery1@llnl.gov>, Mention <mention@noreply.github.com>Subject: Re: [LLNL/RAJA] Updated OpenMP4.5 implementation (#247)Date: Wed, May 24, 2017 5:52 PM 
Right now I face an unfortunate consequence of not having access to rzmanta. I should be able to test/revisit next week when I arrive on site for the summer.

willkill07:
You're welcome to push to my branch or integrate here as desired. I only thought `ptr2this` was used to avoid the tricky `const_cast<SelfType*>(this)->` shenanigans in const member functions to allow for self mutation. Storing an extra pointer as a workaround isn't a deal breaker so long as it works.

Interesting that XL fails during link. Do you know what it's complaining about? Feel free to email me directly (or raja-dev) if you don't want things showing up publicly on GitHub.

davidbeckingsale:
This PR is outdated by #255

PR #249, opened 2017-05-19
[Enhancement] Stricter CXX standard flag emission

willkill07:
RAJA still defaulted to potentially emitting `-std=gnu++11` or `-std=gnu++14`

This PR resolves this issue and will emit `-std=c++14` if the compiler supports the C++14 standard and emits `-std=c++11` otherwise.

I added in a compiler flag check so even if an older CMake (like 3.3) is fed a newer compiler without any feature tests (such as Intel 17.0), it will still emit the appropriate standard flag.

I wanted to remove `LANGUAGE C` from the `project` configuration (so we would never need to specify `CMAKE_C_COMPILER`, but googletest requires it because of the check for `find_package(Threads)`. A potential workaround could work by setting `CMAKE_C_COMPILER` to `CMAKE_CXX_COMPILER`, but that just feels a little too dirty.


davidbeckingsale:
@willkill07  does this add the `std` flag twice?

davidbeckingsale:
Nevermind, I think I have answered my own question. It doesn't right now, but if CMake starts correctly providing the flag corresponding to the `CMAKE_CXX_STANDARD` it might.

willkill07:
Yeah, it doesn't right now, and currently doesn't with CMake 3.8. When the time comes, we can always add a check on the CMake version

davidbeckingsale:
Nice, thanks!

PR #250, opened 2017-05-22
Add typed layout

trws:
There may be some more refactoring to be done here, factoring out TypedView for example, but this adds a strongly typed Layout to assist with the kripke update.  It also modifies the strongly typed index types such that a reference can be taken to their enclosed Index_type value.


davidbeckingsale:
@trws should we merge this?

trws:
Go for it, this should be GTG.

PR #251, opened 2017-05-23
partial rework to move multipolicy

trws:
This is a partial rework adding a layer to allow things like the chai integration to operate after all wrapper policies have been resolved.  I'm not sure I love the way it works, but it adds another layer between the user interface layer and impl such that wrapper policies live in the RAJA::wrap namespace, and are all resolved by the time RAJA::wrap::forall generic is invoked.  If this solution is selected, there's more to do, but since I did this much to see if it was going to be feasible at all it seemed worth putting it up.

### DIFF
--- a/include/RAJA/pattern/forall.hpp
+++ b/include/RAJA/pattern/forall.hpp
@@ -119,6 +119,67 @@ namespace RAJA
 /*!
  ******************************************************************************
  *
+ * \brief The RAJA::wrap layer unwraps dynamic policies before dispatch
+ *
+ ******************************************************************************
+ */
+namespace wrap {
+
+/*!
+ ******************************************************************************
+ *
+ * \brief Generic dispatch over containers with a value-based policy
+ *
+ ******************************************************************************
+ */
+template <typename EXEC_POLICY_T, typename Container, typename LOOP_BODY>
+RAJA_INLINE void forall(EXEC_POLICY_T&& p, Container&& c, LOOP_BODY loop_body)
+{
+  impl::forall(std::forward<EXEC_POLICY_T>(p), std::forward<Container>(c), loop_body);
+}
+
+/*!
+ ******************************************************************************
+ *
+ * \brief Generic dispatch over containers with icount
+ *
+ ******************************************************************************
+ */
+template <typename EXEC_POLICY_T, typename LOOP_BODY>
+RAJA_INLINE void forall_Icount(EXEC_POLICY_T&&p, const IndexSet& c, LOOP_BODY loop_body)
+{
+
+  impl::forall_Icount(p, c, loop_body);
+}
+
+/*!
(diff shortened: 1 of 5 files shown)


davidbeckingsale:
So the idea would be to put the Chai calls into the `wrap` namespace version? If so, looks fine to me.

trws:
Yes, they would go into the generic forall in the wrap namespace.  We could do this same thing at the outermost level, but it would mean that all of the outermost foralls would have to invoke the generic outer forall to get everything to resolve, and we still wouldn't be able to tell which invocation it is.  This way, I think we can make it so nothing gets to the RAJA::wrap::forall generic until multipolicy and similar are done flipping things around.

davidbeckingsale:
Yeah I think that makes sense. Having a single point things pass through at a lower level is good.

include/RAJA/pattern/forall.hpp
```diff
@@ -127,7 +188,7 @@ template <typename EXEC_POLICY_T, typename LOOP_BODY>
 RAJA_INLINE void forall_Icount(const IndexSet& c, LOOP_BODY loop_body)
```
rhornung67: Do we have a clear policy when we should use reference vs. universal reference for method args? We probably should and document it in the coding guidelines that I will eventually get back to...

trws:
We do not, and that's something that needs to be fixed actually.  It's a 
touch-everything kind of change but normalizing it would be a very good 
thing...

rhornung67:
I'll add it to my list of coding guidelines we need....Thanks.

davidbeckingsale:
@trws is this ready to merge?

trws:
It is if it's how we want to do this.  I'm good with it, but it would be good to know if it helps with the multipolicy issue you were having @davidbeckingsale.  If not, it would be nice to avoid the extra layer. =/

davidbeckingsale:
The only other way I saw to solve my problem was to actually have CHAI code in the multi policy. That might end up being the best solution, because how would this approach work with nested multipolicy?

trws:
The chaining should still just work actually, it should only hit the generic forall in the wrap layer after all wrappers have been resolved.  If not, then yeah that's probably the better solution.

RAJA::forall<generic>->RAJA::wrap::forall<multipolicy<range,multipolicy>>->RAJA::wrap::forall<multipolicy>->RAJA::wrap::forall<generic>->RAJA::impl::forall

davidbeckingsale:
@trws let's hold on this a little longer, I'll try my branch against these changes and see if it still works as we want.

trws:
Fine by me.  Like I said, I'm ok with this code but want to know it actually addresses a problem, and hopefully doesn't adversely effect performance, before we actually merge it.

davidbeckingsale:
Closing because this got pulled into #257

PR #257, opened 2017-06-08
Add CHAI integration

davidbeckingsale:
This makes a number of small changes to allow integration with CHAI directly from RAJA.

- A new `wrap` namespace is added to allow all `forall` calls to pass through a single point before being dispatched to an implementation function (thanks @trws).
- A new enum class, `Platform`, has been added to PolicyBase. This is used to map RAJA execution policies to CHAI spaces.


include/RAJA/pattern/forall.hpp
```diff
@@ -119,6 +127,97 @@ namespace RAJA
...
+template <typename EXEC_POLICY_T, typename Container, typename LOOP_BODY>
+RAJA_INLINE void forall(EXEC_POLICY_T&& p, Container&& c, LOOP_BODY loop_body)
+{
+#if defined(RAJA_ENABLE_CHAI)
+  chai::ArrayManager* rm = chai::ArrayManager::getInstance();
+  auto space = detail::get_space< typename std::remove_reference<decltype(p)>::type >::value;
+  rm->setExecutionSpace(space);
+#endif
+
+  typename std::remove_reference<decltype(loop_body)>::type body = loop_body;
```
davidbeckingsale: This is required to ensure the copy-constructor is run once the chai::ExecutionSpace has been set.

include/RAJA/policy/PolicyBase.hpp
```diff
@@ -69,16 +69,26 @@ enum class Pattern {
...
+};
+
 struct PolicyBase {
 };
 
 template <Policy P = Policy::undefined,
           Launch L = Launch::undefined,
-          Pattern Pat = Pattern::undefined>
+          Pattern Pat = Pattern::undefined,
+          Platform Plat = Platform::host
```
davidbeckingsale: Default `Platform` is always `host`.

include/RAJA/pattern/forall.hpp
DavidPoliakoff: This makes it more important for us to make sure we only hit this layer once. I think we meet that requirement, but wanted to ask and make sure

davidbeckingsale: Definitely. In fact, we probably want to go through and ensure we are using `&&`  as much as possible elsewhere. We could even move this inside the `ifdef` so that a copy is only made when CHAI is used.

include/RAJA/policy/PolicyBase.hpp
DavidPoliakoff: This seems iffy to me in a broader sense. omp_parallel_for_exec might want to default to HBM on a KNL, or it might depend on the problem in question. I'd highly recommend testing this on an application on KNL, and working out of HBM on it, while still hitting main memory on a Xeon (non-Phi) system. 

I also know schemas are out, but if an app has to "pick exactly one memory space" for a policy, we don't work on KNL for a lot of applications.

Answer might be "that's fine," but it's worth thinking about

davidbeckingsale: Yeah this was motivated by laziness :) 

The comment you raise though makes having a default sensible. Otherwise we have to override on a policy by policy basis.

DavidPoliakoff: I'd look at what you do with omp_.*_exec, though. It's a tough thing to work with. Also, don't know if OMP4 goes through omp_.*_exec or is its own thing, but might get interesting there. If you've looked at the tradeoffs and decided tying policies to execution spaces is fine, that's totally kosher, I just want to make sure the tradeoff is a conscious decision

trws:
I like this overall, but I'm not 100% on the chai dependency.  Is there a reason this needs to find chai in cmake for the tests rather than have tests of the raja functionality that enables chai?  I haven't dug through all the way, does this explicitly call chai routines rather than letting chai register something to be called from a general list or similar?

davidbeckingsale:
Hey @trws, thanks for the comments. I'm not entirely sure what you mean w.r.t. the chai dependency of the tests. Would you prefer these to be merged with the regular tests?

trws:
I'm arguing with myself over whether we want to take this into raja such 
that when built for chai raja has a chai dependency, or pay the extra 
effort up-front to make the raja end of that connection independent of 
chai.  Same for things like the ManagedArrayView, which is really a chai 
thing rather than a RAJA thing.

I like what you've done, it just feels really odd to have chai-specific 
code in raja that causes what feels like a backwards dependency ordering 
between them.

Whether it's now or later, we're going to want to have this be a more 
general hook mechanism, probably built on something like the LLVM 
Registry (http://llvm.org/docs/doxygen/html/Registry_8h_source.html) so 
we can factor our code reasonably.

DavidPoliakoff:
To follow up on that, ManagedArrayView could also be useful in other places, say there were another parallel programming framework developed at another national lab, they could pretty easily take advantage of the CHAI mechanisms to manage data movement within their own codes, and they very well might want to have a ManagedArrayView to do it.

davidbeckingsale:
Okay, @trws, that makes sense now. I would indeed prefer CHAI build with RAJA integration, rather than the other way around. We can definitely move in that direction.

The main point of this PR was to have preliminary support, and to avoid having code teams having to work from a forked RAJA repository.

include/RAJA/pattern/forall.hpp
```diff
@@ -119,6 +127,97 @@ namespace RAJA
...
+  impl::forall(std::forward<EXEC_POLICY_T>(p), std::forward<Container>(c), body);
+
+#if defined(RAJA_ENABLE_CHAI)
+  rm->setExecutionSpace(chai::NONE);
```
trws: I know this is normal for chai, but why NONE rather than host?

include/RAJA/policy/PolicyBase.hpp
trws: It is possible that the OpenMP policy may at some point support device vs host selection at runtime, but if it does it will still be a function of the policy.

trws:
Rebase to develop and this should be GTG @davidbeckingsale.

davidbeckingsale:
Sounds good! I'll tidy it up then I was thinking of actually squashing the merge to keep it tidier.

One thing I noticed was that I've managed to pull in the ReduceSum CUDA breakage, so I need to get rid of that.

trws:
Yeah... that's the warning suppression in the ForallNPolicy file.  I'm trying to track down why that's happening, but it persists in eluding me.

include/RAJA/internal/ForallNPolicy.hpp
```diff
@@ -100,7 +100,6 @@ struct ForallN_BindFirstArg_HostDevice {
   {
   }
 
-  RAJA_SUPPRESS_HD_WARN
```
davidbeckingsale: @trws was this the line you think was causing problems?

trws: The three below actually.  Note in the diff they're added in this PR.

davidbeckingsale:
I've removed those too and it still breaks  - and not in any CHAI code, in the RAJA reduction test :(

davidbeckingsale:
@ajkunen can you try this out when you get a chance? I'd like a little more sanity checking before I merge this....

davidbeckingsale:
@trws @ajkunen this is ready to merge, I'll do so when (if) the tests pass.

trws:
Sounds good, squash merge is the plan right?  Also, where are we on bamboo? The icc compiler timeout pain is making me eye travis a bit suspiciously right now.

davidbeckingsale:
I can take a look at the bamboo stuff today. It's been on the TODO list for a while. I think we have agents ready to go.

davidbeckingsale:
And yup, squash merge, it got kind of messy.

PR #258, opened 2017-06-08
adding config.hpp wherever it should be and is not

trws:
Yay for shell one-liners: `vim $(grep -r 'RAJA_[A-Z]' -l include | grep -v config.hpp | xargs grep -L 'config.hpp')`.

Only ended up with two false positives. =)


davidbeckingsale:
Nice 👍

DavidPoliakoff:
Oh that is gorgeous

trws:
Odd travis push-back, apt-get failed on one specific test for no apparent reason, but a re-run passes.

@davidbeckingsale :shipit:?

PR #259, opened 2017-06-09
Adding Intel build (hopefully)

davidbeckingsale:


davidbeckingsale:
Thanks to @jeffhammond it looks like we can now build with Intel on Travis! 😄

trws:
Very nice!

davidbeckingsale:
It works! So I think with an approval we can just merge this. It might be worth forking that script and then using it to support multiple Intel versions, but that can be a future PR. This should be enough to catch any Intel snafus for now.

jeffhammond:
You may want to subscribe to https://github.com/nemequ/icc-travis/issues/10 to see if/when progress is made on version support.  I'll try to do the tedious part at least, since I have every version of the compiler installed locally.

davidbeckingsale:
Nice! Subscribed.

trws:
Thanks @jeffhammond!

PR #260, opened 2017-06-13
[Bugfix] Remove Index_type in forall machinery; Add iterator pair forall policy

willkill07:
Per the email from Peter Robinson @robinson96 regarding some hardcoded `Index_type` code. This lifts that away as much as possible by relying on `auto` and `decltype(var)` where appropriate.

Unfortunately, I don't know how well this will work given that `std::distance` returns `std::ptrdiff_t` which on most systems is just a `long`. With the iCount variants, this works as expected, using the type of the passed in offset value as the type to use for the lambda.

This also changes Index_type for CUDA dim/block/id to unsigned int, as they cannot be negative and shouldn't exceed 4B elements in length.


include/RAJA/pattern/forall.hpp
```diff
@@ -187,15 +183,76 @@ RAJA_INLINE void forall(EXEC_POLICY_T&& p, Container&& c, LOOP_BODY loop_body)
...
+ *
+ * \brief Generic dispatch over iterators with icount
+ *
+ ******************************************************************************
+ */
+template <typename EXEC_POLICY_T,
+          typename Iterator,
+          typename LOOP_BODY,
+          typename IndexType>
+RAJA_INLINE typename std::
```
davidbeckingsale: This formatting is weird.

include/RAJA/internal/Span.hpp
```diff
@@ -0,0 +1,86 @@
...
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#include <type_traits>
+#include "RAJA/internal/type_traits.hpp"
+
+namespace RAJA
+{
+
+template <typename ValueType, typename IndexType>
+struct Span {
```
davidbeckingsale: If this is "internal" code, we might want to consider putting it in an internal namespace - perhaps `impl`.

include/RAJA/internal/type_traits.hpp
```diff
@@ -65,6 +66,14 @@ namespace detail
 {
 
 template <typename T>
+struct is_random_access_iterator
```
davidbeckingsale: Nice!

include/RAJA/pattern/forall.hpp
willkill07: That's what the default clang-format did -- I'll manually fix it.

davidbeckingsale: Great 👍

include/RAJA/internal/Span.hpp
willkill07: Ah, good catch. Should Span be internal or available to end users?

davidbeckingsale: I guess that would be your decision. I think the interface you provided is nice, and means that it's only required internally. So `impl` would be good.

willkill07:
Regarding SFINAE -- This only happens in outer policy invocations, so I'm not too concerned about compilation times. I'll run some compiler diagnostics with this branch compared to `develop`

davidbeckingsale:
Perfect, thanks.

include/RAJA/policy/sequential/forall.hpp
```diff
@@ -104,17 +104,17 @@ RAJA_INLINE void forall(const PolicyBase &, Iterable &&iter, Func &&loop_body)
   }
 }
 
-template <typename Iterable, typename Func>
+template <typename Iterable, typename Func, typename IndexType>
```
davidbeckingsale: Is having template parameter names shadow typedefs a bad idea? I remember this happened in the type traits header....

willkill07: I intentionally called it IndexType instead of Index_type

davidbeckingsale: Oh duh. I obviously can't read. Sorry!

willkill07:
Note from RAJA meeting: Integrate Tom's RangeSegment

davidbeckingsale:
Hey @willkill07 it looks like this actually breaks with CUDA. Can you take a look?

willkill07:
@davidbeckingsale This is fixed. Had an extra `>` at the end of a template list >.<

PR #263, opened 2017-06-23
add header and namespace for getenv and abort

jeffhammond:
this was previously addressed by https://github.com/LLNL/RAJA/pull/231
but that used the C header.  subsequently, that change was undone.

this fix uses the correct C++ solution, which means `cstdlib` instead of
`stdlib.h` and prepending `getenv` and `abort` with `std::`


jeffhammond:
This change is *required* to compile RAJA with Intel 18 update 1 (`icpc (ICC) 18.0.0 20170320`) using the default flags RAJA sets.

trws:
@jeffhammond, I'm good with this, it just needed a quick merge.  I put the necessary change on the bugfix/jeffhammond/abort-and-getenv branch if you just want to snag the merge commit over.

jeffhammond:
I'm not sure what "if you just want to snag the merge commit over" means.  Do you need me to do anything to merge?  I have no trouble cherry-picking and such locally, so nothing I'm doing needs the merge.

trws:
All I meant was, I made the changes in the named branch, you can 
cherry-pick the commit from there if you like, or you can merge from 
LLNL/RAJA and move the changes to the new location for the defines 
header.

willkill07:
I performed the merge to bring this up to develop. I also changed the usage of `NULL` to the more appropriate `nullptr`. Once the CI tests pass this should be good to go :shipit:

davidbeckingsale:
Yes, excellent. We can merge this once the tests pass.

willkill07:
@jeffhammond @davidbeckingsale 

The Intel Compiler Travis CI job is currently failing because the compiler does not install :( Otherwise, this branch should be good to go.

davidbeckingsale:
Bummer. I trust you, but we should fix the Intel stuff... Otherwise I have to change the repo settings to be able to merge branches that don't have passing checks.

@jeffhammond do you know of any recent changes in the script or installer that would break this?

jeffhammond:
No.  I don't know any changes here and https://github.com/nemequ/icc-travis/commits/master shows nothing since August 2016.

willkill07:
What's weird is the Intel Compiler has been working just fine on other PRs. Is it possible to merge without the build fully passing? I tested v18 and v17 on quartz to verify it worked.

jeffhammond:
Maybe move icc to xfail in Travis so the must-pass-CI bit is cleared?

PR #272, opened 2017-07-11
Add header and namespace for getenv and abort [updated]

willkill07:
This is the second attempt at this PR (trying to get Intel build to not error in CI). Trying a clean branch with the commits merged from @jeffhammond 's initial PR

Closes #263 

this was previously addressed by https://github.com/LLNL/RAJA/pull/231
but that used the C header.  subsequently, that change was undone.

this fix uses the correct C++ solution, which means `cstdlib` instead of
`stdlib.h` and prepending `getenv` and `abort` with `std::`

### DIFF
--- a/include/RAJA/util/defines.hpp
+++ b/include/RAJA/util/defines.hpp
@@ -55,7 +55,7 @@
 
 #include "RAJA/config.hpp"
 
-#include <stdlib.h>
+#include <cstdlib>
 #include <stdexcept>
 
 //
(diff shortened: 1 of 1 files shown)


willkill07:
@davidbeckingsale the same code on a different PR passes

davidbeckingsale:
Hey @willkill07 - thanks for this. I had a suspicion it might work on a different branch, since all the others were passing. I'm not sure if something was screwed up since @jeffhammond 's branch was based on `master` not `develop`.

PR #273, opened 2017-07-11
Feature/ajk/projlayout

ajkunen:
Added the ability for a Layout to be a projection... that is: one ore more of the dimensions can have a size of zero.

This didn't change the complexity of operator() or toIndices(), but it did change the size complexity from 3n to 4n integers.

@davidbeckingsale Would this work with the OffsetLayout?


davidbeckingsale:
Do we want this to be in `Layout`, or should it be in a separate class `ProjectedLayout`. If the `OffsetLayout` tests still pass then I assume it works 😄 

I don't have any problems with this PR, but I think we should have some good documentation for it.

ajkunen:
I would prefer not to make a ProjectedLayout, since it's functionally the same.  The only change is that it supports dim sizes of zero, which is really just a change in construction.

trws:
I'm good with this approach. My only immediate question is whether the mods array argument to the constructor from arrays was removed intentionally.


---
Sent from VMware Boxer

On July 11, 2017 at 12:11:10 PM PDT, Adam Kunen <notifications@github.com> wrote:

I would prefer not to make a ProjectedLayout, since it's functionally the same. The only change is that it supports dim sizes of zero, which is really just a change in construction.

ajkunen:
@trws The mods array was being passed in but not used.  I assumed this was a vestigial tail that handn't fallen off yet.  Was there a use for it?

PR #279, opened 2017-07-13
[WIP]foralln reimagined

trws:
Work certainly remains to be done here, but this is far enough along that it seems worth getting comments on it.  The new forallN syntax, now nested::forall, is implemented and working for serial code with permutation and collapsing constructs.  Main items remaining to be resolved:

- [x]  CUDA support
- [x]  Fix travis (also related to how we pull in metal)
- [x] Fix cmake to add the metal workaround define for compilers with poor C++11/14 support for compilers where it does not autodetect it
- [x]  OpenMP policies
- [x]  Reduction support, specifically copy-construction of the reference-captured inner functor to create a privatized version per-thread when necessary
- [ ]  More thorough testing
- [x]  Outer interface taking a tuple or array of segments and a parameter pack of policies to better resemble the original
- [ ]  Documentation
- [x]  Tiling Support


test/unit/nested.cpp
```diff
@@ -0,0 +1,31 @@
...
+                       [=](Index_type i, Index_type j) {
+                         std::cout << i << " " << j << std::endl;
+                       });
+}
+
+TEST(Nested, collapse)
+{
+  using namespace RAJA::nested;
+  using Index_type = RAJA::Index_type;
+  using pol = Policy<Collapse<RAJA::seq_exec, CFor<1>, CFor<0>>>;
```
DavidPoliakoff: Nitpicky, but CFor is a little opaque. Is that "CollapseFor<SomeIndex>"

The mechanics look good, my hope in this PR is to be the fool in the foolproofing test.

rhornung67: @DavidPoliakoff good suggestion! I mean the CFor thing, not the fool part :-)

DavidPoliakoff: @rhornung67 : we have a contributor's code now, you're not allowed to impugn my excellent credentials at playing the fool.

Somewhat more seriously, I do think before this is merged we should get a lot of us who don't use forallN to try to write code using this, and document the experience. I think forallN is great but a little scary, and some "here's how to get started with it" docs could be good, wind up in the tutorial, etc.

That's outside the scope of the review, but maybe @trws when you think there's enough to play with (tiling, CUDA) ping me and I'll test it.

willkill07: As someone who honestly struggled with forallN a lot last summer, I already see substantial improvement. The biggest one for me is writing things out the way you want them to happen. With forallN tiling and permutation didn't always make the most sense. You had to imagine the structure by applying them in your head first.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "external/metal"]
```
willkill07: Do we want to go down the submodule root? If so, should google test be pulled in a similar manner?

include/RAJA/pattern/nested.hpp
```diff
@@ -0,0 +1,228 @@
...
+    metal::apply<metal::lambda<RAJA::util::tuple>, get_for_index_types<Seq>>;
+}
+
+template <metal::int_ ArgumentId,
+          typename Pol,
+          typename IndexType = RAJA::Index_type>
+struct For : public internal::ForList,
+             public internal::ForTraitBase<ArgumentId, IndexType> {
+  using as_for_list = metal::list<For<ArgumentId, Pol, IndexType>>;
+  // TODO: add static_assert for valid policy in Pol
```
willkill07: When you say valid policy, will we know the previously applied policies as well? For example, omp_parallel_for_exec within a cuda_exec makes for a bad time. We could theoretically check for that :)

test/unit/nested.cpp
willkill07: If we know we are going to collapse indices, why not just specify the indices? or do they need to be type lists only?

`Policy<Collapse<seq_exec, 1, 0>>` looks a little cleaner to me, but maybe I'm biased?

rhornung67: @willkill07 and @DavidPoliakoff to onboard Arturo who started on Monday, I'm having him build up some code examples to introduce folks to RAJA concepts, including forallN.  We will build a tutorial document based on these examples.  When he has something to share, we can review it and make suggestions.

ajkunen: For OMP collapse(N) clause, it sufficient to have just the loop indices.

But for something more complex, such as CUDA, we need to be able to specify per-loop options, like how indices map to threads/blocks

ajkunen: Also, we need to be able to specify the type of each index... so even for OMP we need that (so I take back my other statement)

.gitmodules
trws: I'm not sure to be honest, this was a convenient way to do it for now.  It would probably make more sense to have both gtest and metal work the same way whichever way we choose.

include/RAJA/pattern/nested.hpp
trws: We could, but that will have to be checked out in the nested::forall or similar.  Here there's no sense of other policies, but it needs to be a valid RAJA::forall policy.

test/unit/nested.cpp
trws: Yeah, that code is shorthand for:

```c++
using pol = Policy<Collapse<RAJA::seq_exec, CFor<1,Index_type>, CFor<0, Index_type>>>;
```

We need the types to be able to build up the index tuple from the policy.

trws: I'm not sure what the best thing is to do for that name.  As of now, it does stand for "CollapseFor" but I'm trying to figure out if we can get away with just having one kind of For and apply it to both Collapse and the top level Policy to avoid having to have it be different.  Lacking that, I went with CFor to keep it terse for the first cut, but having it be more specific wouldn't be a terrible thing.

ajkunen: @willkill07 what are you talking about? the old forallN policies were trivial to work with :)
jk... I'm glad they are clearer to you.

Also, this will make it possible to use ordinary RAJA exec polices on the tiling loops, something @tepperly has been asking for.

DavidPoliakoff: The "just ask @ajkunen" method _was_ trivial, but we quickly found ourselves running out of @ajkunen's

trws: This has been pretty much blown apart and put back together (this being the For setup).  Thoughts?

ajkunen:
I need to take a cocktail of immodium and tums before I try and grock the implementation, but I really like the user interface.

.gitmodules
rhornung67: @willkill07 and @trws if we change over to BLT, which @davidbeckingsale and I are thinking we should (I'm working this into the perf suite now), things like googletest come along for free.

rhornung67:
@ajkunen you should include a cheez-whiz chaser with that cocktail...

.gitmodules
davidbeckingsale: Googletest comes for free, but BLT itself would be a submodule. I guess the question is whether we might need to make RAJA-specific changes to anything that couldn't be upstreamed?

trws: I don't think so, but I can imagine a time where we want to pin a given RAJA version on a specific metal version.  Specifically if there comes a time when we need to support a compiler they have stopped supporting.

ajkunen: @trws agreed.
I have become a big fan of submodules for this reason... as long as the build system also allows pointing at a user-specified path for that library

davidbeckingsale:
👍  I like the latest commits!

trws:
This most recent commit adds a first cuda test.  It currently builds, but it does not pass, and likely will not until a CUDA adapter is added to unpack the nested LoopData construct and pass it through to the kernel.  Either way, the fact that both cuda 8 and cuda 9 can compile this stuff seems encouraging.

.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "external/metal"]
-	path = external/metal
-	url = https://github.com/brunocodutra/metal
```
ajkunen: OMG

DavidPoliakoff: !!!

@trws , you're a deeply alarming person, you know that? I mean that as the highest form of compliment. Seriously nice though

include/camp/number.hpp
```diff
@@ -0,0 +1,173 @@
...
+using make_int_seq_t = typename make_int_seq<T, Upper>::type;
+
+// TODO: document
+template <typename Cond, typename Then, typename Else>
+struct if_ {
+  static_assert(Cond::value, "got true with a false val somehow");
+  using type = Then;
+};
+template <typename Then, typename Else>
+struct if_<std::false_type, Then, Else> {
```
willkill07: Do we want to restrict to `std::false_type` or allow anything that quacks like `std::integral_constant<bool>`?

trws:
anything like i_c bool, or num or...  the overloads below that attempt 
to cover some of that ground, but I'm not sure it gets them all.

include/RAJA/internal/LegacyCompatibility.hpp
```diff
@@ -186,12 +164,12 @@ RAJA_HOST_DEVICE RAJA_INLINE constexpr auto foldl(Op&& operation,
                                                   Rest&&... rest) ->
     typename foldl_impl<Op, Arg1, Arg2, Arg3, Rest...>::Ret
 {
-  return foldl(forward<Op&&>(operation),
-               forward<Op&&>(
-                   operation)(forward<Op&&>(operation)(forward<Arg1&&>(arg1),
-                                                       forward<Arg2&&>(arg2)),
-                              forward<Arg3&&>(arg3)),
-               forward<Rest&&>(rest)...);
+  return foldl(camp::forward<Op&&>(operation),
```
willkill07: I believe the additional `&&` in the template is unnecessary. The arguments are already specified as universal references and reference collapsing rules would keep all of the types the same.

include/RAJA/pattern/nested.hpp
```diff
@@ -0,0 +1,318 @@
...
+{
+namespace nested
+{
+
+namespace internal
+{
+
+template <typename T>
+using remove_all_t =
+    typename std::remove_cv<typename std::remove_reference<T>::type>::type;
```
willkill07: This could be used in many, many other places. Maybe a move to `camp` makes more sense. I can see users wanting this. It's a less-specialized `decay` :)

include/RAJA/pattern/nested.hpp
```diff
@@ -0,0 +1,318 @@
...
+
+using has_for_list = typename camp::bind_front<std::is_base_of, ForList>::type;
+
+template <typename T>
+using get_for_list = typename T::as_for_list;
+
+template <typename Seq>
+using get_for_policies = typename camp::flatten<typename camp::transform<
+    get_for_list,
+    typename camp::filter_l<has_for_list, Seq>::type>::type>::type;
```
willkill07: 👍 This is awesome

include/RAJA/pattern/nested.hpp
```diff
@@ -0,0 +1,318 @@
...
+    camp::list<>,
+    camp::as_list<camp::idx_seq_from_t<IndexTypes>>>::type;
+
+template <typename Iterator>
+struct iterable_value_type_getter {
+  using type = typename Iterator::iterator::value_type;
+};
+template <>
+struct iterable_value_type_getter<IndexSet> {
+  // TODO: when static indexset drops, specialize properly
```
willkill07: I believe this has happened.

include/RAJA/pattern/nested.hpp
```diff
@@ -0,0 +1,318 @@
...
+}
+
+template <typename Pol, typename SegmentTuple, typename Body>
+RAJA_INLINE void forall(const Pol &p, const SegmentTuple &st, const Body &b)
+{
+#ifdef RAJA_ENABLE_CUDA
+  // this call should be moved into a cuda file
+  // but must be made before loop_body is copied
+  beforeCudaKernelLaunch();
+#endif
```
willkill07: Should this only be done if Pol is a CUDA policy?

include/RAJA/pattern/nested.hpp
willkill07: actually, this is already `camp::types::decay_t`

include/camp/tuple.hpp
```diff
@@ -0,0 +1,253 @@
...
+}
+
+template <typename Fn, typename TupleLike>
+CAMP_HOST_DEVICE constexpr auto invoke(TupleLike&& t, Fn&& f) -> decltype(
+    invoke_with_order(t,
+                      f,
+                      camp::make_idx_seq_t<tuple_size<TupleLike>::value>{}))
+{
+  return invoke_with_order(
+      t, f, camp::make_idx_seq_t<tuple_size<TupleLike>::value>{});
```
willkill07: Universal references should be perfectly forwarded

include/RAJA/pattern/forall.hpp
```diff
@@ -125,10 +125,62 @@ namespace RAJA
...
+};
+DefineTypeTraitFromConcept(is_privatizable, Privatizable);
+
+template <typename T>
+struct PrivatizableWrapper
+    : DefineConcept(
+          concepts::val<T>().privatize(),
+          concepts::val<T>().re_wrap(
+              concepts::val<decltype(concepts::val<T>().privatize())>())) {
+};
```
willkill07: this looks slick 👍

include/camp/helpers.hpp
```diff
@@ -184,6 +184,27 @@ CAMP_HOST_DEVICE constexpr T&& forward(type::ref::rem<T>&& t) noexcept
...
+  T temp{std::move(t1)};
+  t1 = std::move(t2);
+  t2 = std::move(temp);
+#else
+  using std::swap;
+  swap(t1, t2);
+#endif
+}
+
+template <typename T, typename = decltype(sink(swap(val<T>(), val<T>())))>
```
willkill07: should this be conditionally visible through `enable_if<Swappable<T>>` ?

trws:
Yup. Right now there's a dependency problem, but I'm planning to refactor it into its own file that can depend on concepts.hpp to do that.


---
Sent from VMware Boxer

On September 11, 2017 at 1:50:30 PM HST, Will Killian <notifications@github.com> wrote:

In include/camp/helpers.hpp<https://github.com/LLNL/RAJA/pull/279#discussion_r138221393>:

> @@ -184,6 +184,27 @@ CAMP_HOST_DEVICE constexpr T&& forward(type::ref::rem<T>&& t) noexcept
 {
   return static_cast<T&&>(t);
 }
+
+template <typename T>
+CAMP_HOST_DEVICE void safe_swap(T& t1, T& t2)
+{
+#if defined(__CUDA_ARCH__)
+  T temp{std::move(t1)};
+  t1 = std::move(t2);
+  t2 = std::move(temp);
+#else
+  using std::swap;
+  swap(t1, t2);
+#endif
+}
+
+template <typename T, typename = decltype(sink(swap(val<T>(), val<T>())))>


should this be conditionally visible through enable_if<Swappable<T>> ?

trws:
Woah! what happened here? I did not mean to close this.

trws:
Develop is now merged in, and all of the tests I can run locally succeed.  The cuda side needs further testing, but assuming everything passes it may be time to get this in as a first cut we can refine with smaller subsequent PRs.

trws:
I'm not sure what the issue is here anymore.  The log length for the coverage target exceeds the 4mb travis limit.  @willkill07, @davidbeckingsale, any idea how to make it be quiet without completely losing diagnostics?

willkill07:
Do we really need the output from the coverage?

The offending line is:

```bash
if [[ "${CMAKE_EXTRA_FLAGS}" == *"RAJA_ENABLE_COVERAGE"* ]] ; then bash <(curl -s https://codecov.io/bash) -a "-f"; fi
```

What if we just redirected bash to /dev/null?

```bash
if [[ "${CMAKE_EXTRA_FLAGS}" == *"RAJA_ENABLE_COVERAGE"* ]] ; then (bash <(curl -s https://codecov.io/bash) -a "-f") &>/dev/null; fi
```

If coverage succeeds, it will still upload. If it fails, then it won't upload. There's no reason for the coverage stuff to fail either way...

include/RAJA/pattern/nested/internal.hpp
```diff
@@ -0,0 +1,118 @@
...
+    camp::list<>,
+    camp::as_list<camp::idx_seq_from_t<IndexTypes>>>::type;
+
+template <typename Iterator>
+struct iterable_value_type_getter {
+  using type = typename Iterator::iterator::value_type;
+};
+template <>
+struct iterable_value_type_getter<IndexSet> {
+  // TODO: when static indexset drops, specialize properly
```
willkill07: This has dropped, but it currently not well defined. We'd need to create a type trait which extracts and ensures all Index_types are the same.

include/RAJA/internal/LegacyCompatibility.hpp
willkill07: Edit: with `std::forward` that is the case. `camp::forward` this may be required.

include/RAJA/pattern/nested/internal.hpp
trws: Right, that todo is slightly inaccurate now, we need to determine how we're going to handle indices in a static index set.

include/RAJA/internal/LegacyCompatibility.hpp
trws: You were right in the first place, this should be just Op.

include/camp/tuple.hpp
trws: Thanks for the catch on this, not sure how I missed it until now but this is fixed.

trws:
Ok, all tests pass with the exception of the TBB tests which can't be built due to #340.  With that caveat, are folks good with merging this? @rhornung67, @ajkunen?

trws:
Help on documentation would be awesome.  I was hoping to have a great 
deal more of that done, but ended up stuck writing 3 big tickets for the 
OpenMP release next month.  At least that means we’re getting deep 
copy, unified memory support and a way to have the compiler check your 
mapping stuff, but it’s sucked time like crazy. ><

PR #234, opened 2017-04-19
StaticIndexSet and Typed Segments

pearce8:
Static index set implementation.

PR #281, opened 2017-07-20
change to std::ptrdiff_t for Index_type by default

trws:
Now that it doesn't cause compilation errors to do this, I would propose we use a pointer difference type by default for Index_type.  Tests with the perfsuite show differences as in the noise, possibly slightly faster, but this is much less likely to be a down-conversion from the user's index type than int.  I'm relatively surprised that nobody has run into issues trying to do a forall over more than ~4,000,000 elements yet, but it seems like it's just a matter of time.


davidbeckingsale:
Hey @trws is this still something we wanted to merge? I think you mentioned it was causing problems in an email somewhere...

trws:
After @willkill's addition of SFINAE checks to the generic forall, this is actually good to go.  It will be subsumed by Will's concepts branch too actually, so we can either just merge this to have it be done, or kill this and wait for that.


---
Sent from VMware Boxer

On July 25, 2017 at 9:19:28 AM PDT, David Beckingsale <notifications@github.com> wrote:

Hey @trws<https://github.com/trws> is this still something we wanted to merge? I think you mentioned it was causing problems in an email somewhere...

davidbeckingsale:
Okay, let's get this merged.

willkill07:
Also, the TypedRangeSegment and make_segment() function added to the internals of forall (under the IndexSetS branch) also fix any potential problems

trws:
This is resolved by #234.

PR #283, opened 2017-07-25
Feature/burmark1/cuda reducers pinned

MrBurmark:
This pull request is an alternative implementation of the features in #280.

1. Cuda reducers use small pinned memory buffers to hold the result of each kernel launch. This should be more flexible once we begin using streams. It also removes the need to call cudaMemcpy. The downside is that it will slightly reduce performance in atomic reducers compared to #280 because the device memory on which the atomics are used must be initialized on the device.

2. Per thread variables holding information about the current cuda kernel launch are omp threadprivate. ~~no longer c++11 thread_local variables and are stored in an array of length omp_get_max_threads. Thread_local is preferable in multiple ways, but may not work with every openmp implementation, please correct me if I'm wrong. The current implementation only works with a single level of omp parallelism.~~

3. Util lock_guard and omp mutex added to simplify use of locks.


davidbeckingsale:
If we are adding alternative implementations of features we need to have a way to select between them that will be as objective as possible....

include/RAJA/policy/cuda/policy.hpp
```diff
@@ -179,14 +180,10 @@ using cuda_reduce_atomic_async = cuda_reduce_atomic<BLOCK_SIZE, true>;
...
 const int WARP_SIZE = 32;
-const int RAJA_CUDA_MAX_BLOCK_SIZE = 2048;
-
-/*!
- * \def RAJA_CUDA_LAUNCH_PARAMS(gridSize, blockSize)
- * Macro that generates kernel launch parameters.
- */
-#define RAJA_CUDA_LAUNCH_PARAMS(gridSize, blockSize) \
-  gridSize, blockSize, getCudaSharedmemAmount(gridSize, blockSize)
+const int RAJA_CUDA_MAX_BLOCK_SIZE = 1024;
```
willkill07: 1). These should be constexpr.
2). Should there be a way to define the maximum @ configure?

include/RAJA/policy/cuda/forall.hpp
```diff
@@ -197,39 +197,31 @@ RAJA_INLINE void forall(cuda_exec<BLOCK_SIZE, Async>,
...
-       offset += step_size) {
+    RAJA_FT_BEGIN;
 
-    step_size = RAJA_MIN(total_len - offset, max_step_size);
+    cudaStream_t stream = 0;
 
-    auto begin = first_begin + offset;
-    auto end = begin + step_size;
+    forall_cuda_kernel<<<gridSize, BLOCK_SIZE, 0, stream>>>(
+        RAJA::cuda::createLaunchBody(
```
willkill07: if `createLaunchBody` is a "maker" why not call it `make_launch_body`?

include/RAJA/policy/cuda/forall.hpp
```diff
@@ -197,39 +197,31 @@ RAJA_INLINE void forall(cuda_exec<BLOCK_SIZE, Async>,
...
+            gridSize, BLOCK_SIZE, 0, stream,
+            std::forward<LOOP_BODY>(loop_body)),
+        std::move(begin), len);
+    cudaErrchk(cudaPeekAtLastError());
```
willkill07: I noticed the abstraction of cuda synchronize. Is it possible to abstract away error checking into a clean API call as well?

include/RAJA/policy/cuda/policy.hpp
```diff
@@ -179,14 +180,10 @@ using cuda_reduce_atomic_async = cuda_reduce_atomic<BLOCK_SIZE, true>;
...
+const int MAX_WARPS = RAJA_CUDA_MAX_BLOCK_SIZE / WARP_SIZE;
+static_assert(WARP_SIZE >= MAX_WARPS,
```
willkill07: A check should also be made that ensures WARP_SIZE is a multiple of MAX_BLOCK_SIZE too, yes?

include/RAJA/policy/cuda/reduce.hpp
```diff
@@ -62,29 +62,102 @@
...
+template <typename Reducer>
+struct atomic;
+
+template <typename T>
+struct atomic<sum<T>>
+{
+  RAJA_DEVICE RAJA_INLINE
+  void operator()(T &val, const T v)
+  {
+    RAJA::_atomicAdd(&val, v);
```
willkill07: if _atomicAdd is only viable for cuda, shouldn't it live under the `RAJA::cuda` namespace as well?

include/RAJA/policy/cuda/reduce.hpp
```diff
@@ -109,1507 +390,1185 @@ __device__ __forceinline__ T shfl_xor(T var, int laneMask)
...
-        if (threadId < i) {
-          sd[threadId] = RAJA_MIN(sd[threadId], sd[threadId + i]);
-        }
-      }
+//! Reduction data for Cuda Offload -- stores value, host pointer
+template <bool Async, typename Reducer, typename T>
+struct ReduceAtomic_Data {
+  //! union to hold either pointer to PinnedTally or poiter to value
+  //  only use list before setup for device and only use val_ptr after
+  union tally_u {
```
willkill07: I'd consider renaming this to tally_or_val_ptr or something a little longer. 

Also, naming the variable `tally` within the class makes the structure appear to only be a tally when in fact it also stores a pointer to a value (T). This makes code within the scope of this class confusing in a few places:

`tally.val_ptr = tally.list->new_value(currentStream());`

If I know tally is a union, the following makes my eyes hurt a bit. I'd personally add a member function to tally that does this. Less room for programmer error (I'm just trying to identify where I would get burned)

include/RAJA/policy/cuda/reduce.hpp
```diff
@@ -109,1507 +390,1185 @@ __device__ __forceinline__ T shfl_xor(T var, int laneMask)
...
+  }
+
+  //! check and setup for device
+  //  allocate device pointers and get a new result buffer from the pinned tally
+  RAJA_INLINE
+  bool setupForDevice(Offload_Info &info)
+  {
+    bool act = !device && setupReducers();
+    if (act) {
+      device = device_mempool_type::getInstance().malloc<T>(1);
```
willkill07: Why do we need a device_mempool_type and device_zeroed_mempool_type? Is the overhead really that bad? can't we just zero it out?

include/RAJA/policy/cuda/reduce.hpp
```diff
@@ -109,1507 +390,1185 @@ __device__ __forceinline__ T shfl_xor(T var, int laneMask)
...
-  static_assert(powerOfTwoCheck, "Error: block sizes must be a power of 2");
-  static_assert(reasonableRangeCheck,
-                "Error: block sizes must be between 32 and 1024");
-  static_assert(sizeofcheck,
-                "Error: type must be of size <= " RAJA_STRINGIFY_MACRO(
-                    RAJA_CUDA_REDUCE_VAR_MAXSIZE));
-};
+//! Cuda Reduction Atomic entity -- generalize on reduction, and type
+template <bool Async, typename Reducer, typename T>
+struct CudaReduceAtomic {
```
willkill07: I'm confused. I thought we generalized on the Reducer type. Isn't Atomic dependent upon the reducer?

include/RAJA/policy/cuda/reduce.hpp
```diff
@@ -109,1507 +390,1185 @@ __device__ __forceinline__ T shfl_xor(T var, int laneMask)
...
-                                   sd_val[threadId],
-                                   sd_idx[threadId],
-                                   sd_val[threadId + i],
-                                   sd_idx[threadId + i]);
+      // reduce each warp
+      for (int i = 1; i < WARP_SIZE ; i *= 2) {
+        int srcLane = threadId ^ i;
+        T rhs_val = internal::shfl_sync<T>(temp.val, srcLane);
+        IndexType rhs_idx = internal::shfl_sync<T>(temp.idx, srcLane);
+        // only add from threads that exist (don
```
willkill07: You can just say ReduceMin and not create the self type alias. The fully qualified template list is implied.

include/RAJA/policy/cuda/policy.hpp
MrBurmark: 2) Maybe, this might be something that could be configured by looking at the sm version provided.

include/RAJA/policy/cuda/reduce.hpp
MrBurmark: It should.

trws:
> Per thread variables holding information about the current cuda kernel launch are no longer c++11 thread_local variables and are stored in an array of length omp_get_max_threads. Thread_local is preferable in multiple ways, but may not work with every openmp implementation, please correct me if I'm wrong. The current implementation only works with a single level of omp parallelism.

I'll take you up on that. 🙋‍♂️  

thread_local is required by c++11, rather than OpenMP, but any sane OpenMP will let it work.  For the non-sane OpenMP implementations, there's `#pragma omp threadprivate` which does the same thing but is required by OpenMP rather than C++, so it's required by all conforming compilers that could possibly have the OpenMP policies work.

include/RAJA/policy/cuda/reduce.hpp
```diff
@@ -94,12 +167,220 @@ __device__ __forceinline__ T shfl_xor(T var, int laneMask)
...
+    : stream_list(nullptr)
+  {
+
+  }
+  
+  PinnedTally(const PinnedTally&) = delete;
+
+
+#if defined(RAJA_ENABLE_OPENMP)
+  omp::mutex& mutex()
```
willkill07: I'd get rid of this function. using m_mutex is perfectly fine. If it's used elsewhere, there's no reason not to make it public.

jonesholger:
Ok, I've got numbers from the perfsuite kernel that I put together to test ordered reductions

rzmanta
develop: 0.161982
CudaReducersNoMax: 0.059507
CudaReducersPinned: 0.059524

rzhasgpu
develop: 0.605651
CudaReducersNoMax 0.116059
CudaReducersPinned 0.116150

I would recommend at this juncture to clean-up/reconcile #283 with the latest develop, and delete CudaReducersNoMax

willkill07:
@jonesholger the improvements here are really good. Given the minimal variance between the two, this pinned approach does seem better. Out of curiosity, how many reducers were in the ordered reductions perfsuite kernel?

willkill07:
I think NVCC consumes all of Travis' resources:

https://travis-ci.org/LLNL/RAJA/jobs/260429524#L2899

We really need to get that job limit PR merged 😉

jonesholger:
@willkill07 the perfsuite uses three reductions in a recursion (sum,minloc,maxloc), and once we get this into develop we can increase the recursion levels to have "many" reducers active simultaneously. For now the recursion limit is set to not go beyond the default max reducers which is 8.

include/RAJA/policy/cuda/forall.hpp
```diff
@@ -74,22 +74,40 @@
...
-namespace INTERNAL
+/*!
+ ******************************************************************************
+ *
+ * \brief calculate gridDim from length of iteration and blockDim
+ *
+ ******************************************************************************
+ */
+RAJA_INLINE
+dim3 getGridDim(size_t len, dim3 blockDim)
```
trws: Why this rather than cudaOccupancyMaxPotentialBlockSize or cudaOccupancyMaxPotentialBlockSizeVariableSMem?

include/RAJA/policy/cuda/forall.hpp
```diff
@@ -74,22 +74,40 @@
...
+  int device = 0;
+  static int num_sm[8] {0, 0, 0, 0, 0, 0, 0, 0};
+  if (num_sm[device] == 0) {
+    cudaDeviceGetAttribute(&num_sm[device], cudaDevAttrMultiProcessorCount, device);
+  }
+  size_t max_concurrent_blocks = num_sm[device] * (MAX_THREADS_PER_SM/block_size) ;
+
+  size_t gridSizeMax = (len + block_size-1) / block_size;
+
+  return std::min(gridSizeMax, max_concurrent_blocks);
```
trws: So, we still have a maximum number of blocks per kernel with a reducer with this version?

davidbeckingsale: 😞

jonesholger: This is a performance feature so the reducers can use grid stride loops, allowing each thread to reduce multiple values before we start folding. A lot of gain can be had here. This one of the major reasons why CUB could do so much better than our default reduction techniques, so we wanted to induce similar behavior. This benefits larger iteration spaces. @MrBurmark  has a lingering concern that this may impact the streaming memcpy benchmark - and that he will look into whether reducers are used or not and adjust appropriately.

davidbeckingsale: It's worth noting that the stream benchmark was just where we observed the biggest performance impact. This overhead would still exist for other loops with larger iteration spaces but the impact might not be so large.

jonesholger: I'll  setup some additional perfsuite kernels so we can see the comparisons with CUB

trws: That's a fair point, but wouldn't using cudaOccupancyMaxActiveBlocksPerMultiprocessor here be worthwhile? Not all GPUs have a limit on threads of 2048 per SM, nor does the maximum number of blocks per SM necessarily match that calculation.  For the record, what cub does for this is to multiply the result of cudaOccupancyMaxActiveBlocksPerMultiprocessor with the number of SMs , then multiply that with an oversubscription factor between 3 and 10, then either an even share distribution across those blocks or a dynamic distribution through a shared queue.  What I'm driving at with this is that if that's the goal, we probably want to tweak the calculation some.

jonesholger: @trws This all sounds good. Since, we're using CUB as the performance target it's definitely reasonable to tweak this calculation, and besides I'm curious about the oversubscription factor, or anything else that we leverage. Another interesting exercise is to see whether we can gain an advantage with multiple reducers in the same kernel.

trws: It seems they use a static factor based on the PTX version in use, it's in the util_arch header IIRC.

MrBurmark:
Let's address the grid striding issue in another pull request so this pull request can stay focused on generality instead of performance.

trws:
So, this has been sitting for a while.  Aside from an update from develop, is there any reason not to merge?

rhornung67:
@jonesholger and @MrBurmark need to weigh in on whether this covers sufficiently aspects of different approaches they were exploring. If so, then a few others (including myself) should review.

MrBurmark:
I need to add in the old reducer backend so codes can still link with old libraries after this PR. Other than that this covers generalizing the cuda reducers in both size and number.
Let's talk about adding scheduling to cuda policies at the next RAJA meeting. This is a cross policy issue and we should think about how to specify it in a way that isn't completely different for each backend.

rhornung67:
@MrBurmark I will add cuda scheduling to the agenda for next week's meeting. We may not be able to completely resolve what we want to do, but we should be able to identify a small subset of folks to work through the +/- of different options.

trws:
So we can give it some thought, what do you mean by "scheduling" in this 
context?

DavidPoliakoff:
If scheduling is anything like the asynchronous streams stuff, remember that we have #188 from a ways back as one possible template

willkill07:
@MrBurkmark “I need to add in the old reducer backend so codes can still link with old libraries after this PR.”

My alarms just went off here...

Wouldn’t this only be the library functions and not the headers? I foresee massive pain if there are two versions of CUDA reducers available to end users.

MrBurmark:
Scheduling in terms of how much and which work gets assigned to each logical thread. Specifically doing this with respect to the occupancy that can be achieved for the current kernel on the device being used.

MrBurmark:
@trws Do you see anything else that needs to be done before this can be merged?

trws:
Assuming the auto-merge passes that is, but I don't see why it wouldn't.

trws:
It looks like this is all good, but running into the same issues as the 
SIMD PR with atomics and clang versions.  As soon as that's done and 
merged over this should all pass I expect.

trws:
Oddly, we seem to have reversed.  All of the travis tests pass, but MSVC now gives up in the middle of tiling apparently..?

MrBurmark:
Tests passed here. Openmp scan tests failed with cuda enabled though.

trws:
That should be fixed now, there was a cmake issue involved.  I'm happy with this at this point.

PR #284, opened 2017-07-25
Feature/artv3/intro examples

artv3:
//-----[Examples for RAJA]----------- 
AddVectors – Gentle introduction to RAJA
Matrix-Matrix multiply  - Examples of nesting and forallN
Jacobi Solver  - Example of reduction operator
TDFD Acoustic Wave Solver  - Example of stencil operation
Coloring – Example of creating a custom index set  (credit to Will for helping me understand RAJA concepts).


examples/addVectors.cxx
```diff
@@ -0,0 +1,194 @@
...
+  -----[RAJA Concepts]---------------
+  1. Introduces the forall loop and basic RAJA policies
+
+  RAJA::forall<RAJA::exec_policy>(RAJA::range_policy,[=](index i)) {
+
+         //body
+
+  });
+  
+  ----[Arguments]--------------
```
davidbeckingsale: I'm not sure 'arguments' is the best way to describe these.

examples/addVectors.cxx
```diff
@@ -0,0 +1,194 @@
...
+  std::cout<<"Standard C++ Loop"<<std::endl;
+  for(int i=0; i<N; ++i) {
+    C[i] = A[i] + B[i];
+  }
+  checkSolution(C,N);
+  std::cout<<"\n"<<std::endl;
+  //======================================
+     
+  
+  /*----[RAJA: Sequential Policy]---------
```
davidbeckingsale: Comments should be consistent in style.

examples/addVectors.cxx
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  //----[Standard C++ Loop]---------------
+  std::cout<<"Standard C++ Loop"<<std::endl;
+  for(int i=0; i<N; ++i) {
+    C[i] = A[i] + B[i];
+  }
+  checkSolution(C,N);
+  std::cout<<"\n"<<std::endl;
+  //======================================
```
davidbeckingsale: These comments are unnecessary.

examples/addVectors.cxx
```diff
@@ -0,0 +1,194 @@
...
+     
+  
+  /*----[RAJA: Sequential Policy]---------  
+  RAJA::seq_exec -  Executes the loop sequentially
+
+  RAJA::RangeSegment(start,stop) - generates a sequence of numbers
+  on which a RAJA loop may iterate on
```
davidbeckingsale: 👍

examples/addVectors.cxx
```diff
@@ -0,0 +1,194 @@
...
+#endif
+
+
+
+#if defined(RAJA_ENABLE_CUDA)
+  /*----[RAJA: CUDA Policy]---------
+    RAJA::cuda_exec<CUDA_BLOCK_SIZE> - excecutes RAJA-loop using the CUDA API
+    Each thread is assigned to an iteration of the loop
+
+    CUDA_BLOCK_SIZE - specifies the number of threads per cartesian dimension
```
davidbeckingsale: 👍

examples/addVectors.cxx
```diff
@@ -0,0 +1,194 @@
...
+  int *C = allocate(N);
+  
+  //Populate vectors
+  for(int i=0; i<N; ++i) {
+    A[i] = i;
+    B[i] = i;
+  }
+
+  //----[Standard C++ Loop]---------------
+  std::cout<<"Standard C++ Loop"<<std::endl;
```
willkill07: My only note here is to not (ab)use std::endl

Prefer '\n' unless you really have a reason to flush the output stream.

examples/addVectors.cxx
willkill07: I would just add that it generates a contiguous sequence of numbers by the [start, stop) interval specified.

examples/coloring.cxx
```diff
@@ -0,0 +1,138 @@
...
+ 
+
+  //Defining a custom IndexSet
+  RAJA::IndexSet colorset;
+
+  //RAJA::View introduces multidimensional arrays
+  RAJA::View<int,RAJA::Layout<2>> Aview(A, n, n);
+  
+  //Buffer used for intermediate indicy storage
+  RAJA::Index_type *idx = RAJA::allocate_aligned_type<RAJA::Index_type>(64, n * n / 4);
```
willkill07: To allow odd numbers, you can say `(n + 1)*(n + 1) / 4`

examples/coloring.cxx
```diff
@@ -0,0 +1,138 @@
...
+      
+      RAJA::Index_type count = 0;
+      
+      // Iterate over each extent in each dimension, incrementing by two to
+      // safely advance over neighbors
+      for (int xiter = xdim; xiter < n; xiter += 2) {
+        for (int yiter = ydim; yiter < n; yiter += 2) {
+          
+          // Add the computed index to the buffer
+          idx[count] = std::distance(A, std::addressof(Aview(xiter, yiter)));
```
willkill07: In the generalized color example, I had `std::addressof(Aview(0,0))` instead of `A`. Ultimately, these are the same. Just more for sake of a comment.

examples/coloring.cxx
```diff
@@ -0,0 +1,138 @@
...
+      colorset.push_back(RAJA::ListSegment(idx, count));
+    }
+  }  
+
+  // Clear temporary buffer
+  RAJA::free_aligned(idx);
+
+
+  //----[RAJA Sequential/OMP execution]---------
+  //RAJA: Seq_segit - Sequational Segment Iteraion
```
willkill07: typos here. I'd also add a comment about how IndexSet traversal works.

examples/jacobi.cxx
```diff
@@ -0,0 +1,333 @@
...
+  o--[]--o--[]--o--[]--o--[]--o
+  |      |      |      |      |  
+  [x]    []     []     []     []     
+  |      |      |      |      |
+  o--[]--o--[]--o--[]--o--[]--o
+
+
+  ----[RAJA Concepts]---------------
+  1. RAJA Reduction
+  2. RAJA::omp_collapse_nowait_exec
```
willkill07: I'd add another point and mention forallN

examples/jacobi.cxx
```diff
@@ -0,0 +1,333 @@
...
+    
+    //Iteration of the Jacobi Scheme
+    for(int n=1;n<=N;++n) {
+      for(int m=1;m<=N;++m) {
+
+        int id = n*(N+2) + m;
+
+        //Cell (1,1) is a special case due to battery
+        if(n==1 && m==1) {
+          invD = 1./3.0; V = 1.0;
```
willkill07: for the sake of a user looking at these examples, it would be nice to know what invD and V are. You mention the parameters but do not describe them.

examples/matrixMultiply.cxx
```diff
@@ -0,0 +1,272 @@
...
+      C[cId] = dot;
+    }        
+  }
+  checkSolution(C,N);
+  std::cout<<"\n"<<std::endl;
+  //======================================
+
+  //-----[RAJA: Sequential Policy - Single Forall loop]----------
+  std::cout<<"RAJA: Sequential Policy - Single forall"<<std::endl;
+  RAJA::forall<RAJA::seq_exec>(RAJA::RangeSegment(0,NN),[=](int i) {
```
willkill07: It would be really nice to tie in RAJA::View here with matrix multiply

examples/wave.cxx
```diff
@@ -0,0 +1,372 @@
...
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#include <cstdlib>
+#include <iostream>
+#include <cmath>
+
+#include "RAJA/RAJA.hpp"
+#include "RAJA/util/defines.hpp"
+
+#define sr 2 //Stencil Radius
```
willkill07: prefer `constexpr <type> <var> = <expr>` to preprocessor definitions wherever possible

examples/wave.cxx
```diff
@@ -0,0 +1,372 @@
...
+#define PI 3.14159265359
+#define CUDA_BLOCK_SIZE 16
+
+
+typedef struct grid{
+  double ox,dx;
+  int nx;
+}grid_s;
```
willkill07: In C++, you don't need the typedef. It's also generally preferred to have the type definition separate from the variable declaration, but that's my $0.02.

examples/wave.cxx
```diff
@@ -0,0 +1,372 @@
...
+      for(int tx=0; tx<nx; ++tx) {
+        
+        const int id = tx + ty*nx;
+        T P_old  = P1[id]; 
+        T P_curr = P2[id]; 
+        
+        //Compute laplacian
+        T lap = 0.0;
+        
+        for(int r=-sr; r<=sr; ++r) {
```
willkill07: I don't know if we want to try to show anything beyond a simple for-loop, but

`for (auto r : RAJA::RangeSegment(-sr, sr + 1))` would be pretty cool to highlight :)

examples/wave.cxx
```diff
@@ -0,0 +1,372 @@
...
+  double cc = 1./2.0;
+
+  //Multiplier for spatial discretization
+  double factor = 8.0;
+
+  //Discretization of the domain. 
+  //The same discretization of the x-dimension wil be used for the y-dimension
+  grid_s grid; 
+  grid.ox   = -1; 
+  grid.dx   = 0.1250/factor;
```
willkill07: Are the constants here magic numbers? could they be anything?

examples/addVectors.cxx
```diff
@@ -0,0 +1,192 @@
...
+  RAJA::forall<RAJA::exec_policy>(RAJA::range_policy,[=](index i)) {
+
+         //body
+
+  });
+  
+  [=] Pass by copy
+  [&] Pass by reference
+  RAJA::exec_policy  - specifies how the traversal occurs
+  RAJA::range_policy - provides a list in which the index may iterate on
```
trws: RAJA::range_policy should more properly be Iterable, or perhaps segment, these days.  Keeping policies and iterables separate may help keep the concepts clear.

examples/addVectors.cxx
```diff
@@ -0,0 +1,192 @@
...
+  return 0;
+}
+
+void checkSolution(int *C, int in_N){
+
+  bool eFlag = true; 
+  for(int i=0; i<in_N; ++i){
+    if( (C[i] - (i+i)) > 1e-9){
+      std::cout<<"Error in result!"<<std::endl;
+      eFlag = false; break;
```
trws: Why break rather than return? With return the flag becomes unnecessary.

examples/coloring.cxx
```diff
@@ -0,0 +1,138 @@
...
+  std::copy(init.begin(), init.end(), A);
+ 
+
+  //Defining a custom IndexSet
+  RAJA::IndexSet colorset;
+
+  //RAJA::View introduces multidimensional arrays
+  RAJA::View<int,RAJA::Layout<2>> Aview(A, n, n);
+  
+  //Buffer used for intermediate indicy storage
```
trws: indicy -> index

examples/jacobi.cxx
```diff
@@ -0,0 +1,333 @@
...
+  printf("Top right current: %lg \n", I[N+N*(N+2)]);
+  printf("No of iterations: %d \n \n",iteration);
+  //======================================
+
+  //----[RAJA: Sequential Policy - Nested forallN]---------
+  printf("RAJA: Sequential Policy - Nested forallN \n"); 
+  resI2 = 1; iteration = 0; 
+  memset(I,0,NN*sizeof(double));
+  memset(Iold,0,NN*sizeof(double));
+  while(resI2 > tol*tol) {
```
trws: For the RAJA cases in this example, I think it would be worth making the point that one can reuse RAJA code for different backing models here.  A function with the exec and reduce policies as parameters would be all it would take to abstract that out and remove the replication. (well, maybe also a string or two, but close enough)

examples/wave.cxx
trws: Seconded.  The definition and declaration as a single statement isn't too bad, but the typedef just looks odd in C++.

examples/wave.cxx
```diff
@@ -0,0 +1,372 @@
...
+template <size_t BLOCK_SIZE>
+struct Wave<RAJA::cuda_exec<BLOCK_SIZE>, RAJA::cuda_exec<BLOCK_SIZE> > {
+  template <typename T>
+    void operator()(T* P1, T* P2, int nx, double ct) {
+
+    RAJA::forallN<RAJA::NestedPolicy<
+    RAJA::ExecList<
+    RAJA::cuda_threadblock_y_exec<BLOCK_SIZE>,
+    RAJA::cuda_threadblock_x_exec<BLOCK_SIZE> >>>(
+    RAJA::RangeSegment(0, nx), RAJA::RangeSegment(0, nx), [=] __device__ (int tidy, int tidx) {
```
trws: Does this only need to be split out because of __device__?  If so, for the sake of examples at least, I would say __host__ __device__ would be worth it to have just one RAJA version in here.

willkill07: Yes, that was the only reason this was done when I worked with @artv3 on this

trws:
Alternately, if performance was the consideration, a `__host__ 
__device__` functor would work quite well there.  Though I'm not sure if 
we want that in an example or not to be honest.

artv3:
Will mentioned moving the C/C++ version onto a different file. What do you guys think?

examples/wave.cxx
artv3: They could be anything. ox corresponds to the origin of the x-component, dx is the grid spacing (I assume equidistant points), and nx is the total number of grid points. My solver assumes a periodic solution though.... so we can somewhat vary the details.

examples/coloring.cxx
```diff
@@ -0,0 +1,142 @@
...
+  auto init = {1, 2, 1, 2, 
+               3, 4, 3, 4, 
+               1, 2, 1, 2,
+               3, 4, 3, 4};
+  
+  std::copy(init.begin(), init.end(), A);
+ 
+
+  //Defining a custom IndexSet
+  RAJA::IndexSet colorset;
```
rhornung67: @artv3 this should be converted to use the new StaticIndexSet class since IndexSet will go away and be replaced with the static one.

examples/matrixMultiply.cxx
```diff
@@ -0,0 +1,269 @@
...
+      }
+      
+      C[cId] = dot;
+    }        
+  }
+  checkSolution(C,N);
+  std::cout<<"\n"<<std::endl;
+  //======================================
+
+
```
rhornung67: @artv3 it may make this example clearer if you kept the outer loop range [0,N) and put a doubly-nested loop inside the lambda. Thoughts?

examples/wave.cxx
rhornung67: C++ purists :-)

examples/wave.cxx
rhornung67: I agree.

examples/coloring.cxx
willkill07: Currently I do have `IndexSet` living around as an alias to `StaticIndexSet<RangeSegment, ListSegment, RangeStrideSegment>`

For the purposes of this example, the fully qualified required name would be `StaticIndexSet<TypedListSegment<RAJA::Index_type>>`

rhornung67: Thanks @willkill07

examples/wave.cxx
```diff
@@ -43,169 +43,59 @@
...
+*/
+template<typename T, typename fdNestedPolicy>
+void Wave(T* P1, T* P2, int nx, double ct){
 
-//Generic RAJA implementation for CPU
-template <typename Policy1, typename Policy2>
-struct Wave {
-  template <typename T>  
-    void operator()(T* P1, T* P2, int nx, double ct) {
+  RAJA::RangeSegment fdGridx = RAJA::RangeSegment(0,nx);
```
willkill07: The only thing I would change is only specifying the type once:

```cpp
RAJA::RangeSegment fdGridx(0, nx);
```

rhornung67:
@artv3 as long as you're working in the examples directory, please change the file extensions to *.cpp to match the rest of RAJA. We changed to that recently and missed the pi example file. Thanks.

examples/wave.cxx
artv3: The latest version has a single kernel for all the RAJA variants, I removed the c/cpp version. I included the __device__ __host__ keywords to make it truly portable.

examples/coloring.cpp
```diff
@@ -0,0 +1,158 @@
...
+               3, 4, 3, 4, 
+               1, 2, 1, 2,
+               3, 4, 3, 4};
+  
+  std::copy(init.begin(), init.end(), A);
+   
+  /*
+    Instance of an IndexSet
+  */
+  RAJA::StaticIndexSet<RAJA::TypedListSegment<RAJA::Index_type> > colorset;
```
willkill07: Maybe explain this a little more. The template arguments for IndexSet allow the user to indicate the required storage types of various segments. In this example, we only need to store TypedListSegment<Index_type> (aka ListSegment) so that's the only template argument for our IndexSet

examples/jacobi.cpp
```diff
@@ -0,0 +1,274 @@
...
+  }
+  printf("Value at grid point ((N-1), (N-1)): %lg \n", I[N+N*(N+2)]);
+  printf("No of iterations: %d \n \n",iteration);
+    
+  /*
+    RAJA loop calls may be shortened by defining policies before hand
+  */
+  RAJA::RangeSegment jacobiRange = RAJA::RangeSegment(1,(N+1)); 
+  RAJA::RangeSegment gridRange = RAJA::RangeSegment(0,NN); 
+  using jacobiSeqNestedPolicy = RAJA::NestedPolicy< RAJA::ExecList< RAJA::seq_exec, RAJA::seq_exec > >;
```
willkill07: `RAJA::RangeSegment <name> = RAJA::RangeSegment(<args...>)` can be simplified to `RAJA::RangeSegment <name> (<args...>)`

maybe add explanation for what's required with NestedPolicy here.

examples/jacobi.cpp
```diff
@@ -0,0 +1,274 @@
...
+  printf("RAJA: OpenMP Policy - Nested forallN \n");
+  resI2 = 1; iteration = 0; 
+  memset(I,0,NN*sizeof(double));
+  memset(Iold,0,NN*sizeof(double));
+  
+  /*
+    RAJA::omp_collapse_nowait_exec -
+    parallizes nested loops without introducing nested parallism
+    
+    RAJA::OMP_Parallel<> - Creates a parallel region
```
willkill07: mention how this has to be at the end of the Nested<> template

examples/jacobi.cpp
```diff
@@ -0,0 +1,274 @@
...
+  }
+  printf("Value at grid point ((N-1),(N-1)): %lg \n", I[N+N*(N+2)]);
+  printf("No of iterations: %d \n \n",iteration);
+
+#endif
+
+
+#if defined(RAJA_ENABLE_CUDA)
+  using jacobiCUDANestedPolicy = 
+    RAJA::NestedPolicy<RAJA::ExecList<RAJA::cuda_threadblock_y_exec<CUDA_BLOCK_SIZE>,RAJA::cuda_threadblock_x_exec<CUDA_BLOCK_SIZE>>>;
```
willkill07: clang-format this line -- even github won't show all of it to me

examples/matrixMultiply.cpp
```diff
@@ -0,0 +1,247 @@
...
+  });
+  checkSolution(Cview, N);
+
+  printf("RAJA: Sequential Policy RAJA - forallN \n");
+  /*
+    Nested forall loops may be collapsed into a single forallN loop
+  */
+  RAJA::forallN<RAJA::NestedPolicy<RAJA::ExecList<RAJA::seq_exec,
+                                                  RAJA::seq_exec>>>(
+  RAJA::RangeSegment(0,N), RAJA::RangeSegment(0,N), [=](int row, int col) {
```
willkill07: you've isolated the RangeSegment declaration to be before the call to forall in prior examples. Maybe the same should be done here.

examples/matrixMultiply.cpp
```diff
@@ -0,0 +1,247 @@
...
+  for (int id = 0; id < in_N * in_N; ++id) {
+    if (abs(C[id] - in_N) > 1e-9) {
+      printf("Error in Result \n \n");
+      return;
+    }
+  }
+  printf("Correct Result \n \n");
+}
+
+void checkSolution(RAJA::View<double,RAJA::Layout<DIM> > Cview, int in_N){
```
willkill07: why are two checkSolutions provided here?

examples/wave.cpp
```diff
@@ -0,0 +1,263 @@
...
+};
+
+
+/*
+  Wave Propagator 
+*/
+template<typename T, typename fdNestedPolicy>
+void Wave(T* P1, T* P2, RAJA::RangeSegment fdBounds, double ct, int nx){
+
+  RAJA::forallN<fdNestedPolicy >(fdBounds, fdBounds , [=] __device__ __host__ (int ty, int tx) {
```
willkill07: __device__ __host__ can be changed to RAJA_HOST_DEVICE

examples/wave.cpp
```diff
@@ -0,0 +1,263 @@
...
+  dt = T/nt; //Final time-step 
+  ct = (cc*dt*dt)/(grid.dx*grid.dx); //Merge coefficients into a single coefficient
+
+
+  /*
+    Selecting a policy
+  */
+
+  /*
+    Sequential Policy
```
willkill07: Code cleanup in this general section?

willkill07:
@artv3 this is looking good. Our CI tests are still failing with this branch. If you have some cycles, I'd focus on getting Travis to pass. If you have any questions, please let me know.

rhornung67:
:+1: @willkill07 for the travis suggestion.  @artv3 nice work putting these examples together.

.travis.yml
```diff
@@ -157,7 +157,7 @@ script:
 - mkdir travis-build &&
   cd travis-build
 - cmake -DCMAKE_CXX_COMPILER="$COMPILER" $WARN $CMAKE_EXTRA_FLAGS ../
-- make -j
+- make
```
willkill07: Why is parallel compilation being disabled?

artv3: This was in regards to Travis and the Nvidia compiler. It was eating up all the virtual memory and failing the test. @davidbeckingsale 's suggestion was to turn it off.

willkill07: Yeah, after looking at the log it made more sense. 👍

examples/CMakeLists.txt
```diff
@@ -44,3 +44,23 @@
 raja_add_executable(
   NAME example-raja-pi
   SOURCES pi.cpp)
+
+raja_add_executable(
```
willkill07: The only thing I would change at this point is having unified naming for all of the example binaries that get emitted.

* `example-raja-pi` --> `example-pi`
* `addVectors` --> `example-vector-add`
* `matrixMultiply` --> `example-matrix-multiply`
* `jacobi` --> `example-jacobi`
* `wave` --> `example-wave`
* `customIndex` --> `example-custom-index`

Other than that, the source file names are appropriate for the examples :)

examples/addVectors.cpp
```diff
@@ -0,0 +1,165 @@
...
+  return 0;
+}
+
+
+/*
+  RAJA may also be used outside of computational kernels
+*/
+void checkSolution(int *C, int in_N)
+{
+
```
willkill07: Extra whitespace here -- just be consistent

examples/customIndex.cpp
```diff
@@ -0,0 +1,158 @@
...
+                             ListSegments
+*/
+int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
+{
+
+  printf("Example 5. Custom Index Set");
+  int n = 4;
+  int *A = new int[n * n];
+
+  auto init = {1, 2, 1, 2, 3, 4, 3, 4, 1, 2, 1, 2, 3, 4, 3, 4};
```
willkill07: If you want to make this more palatable to folks, you can manually format this to be 4x4 again

examples/matrixMultiply.cpp
```diff
@@ -0,0 +1,255 @@
...
+  
+  
+#if defined(RAJA_ENABLE_OPENMP)
+  printf("RAJA: OpenMP/Sequential Policy - forallN \n");
+  /*
+    Here the outer loop is excuted in parallel while the inner loop
+    is executed sequentially
+  */
+  RAJA::forallN<RAJA::NestedPolicy
+    <RAJA::ExecList<RAJA::omp_parallel_for_exec, RAJA::seq_exec>>>
```
willkill07: I think this is good! If it's in scope, having an example with tiling might be nice. Given that nested::forall is a work in progress I wouldn't invest too much time in forallN examples.

examples/wave.cpp
```diff
@@ -0,0 +1,271 @@
...
+    / (grid.dx * grid.dx);          // Merge coefficients into a single coefficient
+
+
+  /*
+    Predefined Nested Policies
+  */
+
+  // Sequential
+  using fdPolicy = 
+    RAJA::NestedPolicy<RAJA::ExecList
```
willkill07: This formatting looks a little weird

examples/wave.cpp
```diff
@@ -0,0 +1,271 @@
...
+    <RAJA::seq_exec, RAJA::seq_exec > >;
+    
+  // OpenMP
+  //using fdPolicy =
```
willkill07: Extraneous commented code can be removed

examples/CMakeLists.txt
davidbeckingsale: 👍

examples/example-add-vectors.cpp
```diff
@@ -0,0 +1,164 @@
...
+  printf("RAJA: Sequential Policy \n");
+  /*
+    RAJA::seq_exec -  Executes the loop sequentially
+
+    RAJA::RangeSegment(start,stop) - Generates a contiguous sequence of numbers
+    by the [start, stop) interval specified
+  */
+  RAJA::forall<RAJA::seq_exec>
+    (RAJA::RangeSegment(0, N), [=](RAJA::Index_type i) {
+      C[i] = A[i] + B[i];
```
willkill07: It is more appropriate to have the parenthesis on the line above than on the line below.
```cpp
RAJA::forall<RAJA::seq_exec>(
  RAJA::RangeSegment(0, N), [=] (RAJA::Index_type i) {
    C[i] = A[i] + B[i];
  });
```

is arguably cleaner. This is also closer to what autoformatting tools, such as `clang-format`, will yield

examples/example-jacobi.cpp
```diff
@@ -0,0 +1,296 @@
...
+int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
+{
+
+  printf("Example 3: Jacobi Method \n");
+
+  // Jacobi method is invoked until the l_2 norm of the difference of the
+  // iterations are under a tolerance
+  double tol = 1e-5;
+
+  int N       = 100;                // Number of unknown gridpoints per cartesian dimension
```
davidbeckingsale: This is a nitpick, but can you remove all the extra alignment spacing here and below - I think we want the RAJA code to have a uniform feel across the project.

examples/memoryManager.hpp
```diff
@@ -0,0 +1,38 @@
+#ifndef __MEMORYMANAGER_HPP__
+#define __MEMORYMANAGER_HPP__
+
+#include "RAJA/RAJA.hpp"
+#include "RAJA/util/defines.hpp"
+
+/*
+  Developer must manage memory
+*/
+namespace memoryManager{
```
davidbeckingsale: Can you comment this file fully, it would be a good place to explain how RAJA interacts with data.

artv3: Thanks @davidbeckingsale, I think I got them all now. @trws @rhornung67, if there is anymore comments or suggestions please let me know.

artv3:
I made the forall loop description a bit more general. 

  RAJA::forall<exec_policy>(iter_space I, [=] (index_type i)) {                                           
                                                                                                          
         //body                                                                                           
                                                                                                          
  });     

  [=] By-copy capture                                                                                     
  [&] By-reference capture                                                                                
  exec_policy - Specifies how the traversal occurs                                                        
  iter_space  - Iteration space for RAJA loop (any random access container is expected)                   
  index_type  - Index for RAJA loops

trws:
For the forall description, that works, but it might be better to start recommending the by-value style since that works everywhere, while the by type version only works for some policies.  By which I mean that would become

```c++
RAJA::forall(exec_policy{}, [=] (RAJA::Index_type i) {
  //body
}
```

It might even be nice to recommend using auto as an index type, but I haven't tried that across all of our supported compilers so probably wait on that.

artv3:
@trws, I see. Should I simply omit the  [&] By-reference capture description and leave it for the docs? Or simply add a note saying [&] By-reference capture description has limited support?

trws:
Either is probably fine, if you go for the latter, consider saying 
something like "limited support on offload targets" or something 
"non-unified memory targets" or some such thing.

artv3:
Thanks @trws , I ended up adding the "non-unified memory target" remark.

examples/example-add-vectors.cpp
```diff
@@ -0,0 +1,164 @@
...
+  checkSolution(C, N);
+#endif
+  
+  
+#if defined(RAJA_ENABLE_CUDA)
+  printf("RAJA: CUDA Policy \n");
+  /*
+    RAJA::cuda_exec<CUDA_BLOCK_SIZE> - excecutes loop using the CUDA API
+    
+    CUDA_BLOCK_SIZE - specifies the number of threads in a CUDA thread block
```
rhornung67: I would move the comment about the block size up to where it is defined above.

examples/example-add-vectors.cpp
```diff
@@ -0,0 +1,164 @@
...
+  memoryManager::deallocate(A);
+  memoryManager::deallocate(B);
+  memoryManager::deallocate(C);
+
+  return 0;
+}
+
+
+/*
+  RAJA may also be used outside of computational kernels
```
rhornung67: Not sure what is meant by this comment. Also, I would add a comment about what this method checks to where it is declared above.  Basically, explain that the correct result for the vector addition is a vector of all zeros (c[i] = a[i]+b[i], where for each i, a[i] = 1, b[i] = -1).  That way folks will know immediately what the example is doing and they won't have to reverse engineer any code to understand it.

examples/example-add-vectors.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+const int CUDA_BLOCK_SIZE = 256;
+
+void checkSolution(int *C, int in_N);
+
+/*
+  Example 1: Adding Two Vectors
+
+  ----[Details]---------------------
+  Adds two vectors of length N
```
rhornung67: Here is where I would describe the details of the vector addition (as I've noted in the comment below).

examples/example-custom-index.cpp
```diff
@@ -0,0 +1,161 @@
...
+  Example 5: Custom Index Set
+
+  ----[Details]-------------------
+  Assuming a grid with the following contents
+
+  grid = [1, 2, 1, 2,
+          3, 4, 3, 4,
+          1, 2, 1, 2,
+          3, 4, 3, 4];
+
```
rhornung67: I would explain the details of the numerics that this example performs here -- probably pseudo-code is easiest to understand.  Then, it's clear from the start what's going on in this file.

examples/example-custom-index.cpp
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  delete[] idx;
+
+
+  /*
+    -----[RAJA Loop Traversal]-------
+    Under the custom color policy, a RAJA forall loop will transverse
+    through each list segment stored in the colorset.
+    This particular traversal policy willl execute the iterates
```
rhornung67: This sentence is confusing. My interpretation is that 'executing the iterates in each segment' is the same as 'traversing through each segment'. More clearly and concisely stated: iterate over the segments sequentially and traverse each segment in parallel.

Also 'travesing' -> 'traversing' (spelling)

examples/example-jacobi.cpp
```diff
@@ -0,0 +1,324 @@
...
+  to discretize 
+  
+  -U_xx - U_yy = 1.0.
+  The solution is assumed to satisfy
+  U = 0 on the boundary of the domain.
+  The domain is assumed to be a lattice 
+  with unit grid spacing.
+
+  Values inside the domain are computed
+  using the Jocabi method to solve the associated
```
rhornung67: 'Jocabi' -> 'Jacobi' (spelling)

examples/example-jacobi.cpp
```diff
@@ -0,0 +1,324 @@
...
+        
+        int id = n * (N + 2) + m;
+        I[id] = invD * (f - Iold[id - N - 2] - Iold[id + N + 2]
+                        - Iold[id - 1] - Iold[id + 1]);
+      });
+
+    /*
+      ----[Reduction step]---------
+      The RAJA API introduces a thread-safe accumulation variable
+      "ReduceSum" in order to carryout reductions
```
rhornung67: 'carryout' -> 'carry out' (two words). Better: replace with 'perform'.

examples/example-matrix-multiply.cpp
```diff
@@ -0,0 +1,258 @@
...
+const int CUDA_BLOCK_SIZE_Y = 16;
+
+template <typename T>
+void checkSolution(RAJA::View<T, RAJA::Layout<DIM>> Cview, int N);
+
+/*
+  Example 2: Multiplying Two Matrices
+
+  ----[Details]--------------------
+  This example illustrates how RAJA may be
```
rhornung67: Not sure how to interpret this comment. I think it would be better to clearly state what the example is doing.

examples/example-wave.cpp
```diff
@@ -0,0 +1,271 @@
...
+      P1[id] = 2 * P_curr - P_old + ct * lap;
+      
+    });
+}
+
+double waveSol(double t, double x, double y);
+void setIC(double *P1, double *P2, double t0, double t1, grid_s grid);
+void computeErr(double *P, double tf, grid_s grid);
+
+/*
```
rhornung67: I think it's better to explain the example up top before any substantial code.  Also, explain the RAJA features used in the example here as well.

artv3:
Thanks @rhornung67 , I reorganized the documentation and focused on better describing the algorithm being used and how RAJA features are incorporated in the code. Hopefully this structure is better. I suspect I may need to proof read a bit more but any comments or suggestions are more than welcomed @davidbeckingsale , @willkill07  @trws .

artv3:
Do RAJA views have range checking?

davidbeckingsale:
Nope, but we could write one that does. This is for your other project?

trws:
They intentionally do not, for performance reasons.  The checks tend not to optimize out, or prevent other optimizations from happening.

rhornung67:
IMO, this is best handled by developing a strategy for peppering the code with error checking macros so check code is active only for debug builds, otherwise compiled out. I'll add a JIRA issue.

artv3:
Thanks, I don't need range checking I was just curious. Aside from proof reading, is there anything else I can do for this branch?

artv3:
@davidbeckingsale, I added a 6th example ( Gauss Seidel with red black ordering for the PDE in example 3) I was thinking this could replace example 5, what do you think? Lastly, if there is anything else I can do to get this merged please let me know.

davidbeckingsale:
I think this is good to go. Like @rhornung67 suggested yesterday, we may want to clean out some of the comments in the file once the tutorial doc is done, but for now, let's get this in.

davidbeckingsale:
@trws are you happy with the state of this branch now? I don't want to dismiss your review 😄

davidbeckingsale:
Agree, we should just squash it.

PR #289, opened 2017-07-27
Start using BLT

davidbeckingsale:


rhornung67:
@davidbeckingsale does the nvcc debug build use thrust for scans?

CMakeLists.txt
```diff
@@ -53,20 +53,30 @@ set(RAJA_VERSION_PATCHLEVEL 0)
...
-option(RAJA_ENABLE_CUDA "Build CUDA support" Off)
-option(RAJA_ENABLE_CLANG_CUDA "Use Clang's native CUDA support" Off)
-set(RAJA_CUDA_ARCH "sm_35" CACHE STRING "Compute architecture to pass to CUDA builds")
-option(RAJA_ENABLE_CUB "Use cub for scans using CUDA" On)
-option(RAJA_ENABLE_TESTS "Build tests" On)
-option(RAJA_ENABLE_EXAMPLES "Build simple examples" On)
-option(RAJA_ENABLE_NESTED "Enable nested loop support" Off)
-option(RAJA_ENABLE_WARNINGS "Enable warnings as errors for CI" Off)
-option(RAJA_ENABLE_DOCUMENTATION "Build RAJA documentation" Off)
+set(ENABLE_OPENMP On CACHE Bool "Build OpenMP support")
```
willkill07: is there a reason why this is changed from option to set? (same for CUDA and COPY_HEADERS)

Also, what does COPY_HEADERS do? (it's undocumented)

willkill07:
@rhornung67 yes. It defaults to using CUB if found first, else it resorts to thrust

CMakeLists.txt
rhornung67: @willkill07 FWIW, I think the only reason copy headers exists in BLT is because BLT was developed as the Axom build system and BLT was extracted as a standalone capability. We don't ike it and we've been debating about how to change our directory structure to avoid it and then it would probably go away.
@davidbeckingsale do we really need to use COPY_HEADERS?

davidbeckingsale: We don't need copy headers - I can delete it.

The reason they have changed to `set` is because BLT already defines them as `option`. The `set` here just alters the default value. Leaving them as `option` makes things all wonky because of how the CMake cache works.

willkill07:
Alright. I'm really confused. I modified the .travis.yml file to make sure that CMAKE_EXTRA_FLAGS was only set once for clang-3.8-debug, but the log from Travis CI still shows the old config file being used

https://travis-ci.org/LLNL/RAJA/jobs/258371523#L591

davidbeckingsale:
@rhornung67 this now has clang-cuda support with `-DENABLE_CUDA=On -DENABLE_CLANG_CUDA=On`. Can you test it out?

rhornung67:
If I can get through a few time-critical things before my meetings start this afternoon, I'll try it before I leave today. Otherwise, I'll try it this evening or first thing in the morning.

willkill07:
@davidbeckingsale  I have a resolved merge branch on my machine, but I get a lot of warnings when compiling when CHAI is enabled for **every** single test file:

```
In file included from /g/g19/killian4/RAJA/chai/include/gtest/internal/gtest-internal.h:40:0,
                 from /g/g19/killian4/RAJA/chai/include/gtest/gtest.h:58,
                 from /g/g19/killian4/RAJA/test/unit/test-integral-limits.cpp:47:
/g/g19/killian4/RAJA/chai/include/gtest/internal/gtest-port.h:794:0: warning: "GTEST_HAS_DEATH_TEST" redefined
 # define GTEST_HAS_DEATH_TEST 1
 ^
<command-line>:0:0: note: this is the location of the previous definition
```

When BLT tries to do a final link, it complains about not being able to find -lCHAI when the static library, libchai.a, is already in the source list.

1). Thoughts?
2). Should I push the resolved branch up to github?

davidbeckingsale:
Can you update the BLT submodule in CHAI to point at current master and try again? And do the same for this branch. Let me know if that resolves it.

rhornung67:
@davidbeckingsale shouldn't @willkill07 switch those blt submodules to the clang-cuda branch of BLT and pull in the latest changes rather than master?  Same issue we discussed over the phone a few minutes ago...

davidbeckingsale:
Yup, @rhornung67 that's right.

willkill07:
Steps I've taken:

1). update `blt` in *both* RAJA and CHAI to branch `feature/clang-cuda`
2). established a sensible install location for CHAI: `$CHAI_INSTALL_DIR`
3). CHAI: `cmake .. -DCMAKE_INSTALL_PREFIX=$CHAI_INSTALL_DIR -DCUDA_TOOLKIT_ROOT_DIR=/opt/cudatoolkit/8.0 && make && install`
4). RAJA: cmake .. -DENABLE_CUDA=On -DCMAKE_PREFIX_PATH=$CHAI_INSTALL_DIR -DENABLE_CHAI=On -DCUDA_TOOLKIT_ROOT_DIR=/opt/cudatoolkit/8.0 && make`

I still get the same output above. Including the desire for the linker to want `-lchai`

I've pushed this to a `temp/blt` branch

davidbeckingsale:
Annoying. Thanks for checking it out. Do you run into any problems without CHAI? It seems to me that CHAI shouldn't be installing those  gtest headers, but since gtest is bundled up with BLT, it does. If you just go in and delete those files from `$CHAI_INSTALL_DIR` does it fix things?

willkill07:
@davidbeckingsale that resolves the multiple inclusions of google test 👍

Still odd that cmake wants to look for `-lCHAI`

willkill07:
I figured it out -- want me to update the branch here?

davidbeckingsale:
Sure thing, what was the problem?

willkill07:
the check for CHAI:
```cmake
if (ENABLE_CHAI)
  set (raja_depends
    ${raja_depends}
    CHAI)
endif ()
```
needed to be:
```cmake
if (ENABLE_CHAI)
  set (raja_depends
    ${raja_depends}
    chai)
endif ()
```

Capitalization does matter :(

davidbeckingsale:
Weird that this (CHAI) worked without BLT in RAJA....

davidbeckingsale:
We removed it because it was brittle with some compilers (*cough* Intel *cough*) and making sure the test was compiled with the correct flags.

davidbeckingsale:
@rhornung67 can you give this a try, it's up to date with `develop`.

rhornung67:
Will do. I will report back in a bit.

rhornung67:
Will do!

davidbeckingsale:
@rhornung67 I'm working on this and the perfsuite, think I've got it working again...

davidbeckingsale:
I'd like other people to give it a spin... @artv3 @trws @ajkunen

willkill07:
If TBB is disabled by default, it should still be enabled in Travis for our tests (since we DO install tbb)

davidbeckingsale:
Hmmm, Travis failed for every compiler except Intel when TBB was enabled. I'll take another look.

trws:
It looks like the other compilers aren't linking tbb for some reason, even if it's enabled.

trws:
Currently it breaks with cub headers unfound when cub is missing.

trws:
Also doesn't link tbb when requested, found and available.

trws:
Dies with:

```
CMake Error at /usr/local/Cellar/cmake/3.9.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES) (found
  version "1.0")
  /usr/local/Cellar/cmake/3.9.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:377 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/Cellar/cmake/3.9.2/share/cmake/Modules/FindOpenMP.cmake:402 (find_package_handle_standard_args)
  blt/cmake/SetupCompilerOptions.cmake:53 (find_package)
  blt/SetupBLT.cmake:176 (include)
  CMakeLists.txt:58 (include)
```

When ENABLE_OPENMP is Off and only a C++ compiler is specified

include/RAJA/util/basic_mempool.hpp
```diff
@@ -403,7 +403,7 @@ class MemPool {
 
   void free(const void* cptr)
   {
-#if defined(RAJA_ENABLE_OPENMP) && defined(_OPENMP)
+#if defined(RAJA_ENABLE_OPENMP)
```
willkill07: Maybe @MrBurmark has some insight here, but I believe there was a special reason why defined(_OPENMP) was added to this

rhornung67: @willkill07 @trws had a good suggestion to use RAJA_ENABLE_OPENMP in the code for consistency with other PMs and make it an alias to _OPENMP. @davidbeckingsale created a JIRA issue for this.

MrBurmark: I believe the idea was to allow people to configure with OpenMP but only use non-OpenMP parts of RAJA and compile without OpenMP. @trws 's idea sounds good to me. I assume it will be aliased only when configuring with OpenMP, so we aren't potentially using OpenMP all the time.

trws:
I think where we need to go for it to work right now is you can only see 
openMP if you both configure and compile with it.  At some point, if we 
have no compiled component that depends on OpenMP, then we can just have 
it be based on compile-time options.

CMakeLists.txt
```diff
@@ -26,25 +26,38 @@ set(RAJA_VERSION_PATCHLEVEL 1)
...
-option(RAJA_ENABLE_TESTS "Build tests" On)
-option(RAJA_ENABLE_EXAMPLES "Build simple examples" On)
-option(RAJA_ENABLE_NESTED "Enable nested loop support" Off)
-option(RAJA_ENABLE_MODULES "Enable modules in supporting compilers (clang)" On)
-option(RAJA_ENABLE_WARNINGS "Enable warnings as errors for CI" Off)
-option(RAJA_ENABLE_DOCUMENTATION "Build RAJA documentation" Off)
-
-option(RAJA_ENABLE_COVERAGE "Enable coverage (only supported with GCC)" Off)
+
+option(ENABLE_TBB "Build TBB support" Off)
```
davidbeckingsale: @rhornung67  TBB is off by default now.

rhornung67:
@davidbeckingsale looks like develop needs to be merged in. After that is done and all check pass, I'll re-aprrove.

davidbeckingsale:
Done.

PR #295, opened 2017-07-30
add RAJA_SIMD support for many compilers

jeffhammond:
Use OpenMP 4 pragma simd if available.  If not, use the compiler-specific option for GCC, Intel, Clang, and IBM XLC.  Wherever possible, I implement precise preprocessor macro version tests.

This is based in part on https://github.com/jeffhammond/nwchem-tce-triples-kernels/blob/master/src/pragma_vendor.h.  It has been ~3 years since I used IBM XLC, but `ibm independent_loop` was not reliable when I tried it on Blue Gene/Q and POWER7 AIX.

I have not tested this anywhere and since I just learned about the Clang vectorize pragma today, I have never tested it anywhere.


jeffhammond:
This was inspired by email discussion with @trws...

DavidPoliakoff:
@rhornung67 : do we have an early version of the performance suite we could run this PR against?

@jeffhammond : thanks for the contribution, it looks good, just want to put it through its paces a little bit, make sure our use cases weren't (for whatever reason) very different from yours.

jeffhammond:
This change was worth 3x for PRK stencil on KNL. Will forward details in email.

Intel compiler treats SIMD pragma as assert not hint, which can be suboptimal when heavy gather-scatter required. However, you have seq_exec policy for cases where SIMD isn't appropriate. Let simd_exec also be an assertion insofar as it uses SIMD pragma unconditionally.

trws:
I think I'd like to add or adjust one thing to go with this, specifically either adding a sequential policy without the novector pragma or to remove it from sequential. You told me once @rhornung why sequential has novector on it, but IIRC it was only for performance testing rather than production. Is there a reason not to have that be a define flag or something?


---
Sent from VMware Boxer

On July 30, 2017 at 3:45:00 PM PDT, Jeff Hammond <notifications@github.com> wrote:
This change was worth 3x for PRK stencil on KNL. Will forward details in email.

Intel compiler treats SIMD pragma as assert not hint, which can be suboptimal when heavy gather-scatter required. However, you have seq_exec policy for cases where SIMD isn't appropriate. Let simd_exec also be an assertion insofar as it uses SIMD pragma unconditionally.

willkill07:
I like these changes, but GCC does not currently like `#pragma omp simd` on the loop nest. Every CI test that uses GCC is currently failing.

trws:
We can fix that by doing the indexing like we do for cuda. Specifically iterating over an integral type then indexing into the class rather than iterating on the class directly.

jeffhammond:
@willkill07 https://gcc.gnu.org/ml/gcc-bugs/2015-09/msg00740.html shows some history of this issue.

jeffhammond:
Sadly, this is busted in every release version of GCC (I just tested 7.1.0).

willkill07:
@jeffhammond I made a few small changes to see if we can get GCC to cooperate :)

ajkunen:
I agree with @trws, we should really have 3 policies:
1) Force SIMD   (i like simd_exec for this)
2) Let compiler decide (bare loop... we need a name for this?!?!)
3) Force no-SIMD  (i like seq_exec for this)

DavidPoliakoff:
yeah, I also instinctively called (2) bare_exec (BEAR_EXEC), but don't think it's right. native_exec is probably also wrong, undecorated? nohint?

ajkunen:
hope_for_the_best_exec

willkill07:
* `RAJA::seq_exec` -- force sequential
* `RAJA::default_exec` -- no annotations/directives specified
* `RAJA::simd_hint` -- use compiler-appropriate SIMD hints
* `RAJA::simd_force` -- bring out the guns and force SIMD upon all of us

trws:
I kinda like the c++17 names,

sequenced
parallel
parallel_unsequenced

Maybe we have an unseq, to say that we want it to run in one thread but that sequencing is not required?

DavidPoliakoff:
yolo_exec or riot!

trws:
That would be good with me, default for nothing, hint for omp simd, force for whatever the "compiler vectorize this @#%&*^!@$^ you"

willkill07:
@trws the only problem with parallel is that it's not descriptive enough given the policies that RAJA *could* have in the future. is it implied that OpenMP will be the only CPU parallel policy we leverage?

davidbeckingsale:
I almost agree with @willkill07's names, but I'm not sure if `default_exec` is the clearest...

trws:
@willkill07, I wasn't necessarily proposing parallel, but it would be "use openmp if you have it, or back off onto c++11 threads or cilk or whatever if not" if we added it.

ajkunen:
I think there is a more fundamental question about execution policies and segments/IndexSet's that came up when I added negative strides to RangeStrideSegment.

Does the execution policy determine ordering?
Does the segment determine ordering?

If you specify a negative stride for a RangeStrideSegment, you are denoting the execution order in the segment, but then saying seq_exec is also.  Perhaps that was a mistake?  I could also see that there is a reverse_seq_exec, and that RangeStrideSegment has a strictly positive stride.

This is a philosophical point that we haven't really hashed out yet.

trws:
Actually, to be more straightforward, I was proposing we have `seq_exec` and `unseq_exec` for sequential no-vectorize and sequential with no restricting directive.

davidbeckingsale:
I think we need two classes of policies: assertions and requests. E.g. `omp_parallel_for` is an assertion, `parallel` is a request.

DavidPoliakoff:
@ajkunen It has also come up in the context of the whole taskgraph segment thing, the need to determine exactly what a "segment" can be

trws:
It is.  I've always seen it as the Range or Segment is just a bag of items and the policy determines the ordering.  If we ever want to support forward iteration however, that will at least somewhat change.

FWIW, Kokkos combines these concepts into a single thing, for example:

```
Kokkos::RangePolicy<Kokkos::DefaultExecutionSpace>(1,n-1)
```

Where the execution space is not entirely unlike our policy types.

rhornung67:
@jeffhammond thanks for starting this ball rolling! This was something that @artv3 (new RAJA developer) was going to dig into soon. 

There are enough kernels in the initial draft of the RAJA performance suite to try this out. @artv3 will work from this branch to see where we stand.

@DavidPoliakoff @willkill07 @trws @ajkunen @davidbeckingsale if we have time at tomorrow's meeting to talk about this, we will discuss the simd exec policies (names, which we want to support, etc.)

ajkunen:
@DavidPoliakoff Here is one view, that I think I rather like:

Segments define, at a bare minimum, an set if iterates.  They may define some dependencies between them, which are only interpreted by an appropriate execution policy.

Execution Policies determine where and how the segment is executed.  Some policies may be defined to respect the dependency structure of the segment, and others don't.

For example:  A RangeSegment defines an set [begin, end) and an implied sequential ordering.  seq_exec honors that ordering, but cuda_exec does not.

The same could be said for a (yet to be invented) GraphSegment, where omp_for_exec does not honor that graphs ordering, but (another yet to be invented) omp_graph_for_exec does. 


My morning caffeine has been depleted..need to get more.

trws:
BTW: Travis is now passing on everything except ICC-17, seemingly because of a travis install issue.  @davidbeckingsale, @jeffhammond?

davidbeckingsale:
I think it has something to do with the secret travis env variable that has the license key in....

jeffhammond:
I vigorously support the proposal from @trws to use C++17 `seq` and `unseq` here, in addition to `simd`.  These would map to "tell the compiler not to vectorize in the strongest possible way", "let the compiler decide whether or not to vectorize", and "tell the compiler to vectorize in the strongest possible way".

@rhornung67 I'm done traveling for the year, but if you want my opinion, I would be happy to join your arguing party via some digital protocol.

@davidbeckingsale Having prescriptive options like `omp_parallel_for_exec` is interesting, although once you go down this path, you have to consider `omp_parallel_for_simd_exec`, `omp_for_exec`, `omp_for_simd_exec`, `omp_taskloop_exec`, etc.

jeffhammond:
@trws @willkill07 I think the only sensible fallback for RAJA if OpenMP does not exist is C++17 PSTL.  It's a poor use of developers' time to reimplement the OpenMP runtime - if you don't do this, you're not using a thread pool and the performance is likely to disappoint.

I have not done much testing, but I have implemented code that can demonstrate the value of a thread pool empirically:
* https://github.com/ParRes/Kernels/blob/master/Cxx11/transpose-vector-thread.cc#L125
* https://github.com/ParRes/Kernels/blob/master/Cxx11/transpose-vector-async.cc#L125

jeffhammond:
@willkill07 Thanks for that.  I am surprised that you were able to push a change to my fork.  That's a very useful feature in this case, but it is perplexing from a security perspective.

This is a digression from the current topic, but does this mean that any time I submit a PR, the developers of the upstream repo have commit access to the branch associated with that PR?  I sincerely hope that this is all the access that is granted.

trws:
There's a checkbox for whether or not to allow this on the PR submission 
now.  Perhaps it defaults to checked?

trws:
Agreed @jeffhammond.

rhornung67:
@jeffhammond totally agree about the potential explosion of exec policies and we've had some discussions about how to compose these things to ease this.  We have to be sensitive to the fact that our production apps teams prefer do be prescriptive as possible -- trusting an abstraction layer to do what they want is a steep grade to overcome. Also, they tend to build a code-specific layer over RAJA to make it fit their code constructs better so they can hide some of this.

jeffhammond:
@rhornung67 I cannot help myself...
1td7ou

rhornung67:
👍

davidbeckingsale:
@artv3 you should check out #310 in conjunction with this

jeffhammond:
Clang pragma should use "assume_safety" instead of "enable". Will fix later
but noting here so I don't forget. Just remembered it and not near
terminal.
-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

jeffhammond:
See https://github.com/LLNL/RAJA/pull/310#issuecomment-323374175 and https://github.com/LLNL/RAJA/pull/310#issuecomment-323375448 regarding the need for more careful treatment of reductions here.

davidbeckingsale:
Superseded by #311

PR #297, opened 2017-08-02
TBB backend for RAJA

trws:
Nothing fancy, but a working TBB backend for RAJA.  As a sidenote, the TBB reducers also work for OpenMP threads as long as OpenMP is using OS threads, which is rather nice.

Remaining to be done:
- [x] find or fetch TBB in cmake
- [x] add more tests
- [x] add is_tbb_policy type trait
- [x] make (at least some) raja segments TBB range compatible (may also help with nested::forall tiling)

(poke: @jeffhammond )


artv3:
Is there any interest in OpenCL? I started making a wrapper for it as a weekend project.

trws:
Generically I would say yes, but how would you make it work without having to put the kernels in a separate file? Are you looking at SYCL or..?

jeffhammond:
@trws https://github.com/justusc/FindTBB might help.  [MADNESS](https://github.com/m-a-d-n-e-s-s/madness/) has been using TBB for a while, including on IBM Blue Gene/Q, so I guess Justus' implementation is pretty reliable.

trws:
Yup, just grabbed that. It seems pretty nice so far.


---
Sent from VMware Boxer

On August 2, 2017 at 11:45:32 AM PDT, Jeff Hammond <notifications@github.com> wrote:

@trws<https://github.com/trws> https://github.com/justusc/FindTBB might help. MADNESS<https://github.com/m-a-d-n-e-s-s/madness/> has been using TBB for a while, including on IBM Blue Gene/Q, so I guess Justus' implementation is pretty reliable.

jeffhammond:
TBB `blocked_range2d` works well with the PRKs.  At least with OpenMP, RAJA `forallN` performs worse than the explicitly nested version (https://github.com/ParRes/Kernels/blob/master/Cxx11/transpose-vector-raja.cc#L207).  It's possible that I'm doing something wrong though.  Perhaps TBB will fix this.

jeffhammond:
I haven't looked a RAJA taskgraph yet but TBB should have a bunch of stuff you'll find useful for that, including OpenCL support (as TBB flowgraph nodes).

include/RAJA/policy/tbb/scan.hpp
```diff
@@ -0,0 +1,158 @@
...
+******************************************************************************
+*
+* \file
+*
+* \brief   Header file providing RAJA scan declarations.
+*
+******************************************************************************
+*/
+
+#ifndef RAJA_scan_sequential_HPP
```
willkill07: I don't think this is the TBB scan file

include/RAJA/policy/PolicyBase.hpp
```diff
@@ -59,7 +59,7 @@
 namespace RAJA
 {
 
-enum class Policy { undefined, sequential, simd, openmp, target_openmp, cuda };
+enum class Policy { undefined, sequential, simd, openmp, target_openmp, cuda, tbb };
```
willkill07: If we add a Policy, we should also add corresponding type traits (is_tbb_policy)

include/RAJA/policy/tbb/reduce.hpp
```diff
@@ -0,0 +1,388 @@
...
+#include <memory>
+#include <tuple>
+
+namespace RAJA
+{
+
+namespace detail
+{
+template <typename T, typename Op>
+class ReduceTBB
```
willkill07: I like this structure

include/RAJA/policy/tbb/scan.hpp
trws: Ugh, indeed not.  I still need to do the scans, but TBB has optimized primitives for it so it should be a simple mapping.

include/RAJA/policy/PolicyBase.hpp
trws: Good point, will do.

trws:
By the by, this PR is more an experiment than an actual merge request at the moment.  Some numbers from @jeffhammond got me wondering:

```
+ numactl -m 1 ./stencil-vector-pstl 40 10000 grid 4
Rate (MFlops/s): 490050 Avg time (s): 0.0332087
+ numactl -m 1 ./stencil-vector-openmp 40 10000 grid 4
Rate (MFlops/s): 475937 Avg time (s): 0.0341935
+ numactl -m 1 ./stencil-kokkos 40 10000 grid 4
Rate (MFlops/s): 455751 Avg time (s): 0.035708
+ numactl -m 1 ./stencil-vector-raja 40 10000 grid 4
Rate (MFlops/s): 454744 Avg time (s): 0.0357871
```

Something here doesn't add up.  RAJA and Kokkos being behind OpenMP? That I
buy, RAJA, Kokkos _and OpenMP_ being behind pstl, which is apparently built on
TBB, that makes me wonder.  The main goal here is to get an idea of whether
this anomaly is something inherent in the pstl or if there is some win in TBB
that OpenMP is missing.

jeffhammond:
I'll upload more detailed data that I think will help explain some of this...

jeffhammond:
Ignoring the noise in the TBB line, which I attribute to a suboptimal tilesize, we see two trends here.  The OpenMP results are essentially flat and the ones based on work-stealing runtimes - Cilk and TBB - are worse below 5000 and better above it.

image

# Details
* PRK stencil with pattern=grid and radius=4 (not the default)
* Processor: Intel® Xeon Phi™ CPU 7250 @ 1.40GHz
* Memory: MCDRAM-only in quadrant flat mode (`numactl -m 1`)
* Compiler: Intel C/C++ 18.0.0 20170510 (version 18 beta update 1)
* Settings: 1000 iterations of each after a dummy job to warm up the node
* OpenMP affinity: `OMP_NUM_THREADS=68`, `OMP_PROC_BIND=spread`,  `OMP_PLACES=threads`
* OpenMP taskloop grainsize of 32 (still tuning this)
* TBB tilesize of 500 (still tuning this)

trws:
Thanks @jeffhammond, that does help clarify. The question now is whether intel 18 has support for "omp parallel for nonmonotonic schedule(dynamic, <tile size>)" and if so how that stacks up. The extra keyword makes it unambiguously safe to implement dynamic or guided with work-stealing instead of a centralized work queue, and it's going to be the default in 5.0.


---
Sent from VMware Boxer

On August 3, 2017 at 6:15:07 AM PDT, Jeff Hammond <notifications@github.com> wrote:

Ignoring the noise in the TBB line, which I attribute to a suboptimal tilesize, we see two trends here. The OpenMP results are essentially flat and the ones based on work-stealing runtimes - Cilk and TBB - are worse below 5000 and better above it.

[image]<https://user-images.githubusercontent.com/406118/28922951-01670d0e-7811-11e7-80d7-286084967cc3.png>

Details

  *   PRK stencil with pattern=grid and radius=4 (not the default)
  *   Processor: Intel® Xeon Phi™ CPU 7250 @ 1.40GHz
  *   Memory: MCDRAM-only in quadrant flat mode (numactl -m 1)
  *   Compiler: Intel C/C++ 18.0.0 20170510 (version 18 beta update 1)
  *   Settings: 1000 iterations of each after a dummy job to warm up the node
  *   OpenMP affinity: OMP_NUM_THREADS=68, OMP_PROC_BIND=spread, OMP_PLACES=threads
  *   OpenMP taskloop grainsize of 32 (still tuning this)
  *   TBB tilesize of 500 (still tuning this)

jeffhammond:
@trws `nonmonotonic` is not implemented in Intel 18u1 😢

jeffhammond:
@trws You meant `omp parallel for schedule(nonmonotonic, dynamic, <tile size>)` 😉

trws:
Ugh, yes I did sorry. ><

Does that make any performance difference yet?


---
Sent from VMware Boxer

On August 3, 2017 at 9:49:25 AM PDT, Jeff Hammond <notifications@github.com> wrote:

@trws<https://github.com/trws> You meant omp parallel for schedule(nonmonotonic, dynamic, <tile size>) 😉

jeffhammond:
I don't see any interesting change with the following change in the OpenMP implementation, other than a loss of performance of ~10% across the board.
```c
# define OMP_FOR(x) PRAGMA(omp for x schedule(nonmonotonic,dynamic,32) )
```

image

trws:
Ok, that almost certainly means the runtime hasn't been updated to take advantage of it. Dynamic without stealing should be a bit of a drop. I'll poke around and see if I can find an implementation that actually does something with it.


---
Sent from VMware Boxer

On August 3, 2017 at 10:30:54 AM PDT, Jeff Hammond <notifications@github.com> wrote:

I don't see any interesting change with the following change in the OpenMP implementation, other than a loss of performance of ~10% across the board.

# define OMP_FOR(x) PRAGMA(omp for x schedule(nonmonotonic,dynamic,32) )

[image]<https://user-images.githubusercontent.com/406118/28934607-6ff79656-7836-11e7-82a7-91277ae444f5.png>

trws:
So, first off, 🥂 

Putting my OpenMP hat on for a moment 🎩, what on earth? The only thing I can think of that would be different is that the Kokkos and RAJA versions use parallel for rather than a large parallel region.  Maybe that's making a difference for locality or..?

jeffhammond:
This emoji - 🤷‍♂️ - was created specifically for times like this.  I cannot explain this but I'll keep digging.  Obviously, I can implement OpenMP the Wrong Way to see if that's the issue.

DavidPoliakoff:
Seeing things like this makes me think the multipolicy is going to get a heck of a workout in the future :)

jeffhammond:
New data for KNL (top) and SKX (bottom).  OpenMP uses every hardware thread, which matches what Cilk and TBB (and therefore PSTL) do.  OpenMP `for` uses static scheduling.  The TBB block size and taskloop grainsize are 32.

image

image

I will now see what I can do with Cilk and TBB/PSTL using one thread per core, but that requires source changes for TBB because TBB doesn't have an env var for that (Cilk does).

Cilk is not shown for KNL because the C++ results are anomalous.  The C version does well but I haven't had time to generate a full set of data yet.

This is all work-in-progress and I will likely learn something new from the TBB folks that improves those results soon.  I know a lot more about the runtime knobs for OpenMP than I do for TBB...

trws:
Interesting, I wonder what kokkos is doing different on SKX. For TBB, does it honor the cpu sched mask? That might be a practical way to test it with a thread limit if it's reading available cores.


---
Sent from VMware Boxer

On August 4, 2017 at 5:04:51 PM PDT, Jeff Hammond <notifications@github.com> wrote:

New data for KNL (top) and SKX (bottom). OpenMP uses every hardware thread, which matches what Cilk and TBB (and therefore PSTL) do. OpenMP for uses static scheduling. The TBB block size and taskloop grainsize are 32.

[image]<https://user-images.githubusercontent.com/406118/28990845-2551c972-7936-11e7-9241-59b4b9b62472.png>

[image]<https://user-images.githubusercontent.com/406118/28990846-29760612-7936-11e7-9a41-ebdc670c85d3.png>

I will now see what I can do with Cilk and TBB/PSTL using one thread per core, but that requires source changes for TBB because TBB doesn't have an env var for that (Cilk does).

Cilk is not shown for KNL because the C++ results are anomalous. The C version does well but I haven't had time to generate a full set of data yet.

This is all work-in-progress and I will likely learn something new from the TBB folks that improves those results soon. I know a lot more about the runtime knobs for OpenMP than I do for TBB...

jeffhammond:
I tried TBB with taskset but it didn't help. 

Kokkos uses a standard inner loop so maybe it gets a boost from better compiler SIMD. I'll experiment more.

trws:
Assuming travis likes this last version, this backend is in relatively good shape.  It lacks some of the forallN improvements it should have, but @jeffhammond has submitted a good bit of that in another PR.  The only thing I would still call "missing" is an ordered reduction implementation.  TBB has an ordered reduction, but we can't actually use it, so it would have to be another ad-hoc affair most likely.

trws:
Some more to do here apparently.  Unfortunately I can't seem to repro the offset issue off of travis just yet.  Will look into this again later.

trws:
Ok, now this is ready for review.  @davidbeckingsale, @rhornung67?

jeffhammond:
Does it make sense to merge the forallN changeset and review the whole lot at once? My changes are very straightforward...

trws:
Actually, it probably does. I'll merge your changes into this PR when I get back to a computer.


---
Sent from VMware Boxer

On August 12, 2017 at 9:19:26 AM PDT, Jeff Hammond <notifications@github.com> wrote:
Does it make sense to merge the forallN changeset and review the whole lot at once? My changes are very straightforward...

jeffhammond:
@trws I don't know what you all use internally for performance drag races, but I modularized PRK Transpose to exercise a reasonable fraction of the `forallN` exec space in https://github.com/ParRes/Kernels/pull/258.

trws:
Thanks @jeffhammond, I'll be interested to see how this all stacks up.

Also, I think this is all green at this point.

travis/install_llvm.sh
```diff
@@ -0,0 +1,14 @@
+export LLVM_PATH=${HOME}/llvm/clang+llvm-${LLVM_VERSION}-x86_64-linux-gnu-ubuntu-14.04
```
davidbeckingsale: Let's put this in the `scripts` directory, just to keep the root of the repo cleaner. We should add git pre and post hooks to their also at some point...

trws: Fair point, I'll do that now.

davidbeckingsale: Nice! Also, I think that since this is a large PR we should wait for @rhornung67 before merging.

DavidPoliakoff: @trws and then integrate Apollo, and then write the policy raja::best_threaded_exec...

CMakeLists.txt
```diff
@@ -54,6 +54,7 @@ set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/thirdparty" ${CMAKE_MODULE_PA
 
 # Build options
 option(RAJA_ENABLE_OPENMP "Build OpenMP support" On)
+option(RAJA_ENABLE_TBB "Build TBB support" On)
```
rhornung67: Do we want this on by default?  It seems exploratory at this point.

cmake/SetupPackages.cmake
```diff
@@ -81,6 +81,17 @@ if (RAJA_ENABLE_CUDA)
 endif()
 
 
+if (RAJA_ENABLE_TBB)
```
rhornung67: @davidbeckingsale will this be changed to 'ENABLE_TBB' for consistency when we switch to BLT?

include/RAJA/policy/PolicyBase.hpp
rhornung67: 👍

include/RAJA/policy/tbb/forall.hpp
```diff
@@ -0,0 +1,173 @@
...
+//
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#include "RAJA/config.hpp"
+
+#include "RAJA/util/types.hpp"
+
+#include "RAJA/policy/tbb/policy.hpp"
+
+#include "RAJA/index/IndexSet.hpp"
```
rhornung67: Would it be better to remove inclusion of IndexSet and Segment headers in the policy files? It seems like we should have an 'Index.hpp' file that includes all these (for convenience); users should be responsible for including what they need.

If folks agree, we should probably fo this on a subsequent PR.

include/RAJA/policy/tbb/reduce.hpp
```diff
@@ -0,0 +1,407 @@
...
+   */
+  void combine(const T &other) { data->local() = Op{}(data->local(), other); }
+  /*!
+   *  \return update the local value
+   */
+  void combine(const T &other) const
+  {
+    data->local() = Op{}(data->local(), other);
+  }
+};
```
rhornung67: Can we add a blank line here and some below to separate classes/structs and make it easier to visually parse?

include/RAJA/policy/tbb/scan.hpp
```diff
@@ -0,0 +1,239 @@
...
+
+#include "RAJA/util/concepts.hpp"
+
+#include "RAJA/policy/sequential/policy.hpp"
+
+#include <tbb/tbb.h>
+
+#include <algorithm>
+#include <functional>
+#include <iostream>
```
rhornung67: Do we need to include iostream?

include/RAJA/util/Operators.hpp
```diff
@@ -386,7 +386,7 @@ struct logical_and : public detail::comparison_function<Arg1, Arg2>,
   {
     return lhs && rhs;
   }
-  static constexpr const bool identity = true;
+  static constexpr bool identity() {return true;}
```
rhornung67: Why do we need identities for this and some of the operators below? What do they mean?

test/unit/cpu/test-forall.cpp
```diff
@@ -104,3 +105,16 @@ using OpenMPTypes = ::testing::Types<
 
 INSTANTIATE_TYPED_TEST_CASE_P(OpenMP, ForallTest, OpenMPTypes);
 #endif
+
+#if defined(RAJA_ENABLE_TBB)
+using TBBTypes = ::testing::Types<
+    ExecPolicy<seq_segit, tbb_for_exec>,
+    ExecPolicy<tbb_for_exec, seq_exec>,
```
rhornung67: Should we make an alias 'tbb_segit', etc. for consistency with others?

travis/install_llvm.sh
rhornung67: Uhhh, let's not...

CMakeLists.txt
trws: As of now it compiles with TBB if it finds it, but if it doesn't find it  it doesn't fetch TBB or fail but prints a warning. We can turn it off by default, my assumption was that if they make TBB available would be likely to want it.

include/RAJA/policy/tbb/forall.hpp
trws: Now that you mention it, those shouldn't be necessary there anymore. Or in the others for that matter, if they are it should only be a forward declaration they need anyway, so that sounds like a good issue to track IMO.

include/RAJA/policy/tbb/scan.hpp
trws: Nope.  Thanks for catching that!

include/RAJA/policy/tbb/reduce.hpp
trws: Sure, I've added some spaces in the commit I'm pushing in a moment, let me know if it needs more.

CMakeLists.txt
rhornung67: I'm good either way.

test/unit/cpu/test-forall.cpp
trws: I have no objection to that, but had removed it since it was just an alias.  Suppose it's best to just be consistent for now and we can just deprecate them all as a set later maybe.

cmake/SetupPackages.cmake
trws: Just spoke with @davidbeckingsale, leaving this for now but it will probably change with the switch.

include/RAJA/policy/tbb/scan.hpp
rhornung67: Don't worry about trying to address all these things in this PR. We probably should set up a PR for wading through the code and making changes like this.

CMakeLists.txt
trws: Ok, for now I'm going to leave it, but if anyone runs into any issues with it making this off by default before it goes to master would be fine.

include/RAJA/util/Operators.hpp
trws: These were set up by @willkill07 with the operators in the first place.  They provide a default identity value for a given operator, which can be very useful.  I made this change because for some reason the constexpr values weren't being inlined in some cases, and weren't being built into the final binary either.  The function versions get generated correctly however.

rhornung67: I asked because some operators have an identity and some do not and it's clear what identity means for some (+, *, etc.), but it may not be for others (I can see why identity for && is 'true', but 'false' for || seems arbitrary)?  I'm fine with this if it makes sense to others.

trws: These are pretty standard, they're basically what OpenMP uses IIRC.  I'll look over the list and double check, but the ones I've looked at are good.  The issue with some is that there really isn't an identity value for the operator, which means it can't be used in certain situations without having one specified for the specific problem.

DavidPoliakoff: @rhornung67 These are mathematical identities, the identity in a group is supposed to be the thing under an operator such that when any other element is operated on with it, you get the thing you operated on (1 * anything = that thing, 0 + anything = that thing, true AND anything = that thing, false OR anything = that thing). Finally, a part of math matters at the lab that I know

rhornung67: Yeah, got it. I shouldn't try to review PRs while I'm answering emails and talking on the phone at the same time.

DavidPoliakoff: Only doing three things at once? Kind of a light day for you, then? :)

rhornung67: I would laugh if that weren't a painfully true statement...

trws:
Sadly, a light day for many of us, if not to the same extent ><

include/RAJA/util/Operators.hpp
jeffhammond: I was on two telecons at the same time earlier today.  Granted, I time-sliced the audio, because I've tried to do that in the past and it doesn't work very well.  The human brain is much better at visual concurrency than aural concurrency.

trws: I think I've only done that once.  Though I did set up and manage a webex for a meeting while physically in a different meeting another time...

include/RAJA/policy/tbb/fwd.hpp
```diff
@@ -0,0 +1,96 @@
...
+template <typename Func>
+RAJA_INLINE void forall(const tbb_for_exec &,
+                        const PolicyBase &,
+                        const RangeSegment &iter,
+                        Func &&loop_body);
+
+template <typename Iterable, typename Func>
+RAJA_INLINE void forall(const tbb_for_exec &,
+                        const PolicyBase &,
+                        Iterable &&iter,
```
willkill07: These forward definitions aren't defined anywhere. Policies are listed twice. (`tbb_for_exec`, `PolicyBase`)

test/unit/cpu/test-reductions.cpp
```diff
@@ -328,45 +334,32 @@ TYPED_TEST_P(NestedReductionCorrectnessTest, NestedReduceSum)
...
-                    RAJA::omp_reduce>,
-          std::
-              tuple<RAJA::
-                        NestedPolicy<RAJA::ExecList<RAJA::omp_parallel_for_exec,
-                                                    RAJA::seq_exec,
-                                                    RAJA::seq_exec> >,
-                    RAJA::omp_reduce>,
-          std::
-              tuple<RAJA::
-                        NestedPolicy<RAJA::
```
willkill07: Are there any clang-format rules that could help with this sort of mess?

include/RAJA/policy/tbb/fwd.hpp
trws: Thanks, I think this was basically a copy-and-paste error. Good catch.

test/unit/test-multipolicy.cpp
```diff
@@ -44,51 +44,87 @@
...
+}
+}
+
+// NOTE: this *must* be after the above to work
 #include "RAJA/RAJA.hpp"
 
+struct mp_test_body {
+  void operator()(mp_tag1 const &, std::size_t size) const
+  {
+    ASSERT_LT(size, 100);
```
willkill07: Since we are strict with type comparisons, we need to add `lu` suffix to `100` or explicitly cast to `std::size_t`. This is the only reason why travis is failing for GCC

test/unit/cpu/test-reductions.cpp
trws: It depends on what we want.  I think the major thing is how it decides to do the breaks, and that is configurable, but might take some finessing to get it right.  I'll poke at it a bit.

trws:
Quite so, thanks for pointing this out!

test/unit/cpu/test-reductions.cpp
trws: Turns out that the difference is that I'm using clang-format 5.0.0, which is the current version on homebrew.  Maybe we should try and standardize on that since the difference is no noticeable?

trws:
Think this is all covered now.  Re-review @davidbeckingsale, @rhornung67?

test/unit/cpu/test-reductions.cpp
rhornung67: @trws have you tried running clang-format with the clang 4.0 that we have on chaos and TOSS3 systems? I think that's the latest we have on those machines. The latest on our blueos machines is clang 5.0.

cmake/SetupPackages.cmake
```diff
@@ -101,6 +112,7 @@ else()
       CMAKE_ARGS                
           -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
           -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+          -DCMAKE_CXX_COMPILER_ARG1=${CMAKE_CXX_COMPILER_ARG1}
```
davidbeckingsale: What's this for?

include/RAJA/util/Operators.hpp
rhornung67: @jeffhammond it works if you have headphones on and can route the audio for each into a different ear :-)

cmake/SetupPackages.cmake
trws: This allows passing through the compile timing thing that I was using.  It's just a way to pass through something positional to a compiler or compiler wrapper.  It certainly doesn't have to stay in, but I didn't see a reason to keep it out.

test/unit/cpu/test-reductions.cpp
trws: Yeah, the old awful formatting of that block is what happens with 4.0.0 unfortunately.

cmake/SetupPackages.cmake
davidbeckingsale: Okay, that makes sense.

include/RAJA/policy/tbb/forall.hpp
willkill07: Would a forward declaration be enough? This short example fails to compile: https://godbolt.org/g/phjZ9B

scripts/install_llvm.sh
```diff
@@ -0,0 +1,14 @@
...
+export TARFILE=${HOME}/download/llvm-${LLVM_VERSION}.tar.xz
+if ! [[ -d "${LLVM_PATH}" ]]; then 
+    if ! [[ -f ${TARFILE} ]]; then
+        echo "curl -o ${TARFILE} ${DOWNLOAD_URL}"
+        curl -o ${TARFILE} ${DOWNLOAD_URL}
+    fi
+    cd ${HOME}/llvm 
+    tar xf ${TARFILE}
+fi
+
```
willkill07: 👍

trws:
Dox added, ready to merge @davidbeckingsale?

jeffhammond:
I'll add the forallN docs ASAP, hopefully tonight.

CMakeLists.txt
davidbeckingsale: Our other options shouldn't fallback like this (and definitely won't with BLT), i.e. if you ask for an option and we can't find it, CMake should fail...

cmake/SetupPackages.cmake
davidbeckingsale: This is fine.

include/RAJA/policy/tbb/forall.hpp
```diff
@@ -0,0 +1,200 @@
...
+      loop_body(static_cast<IndexType>(i + icount), begin[i]);
+  });
+}
+
+///
+/// TBB parallel for static policy implementation
+///
+
+/** 
+ * @brief TBB static for implementation
```
davidbeckingsale: Thanks for adding the docs 😄

test/unit/cpu/test-forall.cpp
davidbeckingsale: Did you add the alias?

test/unit/cpu/test-reduce.cpp
```diff
@@ -65,63 +65,73 @@ using namespace RAJA;
 
 
 using TestingTypes = ::testing::
-    Types<std::tuple<ExecPolicy<seq_segit, seq_exec>, seq_reduce>,
-          std::tuple<ExecPolicy<seq_segit, simd_exec>, seq_reduce>
+    Types<  // std::tuple<ExecPolicy<seq_segit, seq_exec>, seq_reduce>,
```
davidbeckingsale: Why are you commenting out these policies?

test/unit/cpu/test-reduce.cpp
```diff
@@ -65,63 +65,73 @@ using namespace RAJA;
...
   Index_type alen;
   RAJAVec<Index_type> is_indices;
   Real_ptr in_array;
+  Real_ptr test_array;
 
   virtual void SetUp()
   {
     alen = buildIndexSet(&iset, static_cast<IndexSetBuildMethod>(0)) + 1;
     in_array = (Real_ptr)allocate_aligned(DATA_ALIGN, alen * sizeof(Real_type));
+    test_array =
```
davidbeckingsale: 👍

test/unit/test-multipolicy.cpp
```diff
@@ -44,51 +44,87 @@
...
-              return 1;
-            }
-          });
-  RAJA::forall(mp, RAJA::RangeSegment(0, 5), [](RAJA::Index_type) {
-    ASSERT_EQ(omp_get_num_threads(), 1);
-  });
-  RAJA::forall(mp, RAJA::RangeSegment(0, 101), [](RAJA::Index_type) {
-    ASSERT_TRUE(omp_get_num_threads() > 1);
-  });
+  auto mp = RAJA::make_multi_policy<mp_tag1, mp_tag2>(
```
davidbeckingsale: I like the structure of this, but can you comment this file some to explain what's expected and how the assertions relate to the tests. (I can see what's going on, but it would be useful to have it recorded for posterity).

trws:
> Our other options shouldn't fallback like this (and definitely won't with BLT), i.e. if you ask for an option and we can't find it, CMake should fail...

So, as of now TBB is doing exactly what OpenMP does.  It defaults to on, but fails with a warning if it can't be found.  I would prefer that it be an error if the user explicitly requested it and its not found, but we don't seem to have a setup for that.  How would you like it changed?

trws:
> Did you add the alias?

If you mean TBBTypes then yes, just to declare the list of types to be used to instantiate the TBB tests.

test/unit/test-multipolicy.cpp
trws: Sure, no problem.

test/unit/cpu/test-reduce.cpp
trws: Because they are both expensive and slow without testing anything noticeably different than the others.  If you prefer we can make it an ifdef for something like "RAJA_TEST_EXHAUSTIVE", but especially the tests with parallel for on the inner loop are needlessly expensive.

davidbeckingsale: I thought that's what it would be. Remove them, rather than commenting.

davidbeckingsale:
I meant the alias for `tbb_segit`

trws:
I did add the tbb_segit alias back in yes, it's aliased to tbb_for_exec which defaults to static for now if it's supported.

trws:
Will do.

trws:
Huh... that's odd, somehow the approval got dismissed by pushing the removal of those comments.

davidbeckingsale:
Yup - I enabled that option after the test debacle....

trws:
Ah, good thinking.

trws:
Looks all green.  Merge?

PR #298, opened 2017-08-02
Revert single test executable

davidbeckingsale:
Note that the list of names can be acquired from this executable by doing:

`./test/raja_test_suite.exe --gtest_list_tests`

and subsets run with `--gtest_filter`, e.g.

`.//test/raja_test_suite.exe --gtest_filter="*Forall*"`


DavidPoliakoff:
Note the possible compromise of establishing executables/scripts which run the suite with various filters that make it behave like it used to.

This will make CI easier, but I'm agnostic on it

willkill07:
It doesn't sit well with me to undo the google-test refactor because of the way the cmake files yielded a single binary instead of a set. #299 solves this problem by only reverting the CMake configuration instead of ripping out the entire PR.

PR #299, opened 2017-08-02
Switch back to multiple test executables

willkill07:
Closes #298 

Instead of undoing all of the google-test refactor, this PR just updates the binary generation to it's previous state (one binary per test source file).

There are some performance implications with having the tests be separate binaries though -- on the order of 15% for compilation on `wiz` and 10% for execution.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -43,37 +43,6 @@
 
 include_directories(include)
 
-
-set(test_macro__internal_dir ${CMAKE_CURRENT_LIST_DIR} CACHE INTERNAL "")
-
-# use this macro to update the parent scope with the current list of sources used
-# for testing. NOTE: this is only required when the last command in a CMakeLists.txt
-# is add_directory(). add_gtest_sources() propagates the sources as expected
-macro(propagate_sources)
-  set (raja_test_SOURCES ${raja_test_SOURCES} PARENT_SCOPE)
-endmacro()
-
-# use this macro to add sources for testing. It will automatically add the
-# relative path (if required). NOTE: this internally updates @raja_test_SOURCES
-macro (add_gtest_sources)
-  file (RELATIVE_PATH _relPath "${test_macro__internal_dir}" "${CMAKE_CURRENT_SOURCE_DIR}")
-  foreach (_src ${ARGN})
-    if (_relPath)
-      list (APPEND raja_test_SOURCES "${_relPath}/${_src}")
-    else()
-      list (APPEND raja_test_SOURCES "${_src}")
-    endif()
-  endforeach()
-  if (_relPath)
-    propagate_sources()
-  endif()
-endmacro()
-
 add_subdirectory(unit)
 
 add_subdirectory(integration)
-
-# create the test once the sources have been visited
-raja_add_test(
-  NAME raja_test_suite
-  SOURCES ${raja_test_SOURCES})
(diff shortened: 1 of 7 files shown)


DavidPoliakoff:
I'm honestly a fan of leaving it as it is (one executable) and maybe having scripts that run that executable with filter arguments in order to get the old behavior

willkill07:
I figure I should at least quantify the changes instead of just state an opinion.

Under a compute node on rzwiz I configured RAJA with CUDA and CHAI support and built RAJA under the `develop` (single executable) and `enh/revert-multiple-tests` (multiple executables) for both CUDA 8 and CUDA 9.

The script i used:
```bash
echo 'Configure + Compile'
time (cmake \
-DRAJA_ENABLE_CUDA=On \
-DCUDA_TOOLKIT_ROOT_DIR=$CUDA_ROOT_DIR \
-DCMAKE_PREFIX_PATH=$CHAI_INSTALL_DIR \
-DRAJA_ENABLE_CHAI=On .. >&/dev/null
make -j >&/dev/null)
echo 'Test Suite'
time ctest >&/dev/null"
```

## CUDA 8

| Version | `enh/revert-multiple-tests` | `develop` | % improvement |
|:------:|:-------:|:-------:|:--:|
| Configure + Compile (Real) | 1:24.30 | 0:53.14 | 36.96% |
| Configure + Compile (User) | 4:25.32 | 4:23.88 | 0.54% |
| Execute Test Suite | 0:23.40 | 0:13.18 | 43.67% |

## CUDA 9

| Version | `enh/revert-multiple-tests` | `develop` | % improvement |
|:------:|:-------:|:-------:|:--:|
| Configure + Compile (Real) | 0:45.06 | 0:43.29 | 3.92% |
| Configure + Compile (User) | 3:59.71 | 3:58.60 | 0.46% |
| Execute Test Suite | 0:20.27 | 0:10.09 | 50.22% |

So from these times, we see there is not so much benefit with CUDA 9 compared to CUDA 8, but the tests still run in half the time.

Of course there's more to it than just end-to-end time. Yes, it was a breaking change, but I wasn't aware that there was dependencies on analyzing tests. I should have probed for more input on it. So I'm sorry about that. It won't happen again.

Regardless, I don't mind either form since I'm just going to run `ctest -V` anyways. For me, I just care whether or not the tests work as intended. The complete switch to google-test helps out with that a lot.

willkill07:
I did observe an interesting (read: bad) property. With the additional examples under @artv3's branch, Travis consumed all of the virtual memory:

https://travis-ci.org/LLNL/RAJA/jobs/260414929#L2904

I'm not sure if this is an effect of the additional examples or if this is happening due to concurrent linking. @artv3 got around it by disabling parallel compilation (removing `-j` from the `make` invocation), but I'm not really sure. But if we are indeed consuming all of Travis' resources because of the single binary, this is definitely a tick in box for multiple test files.

willkill07:
Alternatively, we can also create a binary per directory. We already split our tests in the following way:

* `test/integration/<THING>` -- tests for integrating *THING* with RAJA
* `test/unit/*.cpp` -- agnostic/generic tests
* `test/unit/cpu/*.cpp` -- tests for CPU (including sequential and OpenMP)
* `test/unit/cuda/*.cpp` -- tests for CUDA
* `test/unit/omp-target/*.cpp` -- tests for OpenMP 4.5 Target Offload

trws:
Consuming all of the memory is a surprise.  Does this generate a compile of an object for each file followed by a link, or does it generate a single compile command?  Something is making me wonder whether it would work to link each as a shared library then have a single binary made up of gtest_main and all of those .so files.

willkill07:
Closes #304

davidbeckingsale:
Pinging @ajkunen and @rhornung67...

PR #300, opened 2017-08-03
Introduce deprecation macros so we can warn users about any changes

willkill07:
**What is the purpose of this pull request?**

Whatever future changes we do make to RAJA, we should let our users know if a feature will be deprecated or removed. This PR introduces a set of macros that allow us to add deprecation tags function definitions, type definitions, type aliases (using), typedefs, and member variables.

**Changlog**

| Date | Time | Description |
|:----:|:----:|:----|
| 8/2/2017 | 21:16 | revised implementation base on @trws feedback |
| 8/3/2017 | 09:17 | added test results for XL and PGI compilers |

**Was this code tested?**

This has been tested with:
* GCC 4.7+,
* clang 3.3+
* MSVC (through travis)
* Intel 13+
* NVCC 8+
* PGI 15.7+
* XL 14.1 Beta -- the compiler fully supports [[deprecated]] but warns about using it without any warning flags turned on 🤔 

**What features are introduced?**

* `RAJA_HAS_CXX14` -- defined IFF we detect a C++14 compiler
* `RAJA_HAS_ATTRIBUTE_DEPRECATED` -- defined IFF we detect support for `[[deprecated()]]`
* `RAJA_DEPRECATE()`
  * deprecate a variable: place before the definition
  * deprecate a function: place immediately before return type
  * deprecate a type: for a struct/class place immediately after the `struct`/`class` keyword. For `typedef`, place before the typedef.
* `RAJA_DEPRECATE_ALIAS()` -- **requires C++14 or a compiler that acts like a C++14 compiler** -- deprecates a type alias; place between alias name and equals sign (=)

**Are there any removals?** No

**Are there any end user changes?** No

**Are there any internal changes?** Not yet -- these can be used to indicate what we will deprecate in the future.

## Sample code:

```cpp
template <typename T>
struct Pair {
  T first;
  RAJA_DEPRECATE("Second has been removed") T second;
};

struct RAJA_DEPRECATE("The Cilk execution policy has been removed from RAJA")
cilk_exec {
};

template <typename Exec, typename Index1, typename Index2, typename Body>
RAJA_DEPRECATE("This version of forall is deprecated. switch to using RAJA::make_range(begin, end) instead of listing the index parameters)")
int forall(Exec && p, Index1, Index2, Body &&) {
  return 0;
}

RAJA_DEPRECATE("RAJA::Index_type will be removed in 2019 (JK)")
typedef unsigned long Index_type;


using Real_ptr RAJA_DEPRECATE_ALIAS("Real_ptr will be removed in 2018 (JK)") = double*;

int bar() {
  return 0;
}

using exec_pol = cilk_exec; // warning is emitted here

int main() {
  Index_type ind = 4; // warning emitted here (Index_type)
  Real_ptr ptr = nullptr; // warning emitted here (Real_ptr) (C++14 only)
  forall(exec_pol{}, 0, 100, [] { }); // warnings emitted here (exec_pol, forall specialization)
  Pair<int> pair;
  pair.first = 4;
  pair.second = 4; // warning here (second member)
  return pair.first + ind + (ptr - ptr);
}
```


include/RAJA/util/defines.hpp
```diff
@@ -155,4 +155,46 @@ inline void RAJA_ABORT_OR_THROW(const char *str)
...
+//! deprecate a member of a class or struct. Place immediately before the declaration
+#define RAJA_DEPRECATE_MEMBER(Msg) __attribute__((deprecated(Msg)))
+
+//! deprecate a type. For a struct, place after 'struct'. For a typedef, place before 'typedef'
+#define RAJA_DEPRECATE_TYPE(Msg) __attribute__((deprecated(Msg)))
+
+//! [C++14 only] deprecate a using alias. Place between the alias name and equals sign (=)
+#define RAJA_DEPRECATE_ALIAS(Msg)
+
+#if __cplusplus >= 201402L
```
trws: Maybe add an `|| (__cpp_attributes && __has_cpp_attribute(deprecated)`?

willkill07: 👍

willkill07:
@trws hmmm as I was building it up, I thought I needed different specifiers, as something would either exist or not exist. Given that it's reduced down to two groups (alias vs no alias), I can simplify the macros to `RAJA_DEPRECATE("Message")` and `RAJA_DEPRECATE_ALIAS("Message")`

DavidPoliakoff:
Looks good to me!

willkill07:
@rhornung67 that's a really good idea.

I know GCC supports `__attribute__((warning("message")))` but other compilers don't seem to have an equivalent.

`#pragma warning` will always emit when parsed by the preprocessor, which is not what we'd want.

davidbeckingsale:
Yeah, I've used the `warning` pragma before and it's incredibly noisy ☹️

trws:
I'd probably say we have those in an explicitly experimental named folder and/or namespace, with the idea that such things may go through frequent breaking changes without a deprecation period. This seems more for things that users may be relying on that we want to change over while providing the old interface during a migration period.


---
Sent from VMware Boxer

On August 3, 2017 at 1:36:37 PM PDT, David Beckingsale <notifications@github.com> wrote:

Yeah, I've used the warning pragma before and it's incredibly noisy ☹️

willkill07:
I added information about deprecation under `docs/style_guide.md`

davidbeckingsale:
Yeah, do it!

docs/style_guide.md
```diff
@@ -97,3 +97,85 @@ Other usages can be found under:
 * `include/RAJA/util/concepts.hpp`
 * `include/RAJA/policy/PolicyBase.hpp`
 * `include/RAJA/util/Operators.hpp`
+
+# Deprecating older features in RAJA
+
+As the RAJA codebase evolves, there may come a point where features onced used have been replaced with more viable options. To aid users in transitioning from an older API call to a prefered API, we introduce deprecation macros which should *cautiously* and *effectively* be used by RAJA developers.
```
rhornung67: onced -> once

PR #302, opened 2017-08-03
tweak the timer tests to be less noisy and faster

trws:
Save us all ~2 seconds per build, check both lower and upper bounds for time.

### DIFF
--- a/test/unit/test-timer.cpp
+++ b/test/unit/test-timer.cpp
@@ -49,6 +49,7 @@
 #include "RAJA/util/Timer.hpp"
 
 #include <iostream>
+#include <sstream>
 #include <string>
 
 #include <chrono>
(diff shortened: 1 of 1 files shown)


test/unit/test-timer.cpp
```diff
@@ -84,15 +88,14 @@ TEST(TimerTest, No2)
...
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
   }
 
   timer.stop();
 
   RAJA::Timer::ElapsedType elapsed = timer.elapsed();
 
-  EXPECT_TRUE(elapsed > 2.0);
+  EXPECT_GT(elapsed, 0.02);
+  EXPECT_LT(elapsed, 0.05);
```
willkill07: Is 0.05 a safe upper limit? Do we even need an upper limit check?

trws: It seemed high enough to be clearly safe, while low enough to be sure the timer isn't returning random values (I'm looking at you BlueGene/Q 👈 ).  I have no issue with making it higher, but do rather want to have an upper bound just as a sanity check.

davidbeckingsale: 👍

PR #304, opened 2017-08-03
[RFC] Introduce better CMake macros for gtest binary creation

willkill07:
yin : yang as this : #299  (up for debate on which is yin and yang)

The first PR unifying the test binary had some wonky logic required (e.g. `propagate_sources`). This PR introduces a few CMake macros to make defining our tests much easier.

**Additions**

* `raja_gtest_add_subdirectory` -- adds a subdirectory and updates the parent directory with all added sources. NOTE: if you want to create a test binary in the current directory, you only need to state `add_subdirectory`
* `raja_gtest_add_sources` -- adds any number of source files to the current working list of test source files.
*  `raja_gtest_add_binary` -- create a google-test binary. arguments: `NAME <name>` name of executable to generate; `INCLUDES <includes...>` zero or more include paths (ultimately calls `target_include_directories(<name> PUBLIC <includes...>)`)

These macros are added to cmake/RAJAMacros.cmake

This PR just refines the existing implementation we have, but also allows us to cleanly build up any number of google-test binaries without extra cruft. Maybe it makes more sense for an option to be specified to indicate parent propagation or not. Ideally, something flexible would be nice.

**If you want to create a test that only depends on source files found in one directory, `raja_add_test()` is a better option**

Here's an example of constructing binaries from multiple subdirectories.
```
test/CMakeLists.txt
      raja_gtest_add_sources(...) # add files in current dir
      # visit child directory. some source files may be propagated to here
      add_subdirectory(unit) 
      raja_gtest_add_binary(host-tests)

  unit/CMakeLists.txt
        # create a gtest binary only for GPU tests
        add_subdirectory(cuda) # normal add_subdirectory works here, since we don't need to propagate
        add_subdirectory(omp-target)
        raja_gtest_add_binary(gpu-tests)
        # propagates all source files from cpu directory to parent
        raja_gtest_add_subdirectory(cpu)

    cpu/CMakeLists.txt
          raja_gtest_add_sources(...)

    cuda/CMakeLists.txt
          raja_gtest_add_sources(...)

    omp-target/CMakeLists.txt
          raja_gtest_add_sources(...)
```

### DIFF
--- a/cmake/RAJAMacros.cmake
+++ b/cmake/RAJAMacros.cmake
@@ -134,3 +134,46 @@ macro(raja_add_test)
   add_test(NAME ${arg_NAME}
     COMMAND ${TEST_DRIVER} $<TARGET_FILE:${arg_NAME}>)
 endmacro(raja_add_test)
+
+
+set(__internal_test_dir ${CMAKE_SOURCE_DIR}/test CACHE INTERNAL "")
+
+# use this macro to add a directory for tests. This will internall update @raja_gtest_SOURCES
+macro (raja_gtest_add_subdirectory)
+  add_subdirectory(${ARGN})
+  set (raja_gtest_SOURCES ${raja_gtest_SOURCES} PARENT_SCOPE)
+endmacro(raja_gtest_add_subdirectory)
+
+# use this macro to add sources for testing. It will automatically add the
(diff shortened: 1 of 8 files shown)


davidbeckingsale:
Closing, choosing #299.

PR #307, opened 2017-08-08
Extra Copy in ForallN_BindFirstArg_HostDevice Constructor

MrBurmark:
The constructor of ForallN_BindFirstArg_HostDevice took body by value which caused an extra copy to be made.
This fixes the ReduceSumCUDA.indexset_aligned test. The current cuda reducers assume that the first copy in device code will be destroyed after running the lambda. The cuda reducers in #283 assume that their constructors and destructors are called in a nested fashion in device code.


ajkunen:
I agree with Tom..  and this will become less of an issue with the forallN rewrite

PR #310, opened 2017-08-09
RAJA_SIMD configured to force inline lambdas in forall loop with simd…

Ram3sh:
…_exec policy when compiled with the Intel compiler.

Sw4lite triple nested kernel shows 2.5X speedup on Haswells and KNLS with this tweak.

### DIFF
--- a/include/RAJA/config.hpp.in
+++ b/include/RAJA/config.hpp.in
@@ -204,7 +204,10 @@ const int COHERENCE_BLOCK_SIZE = @RAJA_COHERENCE_BLOCK_SIZE@;
 
 #endif
 
-#define RAJA_SIMD  // TODO: Define this...
+#define RAJA_SIMD \
+  _Pragma("ivdep") \
+  _Pragma("simd") \
+  _Pragma("forceinline recursive")
 
 
 #elif defined(RAJA_COMPILER_GNU)


jeffhammond:
Wow.  That's a big jump.  I wonder if ICC has a knob to do this better automatically without the pragma.  I will inquire internally.

davidbeckingsale:
Thanks @jeffhammond !

jeffhammond:
Can I reproduce this with https://github.com/geodynamics/sw4/tree/master/sw4lite or do I need a different version for KNL?

davidbeckingsale:
You best check with @Ram3sh - I'm not sure that the RAJA version has been pushed there yet.

jeffhammond:
@Ram3sh If this isn't in your fork on GitHub, but you are willing to share privately, please let me know.  My email should be easy to find.  Thanks!

DavidPoliakoff:
Forgot to tag, a hydro code that @ajkunen might care a little bit about

Ram3sh:
Testing failing at RAJA/test/unit/cpu/test-nested.cpp:157

The problem seems to be the vectorization of this reduction:

  forallN<Policy>(RangeSegment(1, size_i),
                    RangeSegment(0, size_j),
                    [=](Index_type i, Index_type j) {
                      view(0, j) += view(i, j);
                    });

jeffhammond:
@Ram3sh If you put SIMD (OpenMP or Intel) on a reduction, you need to add the reduction clause.  Otherwise, the compiler is allowed to break your code 😉 

* https://software.intel.com/en-us/node/524555
* https://developers.redhat.com/blog/2014/09/30/openmp-4-0-support-in-developer-toolset-3-beta-parallel-programming-extensions-for-todays-architectures/

jeffhammond:
@Ram3sh ...which means I share the blame here, because I did not address this.  So we have two options:
* Remove `RAJA_SIMD` from reductions.
* Add `RAJA_SIMD_REDUCE(op_var)` and use it with reductions.
To get the latter right, we need to map RAJA reduce operators and variables into what OpenMP etc. require for `op_var`, e.g. `max:norm`.
I propose that Option 1 be implemented now and Option 2 be a second PR.

trws:
It probably is, we may be able to do something about that but it's not 100% clear.  Maybe something to run by Xinmin now that I think about it, it may work to make that an atomic in some cases, or have a way to request simd ordered or similar.

trws:
Should be resolved by #514.

PR #311, opened 2017-08-15
Feature/artv3/simd

artv3:
This branch builds from @jeffhammond 's  #295 branch. Here I introduce no-vectorization defines/pragmas to Jeff's list. The no-vectorization hints are added to the seq_exec forall loops. I also introduce a new execution policy "loop" which may vectorize under the right flags.


artv3:
I created this pull request to share observations. It seems that the only way I can get the GNU compiler to NOT vectorize is by removing inlining & using the flag –fno-tree-vectorize. I find this odd since I would expect that –fno-tree-vectorize would be enough.

More precisely I am using
__attribute__((noinline,optimize("no-tree-vectorize")))  to remove vectorization.

DavidPoliakoff:
I'd just like to highlight that it would be great if this got integrated quickly if possible, I mentioned testing this on code from @ajkunen and seeing really massive speedup from it

artv3:
@Ram3sh shared his code recently with me, I will tryout his code and compare his SIMD macros with the SIMD macro defined here.

rhornung67:
@artv3 the aggressive nature of the gnu compiler w.r.t. vectorization (compared to other compilers) is something we have observed over the past several years.  So I wouldn't be too concerned about it.

jeffhammond:
GCC, vectorization and aggressive are not terms I would have predicted to be strongly associated 🙃

rhornung67:
@jeffhammond for example, in the past, I've observed similar assembly generated (with packed loads/stores and vector instructions) whether I've applied the restrict decoration or not. Thoughts on that?

jeffhammond:
@rhornung67 If the compiler can prove that vectorization is safe, it can do it, whether or not the user provides any assistance.  The only danger is if that proof depends on compliance with the strict-aliasing rule, which is frequently violated.

rhornung67:
@jeffhammond I understand. The observations of Jeff K. and me were frequently that gcc would vectorize code (albeit less than optimal) in many cases that icc would not unless given additional hints.  The fact that the restrict keyword, for example, is not standard, and that different compilers have different pragmas, directive, optimization heuristics, etc. makes this all somewhat of a dark art for C/C++ code...In particular, you often do not know what you're going to get until you try it and do some analysis. To pour salt on this wound, vendor optimization quality varies from compiler version to compiler version and frequently not in a monotonically-improving manner.

jeffhammond:
@rhornung67 Yeah, I hear you that Fortran and C99 are a lot better than C++ 😜 

I haven't dug into C++17 attributes yet but those may be a good path to deal with vectorization control without resorting to pragmas or non-standard sleaze like `__restrict__`.

If you find either of the following particularly compelling, we can discuss over email.
* https://stackoverflow.com/questions/37016119/can-c-attributes-be-used-to-replace-openmp-pragmas
* http://stroustrup.com/C++11FAQ.html#attributes

trws:
If they were supported and/or standardized that would be great. We use the C++ standard attributes for alignment control already when they're supported, so vectoeizaion control should fit right in.

The OpenMP example always amuses me since that actually came from the original proposal for C++11 attributes, the author was in the OpenMP ARB at the time, but we haven't made an official version for that (though we discussed something along those lines last week, attaching OpenMP information to types so it can propagate through templates).


---
Sent from VMware Boxer

On August 16, 2017 at 12:43:39 PM PDT, Jeff Hammond <notifications@github.com> wrote:

@rhornung67<https://github.com/rhornung67> Yeah, I hear you that Fortran and C99 are a lot better than C++ 😜

I haven't dug into C++17 attributes yet but those may be a good path to deal with vectorization control without resorting to pragmas or non-standard sleaze like __restrict__.

If you find either of the following particularly compelling, we can discuss over email.

  *   https://stackoverflow.com/questions/37016119/can-c-attributes-be-used-to-replace-openmp-pragmas
  *   http://stroustrup.com/C++11FAQ.html#attributes

DavidPoliakoff:
> though we discussed something along those lines last week, attaching OpenMP information to types so it can propagate through templates

!!! This would be absolutely massive. I remember discussing this at the last performance portability meeting. Is there an ETA on that?

trws:
Unfortunately no.  It would be a large change and a significant 
implementation burden, so it has quite a ways to go.

artv3:
Hi all, I was wondering if we should introduce two variants of the simd policy? omp_simd_exec, and simd_exec? Or if we should leave as is where simd_exec will use omp simd if omp is enabled.

rhornung67:
@artv3 this is definitely something we need to flesh out. Another option is to add a boolean template argument to the appropriate omp policies (e.g., enable_simd) -- with a default value of false. This may require us to revisit our exec policies for OpenMP4.x.  I'm open to any suggestions for what folks think is best here.

trws:
For now, I'd leave it just simd_exec.  If we start seriously investigating composable OpenMP policies, then we can revisit it, but doubling all of the OpenMP policy instantiations isn't something I'm immediately sure we want to do.

include/RAJA/policy/loop/reduce.hpp
```diff
@@ -0,0 +1,522 @@
...
+ *
+ * \brief   Header file containing RAJA reduction templates for
+ *          sequential execution.
+ *
+ *          These methods should work on any platform.
+ *
+ ******************************************************************************
+ */
+
+#ifndef RAJA_reduce_loop_HPP
```
rhornung67: Do we need this file? We should be able to just use the sequential reduction stuff as is, correct?

include/RAJA/policy/sequential/forall.hpp
```diff
@@ -88,16 +88,29 @@ namespace impl
...
+  RAJA_NoSIMD
   for (auto ii = std::begin(iter); ii < end; ++ii) {
     loop_body(*ii);
   }
 }
 
+
 template <typename Iterable, typename Func, typename IndexType>
-RAJA_INLINE typename std::enable_if<std::is_integral<IndexType>::value>::type
+#if defined(RAJA_COMPILER_GNU)
```
rhornung67: This seems confusing. One branch of the conditional relates to simd, the other relates to inline.

include/RAJA/policy/sequential/scan.hpp
```diff
@@ -80,6 +80,7 @@ concepts::enable_if<type_traits::is_sequential_policy<ExecPolicy>>
 inclusive_inplace(const ExecPolicy &, Iter begin, Iter end, BinFn f)
 {
   auto agg = *begin;
+  RAJA_NoSIMD
```
rhornung67: Please add a blank link  before each of the the RAJA_NoSIMD macros in this file to make it easier to visually parse which lines are variable declarations and which is the loop.

include/RAJA/policy/simd/forall.hpp
```diff
@@ -93,10 +93,12 @@ RAJA_INLINE typename std::enable_if<!std::is_base_of<ListSegment,
                                                      Iterable>::value>::type
 forall(const simd_exec &, Iterable &&iter, Func &&loop_body)
 {
-  auto end = std::end(iter);
+  // TODO: if KNL, make sure long is used
+  auto len = iter.size();
+  auto ii = std::begin(iter);
```
rhornung67: Blank line before the RAJA_SIMD macro please.

include/RAJA/policy/loop/scan.hpp
```diff
@@ -0,0 +1,158 @@
+/*!
```
rhornung67: Similar to the comment on the loop/reduce.hpp file. We shouldn't need this file and instead just use the seq scan stuff.

artv3:
Thanks @rhornung67, I ran into some issues when I tried merging with develop. Thus I moved the contributions to a new branch.

davidbeckingsale:
@artv3 if this PR is now obsolete, please close it.

PR #312, opened 2017-08-18
[WIP] Feature/kunen1/atomic

ajkunen:
Added some atomic operations that work in CUDA and OpenMP.. and should work with other things like TBB (but need to check that).

Feedback welcome!  Especially on design and organization

Currently only support for atomic add and subtract, but other operations could be implemented if needed.

There are 3 main features being added:
1) atomic functions RAJA::atomicAdd and RAJA::atomicSub which operate on arbitrary pointer
2) RAJA::AtomicRef class which is similar to std::atomic, but the ctor takes a pointer to the data
3) RJA::AtomicViewWrapper (created with RAJA::make_atomic_view) which is a View that returns AtomicRef objects

Eventually I would remove the atomics that are in the CUDA reducers, and use these instead.


include/RAJA/RAJA.hpp
```diff
@@ -111,6 +111,13 @@
 #include "RAJA/util/PermutedLayout.hpp"
 #include "RAJA/util/View.hpp"
 
+
+//
+// Atomic operations support
+//
+#include "RAJA/util/atomic.hpp"
```
willkill07: I think `util` should be `pattern` since we have other patterns (forall, scan, reduce) for each policy. Placing in `util` doesn't quite fit this existing model

include/RAJA/policy/cuda/atomic.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+
+#if __CUDA_ARCH__ < 600
+#define RAJA_CUDA_DOUBLE_ATOMIC_CAS
+#endif
+
+//
+// Note: I would much rather all of these functions be device-only, but ran
+//       into issues with their call sites... so instead we have this dumb
+//       #ifdef __CUDA_ARCH__ stuff for essentially device-only code. (AJK)
```
willkill07: 👍

include/RAJA/util/View.hpp
```diff
@@ -63,20 +64,24 @@
...
-          typename LayoutT,
-          typename DataPointer = DataType *>
+template <typename DATA_TYPE,
+          typename LAYOUT_TYPE,
+          typename DATA_POINTER = DATA_TYPE *>
 struct View {
-  LayoutT const layout;
+  using DataType = DATA_TYPE;
+  using DataPointer = DATA_POINTER;
+  using LayoutType = LAYOUT_TYPE;
```
willkill07: What does changing the template names to CAP_CASE with `using`s defined immediately at the top of View accomplish?

include/RAJA/util/View.hpp
```diff
@@ -138,6 +143,45 @@ using TypedManagedArrayView = TypedViewBase<DataType,
...
 
+template <typename ViewType,
+          typename AtomicPolicy = RAJA::auto_atomic>
+struct AtomicViewWrapper {
+  using Base = ViewType;
+  using DataPointer = typename Base::DataPointer;
+  using DataType = typename Base::DataType;
+  using AtomicType = RAJA::AtomicRef<DataType, AtomicPolicy>;
+
+  Base base_;
```
willkill07: Should base be a member or have `private` inheritance? Regardless, it should at least be private and not publicly accessible.

include/RAJA/util/View.hpp
```diff
@@ -138,6 +143,45 @@ using TypedManagedArrayView = TypedViewBase<DataType,
...
+  }
+
+  RAJA_INLINE void set_data(DataPointer data_ptr) { base_.set_data(data_ptr); }
+
+  template<typename ... ARGS>
+  RAJA_HOST_DEVICE
+  RAJA_INLINE
+  AtomicType operator()(ARGS &&... args) const
+  {
+    return AtomicType(&base_.operator()(args...));
```
willkill07: Prefer perfect forwarding the variadic args (since they were defined as a universal reference)  `std::forward<ARGS>(args)...`

include/RAJA/util/atomic/auto.hpp
```diff
@@ -0,0 +1,113 @@
...
+ * policy.
+ *
+ * Next, if OpenMP is enabled we always use the omp_atomic, which should
+ * generally work everywhere.
+ *
+ * Finally, we fallback on the builtin_sync_atomic, which is non-standard but
+ * has broad compiler support.
+ */
+#ifdef __CUDA_ARCH__
+  #define RAJA_AUTO_ATOMIC RAJA::cuda_atomic{}
```
willkill07: I'm not a fan of this, but I also don't think there's a better alternative.

Although, this would be much better if it was a type alias instead of a preprocessor definition.

ajkunen: Will, not sure if it matters, but I didn't want a type (or alias) defined differently on either the host or device side translation units.  This approach limits it to changes in function definitions.

include/RAJA/util/View.hpp
ajkunen: i agree, but I was following the pattern used previously.  perhaps I'll change them all.

include/RAJA/util/View.hpp
ajkunen: That makes the template arguments visible outside of the class.  That's used in the wrapper to grab the DataType.

include/RAJA/RAJA.hpp
ajkunen: Agreed!  I'll move it

ajkunen:
@willkill07 thanks for your feedback, I'll try and work some of these issues out

include/RAJA/policy/cuda/atomic.hpp
trws: What happened at the call sites that was a problem?  I've been playing with this stuff, and it seems that using overload resolution to select between device-only and host-only functions works ok so long as we have a type or at least compile time constant to check.

include/RAJA/policy/cuda/atomic.hpp
```diff
@@ -0,0 +1,152 @@
...
+ */
+RAJA_SUPPRESS_HD_WARN
+template<typename T>
+RAJA_INLINE
+__host__ __device__
+T atomicAdd(cuda_atomic, T *acc, T value){
+#ifdef __CUDA_ARCH__
+  return ::atomicAdd(acc, value);
+#else
+  throw std::logic_error("Cannot call cuda_atomic operations on host");
```
trws: Unfortunately I don't think it can.  It would fail to compile on the host side if it were a static-assert.

include/RAJA/policy/cuda/atomic.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+// Before Pascal's, no native support for double-precision atomic add
+// So we use the CAS approach
+#ifdef RAJA_CUDA_DOUBLE_ATOMIC_CAS
+
+//RAJA_SUPPRESS_HD_WARN
+template<>
+RAJA_INLINE
+__host__ __device__
+double atomicAdd<double>(cuda_atomic, double *acc, double value){
```
trws: This looks good, but is there a reason not to factor this out into a generic "retry atomic" or similar that takes an Operator functor argument so you only have to have the loop once, or at least once per size-class? I realize there's little to be done about double-wide atomics, but it could probably save some maintenance/duplication.

include/RAJA/policy/openmp/atomic.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+struct omp_atomic{};
+
+
+RAJA_SUPPRESS_HD_WARN
+template<typename T>
+RAJA_INLINE
+T atomicAdd(omp_atomic, T *acc, T value){
+#pragma omp atomic
+  *acc += value;
```
trws: The semantics of this are probably not quite what you intend them to be. Currently this atomically updates the value then returns whatever value it has at an unspecified later time.  You probably want `pragma omp atomic capture` followed by something of the form `{ret = *acc; *acc += value}`, which when returning ret would return the value of *acc from before the update took effect.  I'd note here that omp atomics have confusing semantics, but we appear to be stuck with them. ><

include/RAJA/util/View.hpp
trws: I'd go with member personally, in fact it's on my list to rework camp::tuple to work that way, it helps keep certain things easier to read.

include/RAJA/util/atomic.hpp
```diff
@@ -0,0 +1,210 @@
...
+ * @TODO DOCUMENT THIS
+ */
+template<typename T, typename Policy = RAJA::auto_atomic>
+class AtomicRef {
+  public:
+    RAJA_INLINE
+    RAJA_HOST_DEVICE
+    constexpr
+    explicit
+    AtomicRef(T *value_ptr) : m_value_ptr(value_ptr){};
```
trws: Maybe I'm being picky, but no by-reference constructor for a type named *Ref?

include/RAJA/util/atomic/auto.hpp
trws: I'd make one request of these, undef them at the end of this file.  They're a perfectly reasonable way to implement this, but probably not something we want anything external relying on.

test/unit/cuda/test-atomic.cpp
```diff
@@ -0,0 +1,257 @@
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
```
trws: Why separate CPU and cuda tests? Especially for auto_atomic it seems worthwhile to have the same typed test harness, or similar, exercise both sides.

include/RAJA/policy/cuda/atomic.hpp
ajkunen: I was getting an error for calling a device function from a host-device function

test/unit/cuda/test-atomic.cpp
ajkunen: I'm fine with that. I could arrange them all in one unit test file and get more reuse.

include/RAJA/policy/cuda/atomic.hpp
trws: Right, because of host device lambdas. Dangit nvidia. ><

include/RAJA/policy/cuda/atomic.hpp
willkill07: `__CUDA_ARCH__` should only be defined when compiling device code, which is why I suggested it. Valid point.

trws:
Actually, that's a fair point. If the one with the cuda tag is ever called from a host device function that's an error.


---
Sent from VMware Boxer

On August 19, 2017 at 3:48:59 PM PDT, Will Killian <notifications@github.com> wrote:

In include/RAJA/policy/cuda/atomic.hpp<https://github.com/LLNL/RAJA/pull/312#discussion_r134102392>:

> +
+/*!
+ * Catch-all policy passes off to CUDA's builtin atomics.
+ *
+ * This catch-all will only work for types supported by the compiler.
+ * Specialization below can adapt for some unsupported types.
+ */
+RAJA_SUPPRESS_HD_WARN
+template<typename T>
+RAJA_INLINE
+__host__ __device__
+T atomicAdd(cuda_atomic, T *acc, T value){
+#ifdef __CUDA_ARCH__
+  return ::atomicAdd(acc, value);
+#else
+  throw std::logic_error("Cannot call cuda_atomic operations on host");


__CUDA_ARCH__ should only be defined when compiling device code, which is why I suggested it. Valid point.

include/RAJA/policy/cuda/atomic.hpp
trws: Actually, as I mentioned on the other one, why would it ever be correct to call this from a host-device function?  I ran a little test, and the compiler behaves well using overload resolution to solve this kind of thing, for example:

```cuda
struct htag {};
struct dtag {};

void overload(htag) {}
__device__ void overload(dtag) {}

#ifdef __CUDA_ARCH__
using tag = dtag;
#else
using tag = htag;
#endif

__host__ __device__ void hd_test() {
  overload(tag{});
}
```

I know we need to do a great deal more to clean RAJA up in this sense, but it seems better to get a compile error if one is warranted.

willkill07: I sense an astronomical PR in the near future...

test/unit/test-atomic.cpp
```diff
@@ -70,7 +70,7 @@ void testAtomicFunctionBasic(){
   dest[1] = (T)N;
 
   RAJA::forall<ExecPolicy>(seg,
-    [=] RAJA_HOST_DEVICE (RAJA::Index_type i){
+    [=] RAJA_HOST_DEVICE (RAJA::Index_type){
```
rhornung67: Alternatively, there is a RAJA macro in defines.hpp that can be used to removed the named argument to prevent compiler warnings.  It makes it easier to find such things in the code if needed.

include/RAJA/pattern/atomic.hpp
```diff
@@ -0,0 +1,210 @@
...
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#include "RAJA/config.hpp"
+#include "RAJA/util/defines.hpp"
+#include "RAJA/policy/atomic.hpp"
+
+namespace RAJA
+{
+
+
```
rhornung67: Eventually, we need to document usage elsewhere. But, for now, it would be good to add a few simple usage examples -- maybe an atomic in a CUDA kernel, OpenMP threaded loop, etc.

include/RAJA/policy/cuda/atomic.hpp
rhornung67: We should start compiling a list of things like this (or track them as issues) so we don't forget them and can address them in a systematic way.

include/RAJA/policy/cuda/policy.hpp
```diff
@@ -220,19 +220,19 @@ const int RAJA_CUDA_MAX_BLOCK_SIZE = 2048;
 template <typename T>
 __device__ inline T _atomicMin(T *address, T value)
 {
-  return atomicMin(address, value);
+  return ::atomicMin(address, value);
```
rhornung67: I like this use of the scope operator to make it clear when something is in the global namespace 👍

include/RAJA/pattern/atomic.hpp
```diff
@@ -0,0 +1,313 @@
...
+    RAJA_INLINE
+    RAJA_HOST_DEVICE
+    T max(T rhs)const{
+      return RAJA::atomicMax<Policy>(m_value_ptr, rhs);
+    }
+
+    RAJA_INLINE
+    RAJA_HOST_DEVICE
+    T operator&=(T rhs)const{
```
willkill07: This should only be permitted if T is an integral type

rhornung67:
@willkill07 @ajkunen @trws I agree with the namespace suggestion. Looking through the code, it seems that we need to take a broader look at how we are using namespaces and structure them more cleanly. We should probably discuss as a small group, you 3 + @davidbeckingsale + me, and do the namespace thing as a separate PR.

trws:
A namespace is a nice idea. As to static asserts, it should probably only integral or pointer types for those if we want to check it.

Sent from my iPhone

On Aug 24, 2017, at 5:39 AM, Rich Hornung <notifications@github.com<mailto:notifications@github.com>> wrote:


@willkill07<https://github.com/willkill07> @ajkunen<https://github.com/ajkunen> @trws<https://github.com/trws> I agree with the namespace suggestion. Looking through the code, it seems that we need to take a broader look at how we are using namespaces and structure them more cleanly. We should probably discuss as a small group, you 3 + @davidbeckingsale<https://github.com/davidbeckingsale> + me, and do the namespace thing as a separate PR.

ajkunen:
@trws, @willkill07 and @MrBurmark  what kind of memory order constraints should be applied?

Currently the CUDA and MSVC operations have fairly strict ordering requirements.  What should the OpenMP and __atomic_XXX ordering be?

trws:
The c++ and c standards both default to seq_cst. It is the safest by far, but if that's too strong you could go as weak as acquire_release as long as you also provide a flush. Any weaker than that needs to allow the user to pick. Many of our add ops only need relaxed, but others could never be made correct at that level.

The OpenMP atomics default to roughly acquire_release unless you add the seq_cst clause or the compiler can prove that something weaker is safe. We're moving to an acquire release memory model for 5, but for now that's where we are.


---
Sent from VMware Boxer

On August 30, 2017 at 11:39:49 AM PDT, Adam Kunen <notifications@github.com> wrote:

@trws<https://github.com/trws>, @willkill07<https://github.com/willkill07> and @MrBurmark<https://github.com/mrburmark> what kind of memory order constraints should be applied?

Currently the CUDA and MSVC operations have fairly strict ordering requirements. What should the OpenMP and __atomic_XXX ordering be?

ajkunen:
@trws I think that sequentially consistent is probably not required in situations that people are using RAJA.  It also incurs a large performance hit.

Since OpenMP defaults to an acquire/release behavior by defualt... and it seems that's what CUDA provides (from their docs: " Atomic functions do not act as memory fences and do not imply synchronization or ordering constraints for memory operations")... then I think I'll make the atomics in RAJA follow the acquire/release behavior as well.

The problem here is portability:  We don't want different memory ordering based on what execution policy you chose... it should always do the same thing.  Clearly the C++ std::atomic stuff allows quite a lot of flexibility for memory ordering, but things like CUDA do not.  So the acquire/release behavior seems the most consistent.  

**Unless there is any objection, I'm going to make everything use acquire/release.**

trws:
Cuda is slightly weaker, for the record, in that it doesn't inhibit instruction reordering, which acq_rel does.  It's almost, sort-of, relaxed and really should be combined with a thread_fence() or end of a kernel (which does thread_fence()) before reading the value with anything non-atomic.

That said, I agree.  It's probably the most reasonably predictable thing without killing performance.

ajkunen:
That's true about CUDA, and I guess I should have read what I quoted :)

It does seem a little dangerous to leave CUDA as less restrictive.  I could see someone getting code to work with OpenMP and then seeing non-consistent behavior on a GPU.  This kind of inconsistency is really hard to track down in practice.  So I'd rather make CUDA slightly slower, but more consistent.

@trws  do you know how expensive a __threadfence() on every CUDA atomic would be?  it clearly introduces a lot of stalls, but I'm not sure what the practical performance hit is (esp when combined with atomics)

trws:
Unfortunately, a global __threadfence() is extremely expensive.  The simplest thing to at least prevent the compiler from reordering operations is to use __syncthreads(), which has that effect but is at least block local.  If you're willing to go a little... closer to the metal, shall we say, there are some PTX-level ops you can put in an inline assembly block, and marking the arguments as volatile should have most of this effect anyway because of the sequencing requirements of c/c++.

ajkunen:
ok, I've changed the builtin_atomic policies to do acquire/release.
I didn't modify the CUDA behavior.  I don't believe I've seen a use case where this is going to bite someone... so I'm going to leave it be.

I believe that I've addressed everyone's concerns.  @rhornung67, @trws, @MrBurmark, does this pass the muster at this point?  Otherwise, I'm willing to call it done-enough-to-merge.

trws:
Functionality looks good.  I think the only thing left is we talked about having it in an atomic namespace at one point?

include/RAJA/util/View.hpp
trws: Just as a style note, most places we use value_type, pointer_type, etc. to match up with std:: for this kind of thing (or are trying to move in that direction).  I wouldn't let that hold this up, but something to think about.

ajkunen:
@trws, I've moved all of the atomic stuff to RAJA::atomic:: and I've fixed the style issue you pointed out with View

PR #314, opened 2017-08-24
Feature/trws/reduction refactor

### LINKED ISSUE #243: Reduction implementations should reuse code as much as possible
Looking through the current issues related to reductions, many can be solved with a generic view on
* the policy in use
* the type of data which is being reduced
* the reduction operator

Currently, we only really template on the type of data and have hard-wired implementations for each policy and reduction operator.

I have a proof of concept which abstracts away:
* the data type (even for min_loc and max_loc) -- with the possibility of it extending to arrays (if we adopt a model such as `span<>`)
* the reduction operator
* the policy in use -- yes, this means that theoretically, you just need to specialize a constructor and destructor to match the policy you are using.

Please note that this proof-of-concept does not adopt (in full) the `PolicyBase` PR (as it is not finalized)

* Issues that are resolved with this concept: #90 #111
* Issues that can be resolved with such a concept: #30 

GitHub gist is here: https://gist.github.com/willkill07/440fbd97f4fa12346598128463b4b8bd

trws:
Short summary, all of the host reducers use the same base now.  It could be factored out farther than this, but each reducer type is now just an implementation of something similar to TBB combinable.  As a bonus, the OpenMP ordered reduction is also 10x faster than the old one.

@rhornung67, I tried to pull out everything that is no longer necessary, but I'm not sure I got the cmake variable you mentioned as I didn't find one that was obviously related.  If it's still in here I'm happy to pull it out if you'll point me in the right direction.

Fixes #243


include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,291 @@
...
+  constexpr ValueLoc(T const &val) : val{val}, loc{-1} {}
+  constexpr ValueLoc(T const &val, Index_type const &loc) : val{val}, loc{loc}
+  {
+  }
+  operator T() const { return val; }
+  bool operator<(ValueLoc const &rhs) const { return val < rhs.val; }
+  bool operator>(ValueLoc const &rhs) const { return val > rhs.val; }
+};
+}  // end namespace detail
+namespace operators
```
rhornung67: Blank line before, please.

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,291 @@
...
+
+  //! compiler-generated move constructor
+  BaseReduceMin(BaseReduceMin &&) = default;
+
+  //! compiler-generated move assignment
+  BaseReduceMin &operator=(BaseReduceMin &&) = default;
+
+  explicit BaseReduceMin(T init_val,
+                         T initializer = operators::limits<T>::max())
+      : Base(init_val, initializer)
```
rhornung67: Will clang-format fix this vertical alignment weirdness?

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,291 @@
...
+};
+
+/*!
+ **************************************************************************
+ *
+ * \brief  MinLoc reducer class template for use in tbb execution.
+ *
+ **************************************************************************
+ */
+template <typename TT, template <typename, typename> class Combiner>
```
rhornung67: So, unnamed template parameters are OK if you don't need to refer to them in this local declaration? I did not know that....

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,291 @@
...
+
+  //! compiler-generated move assignment
+  BaseReduceMinLoc &operator=(BaseReduceMinLoc &&) = default;
+
+  //! constructor requires a default value for the reducer
+  explicit BaseReduceMinLoc(T init_val, Index_type init_idx)
+      : Base(typename Base::value_type(init_val, init_idx))
+  {
+  }
+  //! reducer function; updates the current instance's state
```
rhornung67: Blank line before, please.

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,291 @@
...
+ **************************************************************************
+ */
+template <typename T, template <typename, typename> class Combiner>
+class BaseReduceMax : public Combiner<T, RAJA::reduce::max<T>>
+{
+public:
+  using Operator = RAJA::reduce::max<T>;
+  using Base = Combiner<T, Operator>;
+  using Base::Base;
+  //! reducer function; updates the current instance's state
```
rhornung67: blank line before

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,291 @@
...
+ **************************************************************************
+ */
+template <typename T, template <typename, typename> class Combiner>
+class BaseReduceSum : public Combiner<T, RAJA::reduce::sum<T>>
+{
+public:
+  using Operator = RAJA::reduce::sum<T>;
+  using Base = Combiner<T, Operator>;
+  using Base::Base;
+  //! reducer function; updates the current instance's state
```
rhornung67: blank line b4

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,291 @@
...
+ */
+template <typename TT, template <typename, typename> class Combiner>
+class BaseReduceMaxLoc : public Combiner<ValueLoc<TT, false>,
+                                         RAJA::reduce::max<ValueLoc<TT, false>>>
+{
+public:
+  using T = ValueLoc<TT, false>;
+  using Operator = RAJA::reduce::max<T>;
+  using Base = Combiner<T, Operator>;
+  //! prohibit compiler-generated default ctor
```
rhornung67: blank line b4

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,291 @@
...
+
+  //! compiler-generated move assignment
+  BaseReduceMaxLoc &operator=(BaseReduceMaxLoc &&) = default;
+
+  //! constructor requires a default value for the reducer
+  explicit BaseReduceMaxLoc(T init_val, Index_type init_idx)
+      : Base(typename Base::value_type(init_val, init_idx))
+  {
+  }
+  //! reducer function; updates the current instance's state
```
rhornung67: blank line b4

include/RAJA/policy/openmp/reduce.hpp
```diff
@@ -15,6 +15,10 @@
 #define RAJA_omp_reduce_HPP
 
 #include "RAJA/config.hpp"
+#include "RAJA/pattern/detail/reduce.hpp"
```
rhornung67: In light of our conversation yesterday about header files, would it be a good idea to pull all header inclusions inside the RAJA_ENABLE_OPENMP guard here? And similarly throughout the code?

include/RAJA/policy/openmp/reduce.hpp
```diff
@@ -69,497 +73,160 @@
...
   {
   }
-
-  //! create a copy of the reducer
-  /*!
-   * keep parent the same if non-null or set to current
-   */
-  RAJA_HOST_DEVICE ReduceMin(const ReduceMin &other)
-      : m_parent(other.m_parent ? other.m_parent : &other), m_val(other.m_val)
+  constexpr ReduceOMP(ReduceOMP const &other)
```
rhornung67: blank line b4

include/RAJA/policy/openmp/reduce.hpp
```diff
@@ -69,497 +73,160 @@
...
+      : parent{other.parent ? other.parent : &other},
+        identity{other.identity},
+        my_data{identity}
   {
   }
-
-  //! Destructor folds value into parent object.
-  RAJA_HOST_DEVICE ~ReduceMin()
+  ~ReduceOMP()
```
rhornung67: blank line b4

include/RAJA/policy/openmp/reduce.hpp
```diff
@@ -69,497 +73,160 @@
...
 
-  //! return the reduced min value.
-  /*!
-   *  \return the calculated reduced value
-   */
-  RAJA_HOST_DEVICE operator T() {
-    return m_val;
-  }
+  void combine(T const &other) { Reduce{}(my_data, other); }
+  void combine(T const &other) const { Reduce{}(my_data, other); }
```
rhornung67: blank line b4

include/RAJA/policy/openmp/reduce.hpp
```diff
@@ -69,497 +73,160 @@
...
-
-  //! create a copy of the reducer
-  /*!
-   * keep parent the same if non-null or set to current
-   */
-  RAJA_HOST_DEVICE ReduceMinLoc(const ReduceMinLoc &other)
-      : m_parent(other.m_parent ? other.m_parent : &other),
-        m_val(other.m_val),
-        m_idx(other.m_idx)
+  constexpr ReduceOMPOrdered(ReduceOMPOrdered const &other)
```
rhornung67: I'm going to stop with the 'blank line b4' comments to avoid being increasingly annoying....

include/RAJA/policy/openmp/reduce.hpp
```diff
@@ -69,497 +73,160 @@
...
-    if (m_parent) {
-#pragma omp critical
-      {
-        Reduce()(m_parent->m_val, m_parent->m_idx, m_val, m_idx);
-      }
+    if (my_data != identity) {
+      Reduce{}((*data)[omp_get_thread_num()], my_data);
+      my_data = identity;
+    }
+    // TODO: determine if parallelizing this is worth it
```
rhornung67: Better to capture this sort of thing in an issue?

include/RAJA/policy/openmp/reduce.hpp
```diff
@@ -568,499 +235,49 @@ class ReduceMaxLoc<omp_reduce, T>
...
-      COHERENCE_BLOCK_SIZE / sizeof(CPUReductionBlockDataType);
-  static constexpr const int s_idx_offset = COHERENCE_BLOCK_SIZE / sizeof(Index_type);
-
-  int m_myID;
-  CPUReductionBlockDataType *m_blockdata;
-  Index_type *m_idxdata;
-  T m_reduced_val;
-  Index_type m_reduced_idx;
-  bool m_is_copy;
+  using Base = detail::BaseReduceMinLoc<T, detail::ReduceOMPOrdered>;
```
rhornung67: Does clang-format fix this formatting?

include/RAJA/policy/openmp/reduce.hpp
```diff
@@ -568,499 +235,49 @@ class ReduceMaxLoc<omp_reduce, T>
...
-  static const int s_block_offset =
-      COHERENCE_BLOCK_SIZE / sizeof(CPUReductionBlockDataType);
-
-  int m_myID;
-  CPUReductionBlockDataType *m_blockdata;
-  T m_init_val;
-  T m_reduced_val;
-  bool m_is_copy;
-
+  using Base = detail::BaseReduceMaxLoc<T, detail::ReduceOMPOrdered>;
```
rhornung67: Formatting?

include/RAJA/policy/sequential/reduce.hpp
```diff
@@ -60,461 +60,111 @@
...
-/*!
- **************************************************************************
- *
- * \brief  Sum reducer class template for use in sequential execution.
- *
- **************************************************************************
- */
 template <typename T>
 class ReduceSum<seq_reduce, T>
+    : public detail::BaseReduceSum<T, detail::ReduceSeq>
```
rhornung67: Can we actually replace detail::ReduceSeq with detail::Seq?  I just noticed it here, but if so, it would apply in other cases too.

include/RAJA/policy/tbb/reduce.hpp
```diff
@@ -74,11 +75,18 @@ namespace RAJA
...
-  using Reduce = Op;
+  struct reduce_adapter {
+    T operator()(T const &l, T const &r)
+    {
+      T tmp = l;
+      Reduce{}(tmp, r);
+      return tmp;
+    }
+  };
   using value_type = T;
```
rhornung67: For consistency, using directives like this should probably be placed before other type definitions, methods, etc.

include/RAJA/policy/sequential/reduce.hpp
trws: That seems reasonable, but I think we'd be best off making it reduce::detail::Seq if you'd like to do that.  Having something called Seq in the RAJA::detail namespace seems like asking for a future conflict.

include/RAJA/policy/openmp/reduce.hpp
trws: We could, but I'm probably just going to remove it.  The cost of spinning up a parallel region is almost certainly a loss here since it's one per thread.

include/RAJA/policy/openmp/reduce.hpp
trws: Fix what about it specifically?  I'm not sure I know what you're asking here.

include/RAJA/pattern/detail/reduce.hpp
trws: I think it's configured to produce this.  I've applied clang-format to the whole thing.  What would you prefer?

include/RAJA/policy/openmp/reduce.hpp
rhornung67: I agree.

include/RAJA/policy/openmp/reduce.hpp
rhornung67: You can ignore this comment. I was going through the PR on my laptop yesterday and it looked weird in a narrower window I had up.

include/RAJA/policy/openmp/reduce.hpp
rhornung67: Ignore this comment also. Looks fine when not viewed in a narrow browser window.

include/RAJA/policy/sequential/reduce.hpp
rhornung67: I was hoping there was a way to unify the traversal and reduction policies so there are fewer names for users to remember.  I suggest leaving it for now. Reordering the words and adding the scope operator between them doesn't help much.

trws:
Oh, we could unify seq_reduce and seq_exec no problem.  The 
detail::ReduceSeq is something the user should never see unless they're 
writing an entirely new reduction type.

The only question there would be what to do about omp_reduce vs 
omp_reduce_ordered.

include/RAJA/pattern/detail/reduce.hpp
trws: They are, but this is also a special case where I'm not sure we could give them names if we wanted to.  This template is taking an un-instantiated template as an argument, so those two inner typenames have no values at this point.

include/RAJA/policy/openmp/reduce.hpp
trws: Fair point, doing it now.

trws:
Thanks @rhornung67.  I think this is basically good to go, but there are a few things to be done to make this mesh well with #283.  I don't think they actually conflict, but there will be some work to do to make everything homogeneous between them.

By the way, be aware that this removes the test-cuda-reduction-maxloc.exe in favor of folding the maxloc tests into the minloc file and calling both test-cuda-reduction-loc.exe.

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,272 @@
...
+  RAJA_HOST_DEVICE constexpr ValueLoc(T const &val, Index_type const &loc)
+      : val{val}, loc{loc}
+  {
+  }
+
+  RAJA_HOST_DEVICE operator T() const { return val; }
+  RAJA_HOST_DEVICE Index_type getLoc() { return loc; }
+  RAJA_HOST_DEVICE bool operator<(ValueLoc const &rhs) const
+  {
+    return val < rhs.val;
```
willkill07: 👍

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,272 @@
...
+
+  /// \brief reducer function; updates the current instance's state
+  const BaseReduceMinLoc &minloc(T rhs, Index_type loc) const
+  {
+    this->combine(value_type(rhs, loc));
+    return *this;
+  }
+
+  //! Get the calculated reduced value
+  Index_type getLoc() { return Base::get().getLoc(); }
```
willkill07: shouldn't this be `const` ?

Also, I see a potential to have getLoc() be CRTP inherited -- not a requirement but it does de-duplicate code.

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,272 @@
...
+ */
+template <typename T, template <typename, typename> class Combiner>
+class BaseReduceSum : public BaseReduce<T, RAJA::reduce::sum, Combiner>
+{
+public:
+  using Base = BaseReduce<T, RAJA::reduce::sum, Combiner>;
+  using Base::Base;
+
+  //! reducer function; updates the current instance's state
+  const BaseReduceSum &operator+=(T rhs) const
```
willkill07: reduction operators could also be CRTP inherited -- not sure if that's what we want.

include/RAJA/util/Operators.hpp
```diff
@@ -336,16 +335,16 @@ static_assert(check<unsigned long long>(),
 template <typename Ret, typename Arg1 = Ret, typename Arg2 = Arg1>
 struct plus : public detail::binary_function<Arg1, Arg2, Ret>,
               detail::associative_tag {
-  RAJA_HOST_DEVICE Ret operator()(const Arg1& lhs, const Arg2& rhs)
+  RAJA_HOST_DEVICE Ret operator()(const Arg1& lhs, const Arg2& rhs) const
```
willkill07: well, I feel dumb about making these non-const originally :\

include/RAJA/pattern/detail/reduce.hpp
trws: Yup, definitely should, thanks for catching that. I'll look at CRTP inheriting some of that too.

include/RAJA/util/Operators.hpp
trws: The surprise on there was that they ever worked without it... with lambda capture all over the place, everything becomes const *all the time*.  Not sure what tickled it, but something did. 🤷‍♂️

trws:
Ok, finally stable again. I rebased to squash out some of the commit noise, but assuming this passes it should be good to go.

trws:
Think this is GTG.  Re-review after final changes @rhornung67, @davidbeckingsale?

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,397 @@
...
+#endif
+
+namespace detail
+{
+
+template <typename T, bool doing_min = true>
+class ValueLoc
+{
+public:
+  T val = doing_min ? operators::limits<T>::max() : operators::limits<T>::min();
```
willkill07: Does this ever pop up anywhere else? This might be a good constexpr function to have separated.

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,397 @@
...
+{
+
+template <typename T,
+          template <typename> class Reduce_,
+          template <typename, typename> class Combiner_>
+class BaseReduce
+{
+  using Reduce = Reduce_<T>;
+  // NOTE: the _t here is to appease MSVC
+  using Combiner_t = Combiner_<T, Reduce>;
```
willkill07: 👍  but also 😡

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,397 @@
...
+  constexpr BaseCombinable(BaseCombinable const &other)
+      : parent{other.parent ? other.parent : &other},
+        identity{other.identity},
+        my_data{identity}
+  {
+  }
+
+  ~BaseCombinable()
+  {
+    if (parent && my_data != identity) {
```
willkill07: Justification for the additional check on identity? Is this to disallow extraneous folding with copies?

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -0,0 +1,397 @@
...
+      Reduce()(parent->my_data, my_data);
+    }
+  }
+
+  void combine(T const &other) { Reduce{}(my_data, other); }
```
willkill07: Should/could this be const? `my_data` is mutable already

include/RAJA/policy/openmp/reduce.hpp
```diff
@@ -60,1008 +59,102 @@
...
-  //! Destructor folds value into parent object.
-  RAJA_HOST_DEVICE ~ReduceSum()
-  {
-    if (m_parent) {
-#pragma omp critical
-      {
-        Reduce()(m_parent->m_val, m_val);
-      }
+    if (Base::parent) {
+#pragma omp critical(ompReduceCritical)
```
willkill07: Eventually will we want to use locks? Or is the named critical section enough still improve our current performance?

include/RAJA/pattern/detail/reduce.hpp
trws: Yes, it should, crud, I'll put up a PR for that.

include/RAJA/pattern/detail/reduce.hpp
trws: It's because both destructors will get run when there's another outside, so it's to avoid extra folding yeah.

include/RAJA/policy/openmp/reduce.hpp
trws: I actually tried to use locks, it worked great with everything but GOMP, which would segfault whenever I tried to `new` an omp_lock_t or any structure containing an omp_lock_t.  It would be great to track that down and switch over, but seemed like something for a subsequent PR.

PR #316, opened 2017-08-29
remove compile-time error from memutils, add portable alligned allocator

trws:
Thanks to @tepperly for reporting this one.  After moving the aligned allocation and free stuff out of implementation and into a header it's possible for the `#error` condition to come up in user code.  This PR makes them dependent on configuration time, rather than compile-time, parameters and adds a portable aligned allocation capability such that the error pragma is never needed.

### DIFF
--- a/include/RAJA/internal/MemUtils_CPU.hpp
+++ b/include/RAJA/internal/MemUtils_CPU.hpp
@@ -90,6 +90,7 @@ inline void* allocate_aligned(size_t alignment, size_t size)
   if (nullptr == mem) return nullptr;
   void **ptr = (void **)((std::uintptr_t)(mem + alignment + sizeof(void *))
                          & ~(alignment - 1));
+  // Store the originall address one position behind what we give the user.
   ptr[-1] = mem;
   return ptr;
 #endif
@@ -116,6 +117,8 @@ inline void free_aligned(void* ptr)
 #elif defined(RAJA_PLATFORM_WINDOWS)
   _aligned_free(ptr);
 #else
+  // Free the address stored one position behind the user data in ptr.
+  // This is valid for pointers allocated with allocate_aligned
   free(((void**)ptr)[-1]);
 #endif
 }
--- a/include/RAJA/internal/MemUtils_CPU.hpp
+++ b/include/RAJA/internal/MemUtils_CPU.hpp
@@ -90,7 +90,7 @@ inline void* allocate_aligned(size_t alignment, size_t size)
   if (nullptr == mem) return nullptr;
   void **ptr = (void **)((std::uintptr_t)(mem + alignment + sizeof(void *))
                          & ~(alignment - 1));
-  // Store the originall address one position behind what we give the user.
+  // Store the original address one position behind what we give the user.
   ptr[-1] = mem;
   return ptr;
 #endif
--- a/include/RAJA/internal/MemUtils_CPU.hpp
+++ b/include/RAJA/internal/MemUtils_CPU.hpp
@@ -59,6 +59,8 @@
 #include "RAJA/util/types.hpp"
 
 #include <cstddef>
+#include <cstdlib>
+#include <memory>
 
 #if defined(_WIN32) || defined(WIN32) || defined(__CYGWIN__) \
     || defined(__MINGW32__) || defined(__BORLANDC__)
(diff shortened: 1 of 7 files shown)


include/RAJA/internal/MemUtils_CPU.hpp
```diff
@@ -102,12 +111,12 @@ inline T* allocate_aligned_type(size_t alignment, size_t size)
...
 inline void free_aligned(void* ptr)
 {
-#if defined(HAVE_POSIX_MEMALIGN)
+#if defined(RAJA_HAVE_POSIX_MEMALIGN) || defined(RAJA_HAVE_ALIGNED_ALLOC)
   free(ptr);
 #elif defined(RAJA_PLATFORM_WINDOWS)
   _aligned_free(ptr);
 #else
-#error "No known aligned allocator available"
+  free(((void**)ptr)[-1]);
```
davidbeckingsale: What does this do?

ajkunen: segfaults :)

DavidPoliakoff: `(*)(int*)(0) = 42`

is the canonical way of doing that

davidbeckingsale: If this is supposed to error, we should just throw an exception 😎

DavidPoliakoff: This is a hazy recollection: do exceptions make linking iffier? I have a memory I don't trust that says they do

trws: The `aligned_allocate` stores the address of the base of the allocated region one slot behind the beginning of the data given to the user.  Basically it does this:

```c
// in alloc
void ** mem = malloc(...);
void ** usermem = mem + some_offset;
usermem[-1] = mem;
// in free
free(usermem[-1]);
```

davidbeckingsale: Thanks for the explanation. Can you add that as a comment in the code?

include/RAJA/policy/openmp/reduce.hpp
```diff
@@ -102,7 +102,7 @@ class ReduceOMP
 
 } /* detail */
 
-RAJA_DECLARE_ALL_REDUCERS(omp_reduce, detail::ReduceOMP);
+RAJA_DECLARE_ALL_REDUCERS(omp_reduce, detail::ReduceOMP)
```
davidbeckingsale: Should these be in this PR?

include/RAJA/internal/MemUtils_CPU.hpp
trws: @DavidPoliakoff not so much anymore, unless you're linking with C and you let one escape, then you're in for a bumpy ride...

include/RAJA/policy/openmp/reduce.hpp
trws: We can break them out, but they came from the same warning checking report from @tepperly.  It's only a warning with -Werror=pedantic apparently, but it is a warning nevertheless.

davidbeckingsale: Okay, fine with me since it's a minor change. 

Perhaps we should add `-Werror=pedantic` to our flags for Travis....

include/RAJA/internal/MemUtils_CPU.hpp
DavidPoliakoff: @trws 

Good to know, and thanks for the actual explanation

include/RAJA/policy/openmp/reduce.hpp
trws: Maybe, I'm not sure what all it will pick up, but if we're already clean we may as well stay that way.

davidbeckingsale: Exactly 😄

include/RAJA/internal/MemUtils_CPU.hpp
```diff
@@ -90,6 +90,7 @@ inline void* allocate_aligned(size_t alignment, size_t size)
   if (nullptr == mem) return nullptr;
   void **ptr = (void **)((std::uintptr_t)(mem + alignment + sizeof(void *))
                          & ~(alignment - 1));
+  // Store the originall address one position behind what we give the user.
```
davidbeckingsale: Typo, but thanks for adding these.

trws:
If there's nothing else, we should get this one in soonish, it's holding up VBL automated tests apparently.

davidbeckingsale:
I'm happy to approve this as soon as the typo is done, then we can merge as soon as travis passes.

PR #317, opened 2017-08-29
add Intel _mm_{malloc,free} to aligned heap managers

jeffhammond:
This fills the (small) gap between POSIX (`posix_memalign`) and C++17 (`std::aligned_alloc`), should one exist.


trws:
I'm good with this, just need the icc compiler test on travis to pass... I'll restart it.

trws:
Looks like we're getting:

```
Sorry, but the Trial License Activation does not work in silent mode on this
environment. If you still want to use it, please run installation in usual mode.
```

From the icpc test.  Thoughts @davidbeckingsale, @jeffhammond?

jeffhammond:
🤷‍♂️

davidbeckingsale:
It's something to do with the `secure` part of our Travis config, and I'm guessing it's because @jeffhammond is setting up this PR from a fork, unlike most of our internal PRs that are branches of this repo.

jeffhammond:
Ah yes, that's the reason.  Travis doesn't know that my PR doesn't print the secure env var to the terminal, thereby defeating security.  They trust the authors of branches to not do this.

trws:
Huh... we need a way to deal with that somehow.  I'm not sure what that would be though.  The obvious solution is to not test intel when it's a PR from an external fork, but that leaves us with part of our test coverage lost.  We can always pull PRs like this into local branches, but it's a much higher overhead workflow at that point.

@jeffhammond, have you run into any good ways to deal with this?

jeffhammond:
https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions

trws:
Thanks @jeffhammond!

PR #324, opened 2017-09-12
Feature/artv3/looptest

artv3:
Replaced simd test with loop test.


trws:
Much better, thanks @artv3.

PR #326, opened 2017-09-15
desnarl forall

### LINKED ISSUE #323: restore constexpr for layout constructors
Somewhere along the way we lost the cxx14 constexpr for LayoutBase.  Bring this back.

trws:
After running into some non-trivial issues getting forall to pick up a fake impl in a test earlier, as well as various other issues with the cyclic nature of the existing setup, this reworks forall to use ADL to resolve the implementation in a per target namespace.  Basically, it does this:

```c++
namespace RAJA {
struct pol {};
namespace impl {
forall(pol...)
}
}

// becomes
namespace RAJA {
namespace policy {
namespace pol_tgt {
struct pol {};
forall_impl(pol...)
}
}
using policy::pol_tgt::pol;
}
```

In summary, new forall variants can be added anywhere, in any order, by creating a policy type and a forall_impl function in the same namespace as one-another.  This change should not change any external APIs.

Also this replaces all internal Icount variants with a wrapper struct, which is possible thanks to the static indexset and the wrap layer.

Fixes #323


trws:
Thanks @willkill07, that would be great.

include/RAJA/index/IndexSet.hpp
```diff
@@ -500,7 +497,7 @@ class StaticIndexSet<T0, TREST...> : public StaticIndexSet<TREST...>
   {
     StaticIndexSet<T0, TREST...> *retVal = new StaticIndexSet<T0, TREST...>();
     int numSeg = getNumSegments();
-    for (auto & seg : segIds)
+    for (auto &seg : segIds)
```
rhornung67: Can we add some curly braces to this like this to avoid logic/conceptual errors/misunderstanding?

trws: Sure, I have no issue with that.  Actually I rather wish clang-format had an option to add them for us, but sadly it does not. 😞

rhornung67: Also, it appears that most of the changes in the PR involve formatting so it's kinda hard to tell what's really changed. I think we need to get to a point where no code would change in develop by running it through our formatting tool. Then, we need to require all new code to be properly formatted.  Agree?

trws:
Absolutely.  We need pre-commit hooks for this, and probably a travis 
test that ensures nothing changes when the tool is run to make sure it 
takes though.

include/RAJA/index/IndexSet.hpp
rhornung67: Another thing we may consider is that when we switch over to BLT (which I assume we will do at this point), we would have uncrustify for free.  That would give us more formatting capabilities.

trws:
Do you have an appropriate formatting file for the format we’ve been 
using for it?  If we’re going to make the switch, we may as well just 
do it with the pre-commit hooks.

davidbeckingsale:
This looks good! I would agree with @rhornung67 's comment that the formatting obfuscates the changes a little, but it's not a big deal.

trws:
Sorry this has bounced about a bit, but I'm finally pretty well done with this one.  @davidbeckingsale, one last review?

rhornung67:
@trws thanks. Looks good! 👍

Issue #327, opened 2017-09-16
Intel 18: policy/sequential/forall.hpp(98): warning #161: unrecognized #pragma     RAJA_NO_SIMD


jeffhammond:
I suspect this is a trivial issue, which I will diagnose and presumably fix shortly.

```
Scanning dependencies of target example-raja-pi
[ 25%] Building CXX object examples/CMakeFiles/example-raja-pi.dir/pi.cpp.o
/Users/jrhammon/Work/DOE/LLNL/RAJA/include/RAJA/policy/sequential/forall.hpp(98): warning #161: unrecognized #pragma
    RAJA_NO_SIMD
    ^
          detected during:
            instantiation of "RAJA::concepts::enable_if<RAJA::concepts::negate<RAJA::type_traits::is_indexset_policy<ExecutionPolicy>>, RAJA::type_traits::is_range<Container>> RAJA::wrap::forall(ExecutionPolicy &&, Container &&, LoopBody &&) [with ExecutionPolicy=RAJA::seq_exec, Container=RAJA::TypedRangeSegment<RAJA::Index_type={ptrdiff_t={long}}>, LoopBody=lambda [](int)->void]" at line 510 of "/Users/jrhammon/Work/DOE/LLNL/RAJA/include/RAJA/pattern/forall.hpp"
            instantiation of "RAJA::concepts::enable_if<RAJA::type_traits::is_integral<IndexType1>, RAJA::type_traits::is_integral<IndexType2>> RAJA::forall(ExecutionPolicy &&, IndexType1, IndexType2, LoopBody &&) [with ExecutionPolicy=RAJA::seq_exec, IndexType1=RAJA::Index_type={ptrdiff_t={long}}, IndexType2=RAJA::Index_type={ptrdiff_t={long}}, LoopBody=lambda [](int)->void]" at line 702 of "/Users/jrhammon/Work/DOE/LLNL/RAJA/include/RAJA/pattern/forall.hpp"
            instantiation of "void RAJA::forall<ExecutionPolicy,Args...>(Args &&...) [with ExecutionPolicy=RAJA::seq_exec, Args=<RAJA::Index_type={ptrdiff_t={long}} &, RAJA::Index_type={ptrdiff_t={long}} &, lambda [](int)->void>]" at line 59 of "/Users/jrhammon/Work/DOE/LLNL/RAJA/examples/pi.cpp"

```

artv3:
[...]

PR #328, opened 2017-09-18
adjusted preprocessor order

artv3:
Addressing the bug in #327. Thanks @jeffhammond !


jeffhammond:
LGTM

in theory, moving below PGI protects against something similar there but it's a completely theoretical problem AFAIK

PR #332, opened 2017-09-25
Add RAJA::View::operator[] for slicing and alternate access to operator()

rrsettgast:
This pull request is to add RAJA::View::operator[] for slicing and c-style access for multidimensional arrays. I have files from previous work on an ArrayAccessor<NDIM> class similar in functionality to RAJA::View as a baseline for performance.

The approach I took with the ArrayAccessor<NDIM>::operator[] return an ArrayAccessor<NDIM-1> and specialize for ArrayAccessor<1>, which is a direct access case. We can take the same approach for RAJA::View, but the flexibility for slicing needs to be pushed down to the RAJA::Layout object without creating a new RAJA::Layout object for RAJA::View returned from operator[].


rrsettgast:
Here is a link to an excel sheet that outlines performance of the ArrayAccessor class against 1d arrays, and RAJA::View::operator() for a matrix multiple with some random operations thrown. Based on these results, I would propose that RAJA::View could be further optimized.

[https://drive.google.com/file/d/0BzD41FgRmns3WFBoU2RNTEVsd0U/view?usp=sharing](url)

willkill07:
Maybe I’m missing something here, but the code attached

 (1) doesn’t integrate with RAJA at all
 (2) wasn’t evaluated on the same class of compilers to which we evaluate RAJA (more modern GCC, a clang-based compiler)

Also, the spreadsheet doesn’t make it obvious what each version is. I’m assuming operator() is RAJA::View but everything else isn’t clear without some sorts of README.

From the implementation side, seeing 3+ versions of the essentially the same file and a lot of code duplication (specialization for N, 2, and 1-ary dim when only N-ary + variadic args is required) doesn’t leave me with a postive feeling. It seems like the compiler may just be able to do a better job because of how specialized the implementation is.

Regardless, I do think we can certainly improve upon RAJA::View. I’ve worked on a few View-like abstractions and chaining operator[] has resulted in poor inlining and reuse for index calculations.

davidbeckingsale:
@willkill07 this came up in some discussions with @rrsettgast and I asked him to push the code up here for discussion. Randy has worked on these array access code ideas a lot and it will be important for integrating RAJA into his application. One of the nicest new features is the ability to slice the data.

@rrsettgast I think it would be helpful if you could address @willkill07 's comments on the spreadsheet, and maybe try running your tests with newer compiler versions.

rrsettgast:
@willkill07 @davidbeckingsale 
Sorry about the lack of coherence. I had some old sheets in that excel file that made it rather confusing. I deleted the old sheets, which should make it somewhat clearer. Also I removed various version that I had played around with. The relevant version remains, and most closely resembles RAJA::View. Also I commented out the specialization for NDIM=2, although for the best performance it should be use and return a raw restricted pointer. I reran with the version of the single file that is currently in the repo.

@willkill07 To address the points:
1) That wasn't the point yet. The goal is to add a RAJA::View::operator[] to slice and provide c-style array access. I had previously worked on a class to do that. That is what I included. The ArrayAccessor::operator[] approach seems to be the best one that I have found, although it is not directly implementable in RAJA::View due to the fact that RAJA::View has a Layout object, and the method used in ArrayAccessor::operator[] requires a very lightweight object.

2) It would be nice to use a newer version of gcc, and I can certainly do that. What platform and version would you recommend? I had previously used clang, but it was pretty terrible performance wise. @rhornung67 had asked me to use intel, so I just replaced it. I can add it back in.

I understand the difficulties associated with chaining operator[], but from the results in the attached sheet, it looks better than a RAJA::View::operator() for this specific test case. One thing to address is the fact performance changes depending on whether or not the array is passed into a function, and how it is passed into that function.

Again, sorry if the previous commit was too scattered.

rrsettgast:
I put up the results using gcc7.1, icc17, and clang4 on quartz. The results are a little better, which is kind of a relief, but some performance hits remain for the cases with the small contraction dimension.

https://drive.google.com/file/d/0BzD41FgRmns3WFBoU2RNTEVsd0U/view?usp=sharing

Table 4 is the first thing to look at. It lists the runtimes normalized with minimum runtime for a given combination of compiler and matrix size. 

1) 1D restrict arrays win...most of the time. There appear to be cases where clang fails to maintain the optimization for restrict for small contraction sizes. In that case ArrayAccessor::operator[] solidly beats the other options. Note this is only for clang.

2) For gcc7.1  and intel17 the 1d restrict pointer is a good bet. For GCC, all the accessors slowdown for smaller contraction dimensions, but for ICC, ArrayAcessor::operator[] keeps it together pretty well for the inline case, while all other cases degrade. 

3) for clang, ArrayAccessor::operator[] does really well inline with no degradation. The function calls experience one performance degradation for a specific contraction dimension, but it can be seen that for some reason the compiler just lost it's ability to restrict the data pointer as evidenced by the 1D array. 

One thing to note is that unlike RAJA::View, which has a RAJA::Layout that calculates a stride, the ArrayAccessor has no strides pre-calculated. I tried to mirror the 1D array as much as possible in that the strides are calculated inline rather than pre-computed.

So to the point. Do you guys want to have operator[] in the RAJA::View? I think that it may be actually more performant than operator(). This would require a lightweight view, and a layout that can be used for multiple levels of slice. Sliced Views could have a pointer to the top level View::layout object, and then pass a template argument to layout::operator[] to increment the data pointers to get the right strides for the slices.

One odd thing is that when ArrayAccessor actually pre-calculated the strides like is done in RAJA::Layout, there was a performance hit associated with that. I have no idea why. 

Also, if I defined a non-default copy constructor and a move constructor, there were penalties for that too.

Oh...and this test is for 2d arrays. I don't know what will happen for n-dimension arrays.

rrsettgast:
Hi. OK, I put in a RAJA::View::operator[] and RAJA::Layout_Base_impl::operator[]. The benchmark that I had previously attached runs, although the implantation is not complete. Layout has some template parameter pack peeler classes that need to be specialized for sizeof...(RangeInts) = 1. So it runs, but currently fails the integration tests. RAJA::View was only slightly modified, while most of the works is in the layout. I created a new layout class (LayoutBase_interface) that has pointers to the data members of RAJA::Layout_Base_impl, so it is lightweight. So then RAJA::Layout_Base_impl::operator[] returns a RAJA::LayoutBase_interface with the first range var peeled off. So the original operator() still works at the top level, but probably wont work for slices, and the operator[] works....as long as it is being interfaced though a RAJA::View which provides the specialization for Layouts of dimension 1.

The results are great. Obviously there is more work to be done to clean it up and make Layout self sufficient again.


[operatorCompare_RajaSquare.xlsx](https://github.com/LLNL/RAJA/files/1342388/operatorCompare_RajaSquare.xlsx)

DavidPoliakoff:
Folks,

That spreadsheet is dense, but worth looking at. Of real interest:

In figure 2, the Raja view has worse performance when passed by value, but much better when passed by reference. I don't have time for much more than drive-by "that's interesting, hmmm" comments, but that's interesting. Hmmm.

Thanks for the things to think about, I loved it when you gave the talk on this, reference it regularly.

rrsettgast:
Does anyone have a problem with creating lightweight(i.e. only pointer members) views and layouts for the slicing/access operator[]? I would like some feedback on this design choice prior to completing the implementation. Thanks

rrsettgast:
Slicing using this approach will not work with permuted layouts. Closing PR.

PR #337, opened 2017-10-04
target-omp re-sync

jonesholger:
This PR just re-syncs target-omp.  I would also like to get the example-pi omp-target variant in this PR. 
Also, it would be good to have a target back-end that can compile and run without a "device" target (i.e. host side), so we can put this into Continuous Integration


test/unit/omp-target/test-nested-reduce.cpp
```diff
@@ -27,15 +27,17 @@ static RAJA::Index_type const area = xExtent * yExtent;
 
 TEST(NestedReduceTargetOMP,outer)
 {
+  RAJA::Index_type  l_begin = begin;
```
willkill07: Has this bug been reported to IBM?

include/RAJA/policy/openmp/target_reduce.hpp
```diff
@@ -48,6 +48,59 @@ namespace RAJA
 namespace omp
 {
 
+template <typename T>
+struct sum {
+  static constexpr T identity = T(0);
+  RAJA_HOST_DEVICE RAJA_INLINE
+  void operator()(T &val, const T v) { val += v; }
+};
+
```
willkill07: What was wrong with using `RAJA::reduce::sum` and others? We now have multiple implementations for the same reduction functions. Additionally, the new function objects aren't marked within a `#pragma omp declare target` block.

jonesholger: Thanks for the heads up Will. I was actually thinking about your second point, and will get those marked, though to be fair reduction unit test passed without it.

test/unit/omp-target/test-nested-reduce.cpp
jonesholger: Not yet!

include/RAJA/policy/openmp/target_reduce.hpp
jonesholger: ON your first point there is a nuanced issue with the other RAJA::reduce::sum operator
which errors out with no matching constructor (for 2nd argument, i.e. identity), hence the reason I brought in some older operators.

jonesholger: Ok, took care of the redundant sum,min,max. However minmaxloc didn't change as it's a stretch to utilize the ValueLoc  class.

willkill07: Ah okay. I’ll look into seeing if there’s a way to get it to work.

jonesholger:
@willkill07 - Hey Will can we get a review. We need this for the RAJA Perf Suite. Also, note omp-target reducers are not performant, clocking in at 1/100 of RAJA sequential. This should be looked at in the future, so for now we should just consider this "functional"

include/RAJA/policy/openmp/policy.hpp
```diff
@@ -192,6 +199,7 @@ using policy::omp::omp_reduce_ordered;
 
 #if defined(RAJA_ENABLE_TARGET_OPENMP)
 using policy::omp::omp_target_parallel_for_exec;
+using policy::omp::omp_target_parallel_for_exec_nt;
```
willkill07: I'm assuming `_nt` is num_teams?

include/RAJA/policy/openmp/target_reduce.hpp
```diff
@@ -47,6 +47,39 @@ namespace RAJA
 
 namespace omp
 {
+#ifdef RAJA_ENABLE_TARGET_OPENMP
```
willkill07: I wonder why this is specific for `RAJA::omp` when this is just a common min/max-loc reduction function object. This lived just fine under `RAJA::reducers` before. I'm not sure why it needs moved here

test/unit/omp-target/test-nested-reduce.cpp
willkill07: I still think this is silly but we have to make it work in the meantime :)

include/RAJA/policy/openmp/target_reduce.hpp
jonesholger: Yeah this is the most incongruous piece, where pattern/detail/reduce moved ahead in time wrt to BaseReduceMinLoc and ValueLoc constructs, while target-reduce fell behind. Note, we could generate an new issue to adapt either of these two to work together and gen a separate PR. Those function objects that you notice were pulled from older gen RAJA where it did live in reduce::, and were copied verbatim into RAJA::omp.

include/RAJA/policy/openmp/policy.hpp
jonesholger: Matt experimented here by producing a policy where you "don't specify NUM_TEAMS" explicitly, and he named it thus ( no teams). But, performance didn't necessarily improve. Arguably the compiler/runtime should have a better notion of setting num teams based on grid blocks, or occupancy then just arbitrary numbers that the user selects. Setting thread count explicitly (i.e. blocksize) makes a little more sense.

PR #339, opened 2017-10-09
Feature/ajk/stride1

ajkunen:
Added template argument for RAJA::Layout to identify a stride-one index.
This allows the index calculation to drop the stride term on that index, allowing compilers to optimize better (esp. for SIMD).


rrsettgast:
Is the "stride one" dimension always your last stride entry? For instance if you have a 3 dimensional view with dimensions (2,4,8), then the strides would be (32,8,1)?

willkill07:
@rrsettgast that seems correct from my look at the code.

ajkunen:
not for permuted layouts

rrsettgast:
OK. So the permuted layout rearranges the stride array in an order that corresponds to the chosen permutation from Permutations.hpp for example....which is known at compile time. And the ConditionalMultiply just removes that stride_one multiply operation from the index calculation. So then for the non-permuted case, the default value for StrideOneDim should be "sizeof...(RangeInts)-1" since there always is a stride_one dimension?

ajkunen:
@rrsettgast I just talked to @trws about this very thing.  There is nothing in the design of Layout that enforces that "at least one dimension is stride 1".  I would like to keep it that way.  You could imagine a 1-dimensional layout that has a stride of 4 (for whatever reason).  I would rather not require or default to having anything stride-1.

include/RAJA/util/Layout.hpp
```diff
@@ -318,6 +366,38 @@ struct TypedLayout : public Layout<sizeof...(DimTypes), Index_type> {
 };
 
 
+
+/*!
+ * Convert a non-stride-one Layout to convert to a stride-1 Layout
```
trws: Typo, extra "to convert".

include/RAJA/util/Layout.hpp
```diff
@@ -318,6 +366,38 @@ struct TypedLayout : public Layout<sizeof...(DimTypes), Index_type> {
...
+RAJA_INLINE
+Layout<n_dims, IdxLin, s1_dim>
+make_stride_one(Layout<n_dims, IdxLin> const &l)
+{
+  return Layout <n_dims, IdxLin, s1_dim>(l);
+}
+
+
+/*!
+ * Convert a non-stride-one TypedLayout to convert to a stride-1 TypedLayout
```
trws: Same typo here.

trws:
Hm, that and the layout test on intel.

ajkunen:
@trws I think I've addressed your issues, is there anything else?

trws:
Works for me, thanks @ajkunen!

PR #341, opened 2017-10-10
Fix TBB dependency

davidbeckingsale:


cmake/SetupPackages.cmake
```diff
@@ -44,7 +44,8 @@ if (ENABLE_TBB)
   if(TBB_FOUND)
     blt_register_library(
       NAME tbb
-      INCLUDES ${TBB_INCLUDE_DIRS})
+      INCLUDES ${TBB_INCLUDE_DIRS}
+      LIBRARIES tbb)
```
trws: My only nitpick is that this should probably be ${TBB_LIBRARIES}, or a condition between the release and debug variants thereof if you're feeling ambitious.

cmake/SetupPackages.cmake
```diff
@@ -44,7 +44,8 @@ if (ENABLE_TBB)
   if(TBB_FOUND)
     blt_register_library(
       NAME tbb
-      INCLUDES ${TBB_INCLUDE_DIRS})
+      INCLUDES ${TBB_INCLUDE_DIRS}
+      LIBRARIES ${TBB_LIBRARIES})
```
davidbeckingsale: @trws this uses the variable now. I didn't want to try and be too ambitious 😄

trws:
Github appears to think I am somewhere forward in time: 
image

trws:
By the way, it doesn't have to be in here, but can we get the TBB travis tests turned back on?

davidbeckingsale:
Sure, let's get it in here. Should it be for all compilers?

trws:
It doesn't have to be, since it's the same runtime for all of them it could be just a subset as long as it covers both cuda and non I would think.

davidbeckingsale:
I added it everywhere, let's see what happens.

willkill07:
@davidbeckingsale it worked but I think we're approaching the 4MB log limit with some spew by gtest with NVCC. It doesn't need taken care of immediately, but if we want to be able to analyze travis logs, we'll need to (somehow) clean it up eventually.

davidbeckingsale:
Sure thing. I think this popped up somewhere else too.

willkill07:
Yeah, it happened in forall reimagined -- we suppressed the output from coverage and that helped. but with warnings in templates for gtest, the size quickly grows. Regardless, not a big deal now, but I'll look into correcting some of the warnings present.

PR #347, opened 2017-10-11
merge master back into develop

trws:
Is this what you had in mind @davidbeckingsale?

PR #342, opened 2017-10-11
Merge release/0.4.0 changes to develop

davidbeckingsale:

PR #348, opened 2017-10-11
Bugfix/trws/remove semicolon for pedantic

trws:
Replaces #347 to deal with changes not merged into master.


davidbeckingsale:
This supersedes #342 right? If so, we can close it and merge this as soon as Travis passes.

trws:
It does, at least as far as I can tell it contains all of those commits.

PR #361, opened 2017-11-11
make typed ranges work with IndexValue, remove base_iterator, add inn…

trws:
…er-aliases for ranges and lists in the IndexValue type for convenient use

This makes the following example a valid program:

```c++
#include <RAJA/RAJA.hpp>
#include <iostream>

RAJA_INDEX_VALUE(ID, "ID");
RAJA_INDEX_VALUE(IZ, "IZ");
RAJA_INDEX_VALUE(IG, "IG");
RAJA_INDEX_VALUE(IM, "IM");

{
  using RAJA::nested::For;
  RAJA::nested::forall(
      RAJA::nested::Policy<For<3, RAJA::seq_exec>,
                           For<2, RAJA::seq_exec>,
                           For<1, RAJA::seq_exec>,
                           For<0, RAJA::seq_exec> >{},
      camp::make_tuple(
          ID::range(0, 3), IZ::range(0, 3), IG::range(0, 3), IM::range(0, 3)),
      [=](ID d, IZ z, IG g, IM m) {
        std::cout << *d << *z << *g << *m << std::endl;
      });

  return 0;
}
```

### DIFF
--- a/include/RAJA/index/IndexValue.hpp
+++ b/include/RAJA/index/IndexValue.hpp
@@ -28,6 +28,8 @@
 
 #include "RAJA/config.hpp"
 
+#include "RAJA/index/RangeSegment.hpp"
+#include "RAJA/index/ListSegment.hpp"
 #include "RAJA/util/defines.hpp"
 #include "RAJA/util/types.hpp"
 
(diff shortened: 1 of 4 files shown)


trws:
Assuming all the platforms pass, I'm pretty comfortable with this change.  I would really like @ajkunen's approval though since it touches IndexValue support.  With this, both using TypedFor policies in a nested call and using typed ranges should work.  Also, I rather like having the range, list and strided_range types as dependent types, it takes a lot of the crud out of using it this way IMO.

trws:
Assuming all the platforms pass, I'm pretty comfortable with this change.  I would really like @ajkunen's approval though since it touches IndexValue support.  With this, both using TypedFor policies in a nested call and using typed ranges should work.  Also, I rather like having the range, list and strided_range types as dependent types, it takes a lot of the crud out of using it this way IMO.

trws:
Well... Everything passes, we can tweak later if there are any concerns, but I'm going to merge for now so we can move forward.

PR #363, opened 2017-11-13
RAJA shared memory

artv3:
I was curious to see if I could incorporate GPU shared memory into the RAJA programming model and still have it work on the CPU. With some macro magic, I took the finite difference wave equation solver and added shared memory (the file is waveShared.cpp) . I was wondering if there were any thoughts on this?  This code will run on the CPU with OMP and on the GPU.


examples/waveShared.cpp
```diff
@@ -0,0 +1,441 @@
...
+      
+        /*
+          Coefficients for a fourth order stencil
+        */
+        double coeff[5] = {
+          -1.0 / 12.0, 4.0 / 3.0, -5.0 / 2.0, 4.0 / 3.0, -1.0 / 12.0};
+
+        //int lx, ly; //local index
+        int lx; int ly;
+        RAJA_shared double Lu[By+2*sr][By + 2*sr];
```
davidbeckingsale: Instead of this being a macro, could we do something funny with an object that gets compiled differently for host and device?

examples/waveShared.cpp
```diff
@@ -0,0 +1,441 @@
...
+        double coeff[5] = {
+          -1.0 / 12.0, 4.0 / 3.0, -5.0 / 2.0, 4.0 / 3.0, -1.0 / 12.0};
+
+        //int lx, ly; //local index
+        int lx; int ly;
+        RAJA_shared double Lu[By+2*sr][By + 2*sr];       
+
+
+        //RAJA inner loop
+        RAJA_Par_Inner(ly, By) {
```
davidbeckingsale: Could this just be a `forall` overload? Again, the __device__ could use the thread indices, and the non-device could use some subset of the CPU policies?

examples/waveShared.cpp
```diff
@@ -0,0 +1,441 @@
...
+              if(ly < 2*sr)
+                Lu[ly + By][lx + Bx] = P2[nY2*nx + nX2];
+            }
+            
+            if(ly < 2*sr)
+              Lu[ly + By][lx] = P2[nY2*nx + nX1];                      
+                        
+          } 
+        }
+        RAJA_sync;
```
davidbeckingsale: We've talked about having `RAJA::synchronize<POL>()` for a while for host-side synchronization. Maybe we could generalize something like that and use it here.

examples/waveShared.cpp
artv3: Sure, I wasn't sure if making it a lambda would work though. This was just a quick prototype of what I though could work. This code works and hopefully serves as a starting point.

examples/waveShared.cpp
```diff
@@ -0,0 +1,441 @@
...
+#if CPU
+  //RAJA will give outer Id 
+ RAJA::forallN<fdNestedPolicy>(
+      outerBounds, outerBounds, [=] (RAJA::Index_type outerIdy, RAJA::Index_type outerIdx) {
+#else
+  //RAJA will give global id
+  RAJA::forallN<fdNestedPolicy>(
+      outerBounds, outerBounds, [=] __device__ (RAJA::Index_type gIdy, RAJA::Index_type gIdx) {
+        int outerIdx = blockIdx.x; //What I really want is the block id instead of thread of id... 
+        int outerIdy = blockIdx.y; //stick these in here...
```
artv3: @davidbeckingsale, here I would like to modify the indices to traverse though cuda block Id's instead of the the thread id's. These loops could be referred to as the outer loops.

trws:
I'm glad to see this, but I'm not sure it's something we want to endorse by putting it in the examples necessarily.

artv3:
@trws, I can move it some where else. I just needed a space to try out ideas. Where should I move it to?

davidbeckingsale:
I think the example was just being used as a place to try stuff, since this piece of code could make use of shared memory effectively.

Using this branch to prototype ideas makes sense to me.

trws:
It seems like a useful experiment, and having it up here is fine. If we merge it into the examples directory though, then it becomes something we're showing as an example of how RAJA code should be written.  If it's something we want to show people but not necessarily endorse, then maybe an experiments directory? I'm not sure, it just seems like the kind of thing we probably don't want to recommend until and unless we have a solution we intend to maintain for it.

artv3:
Gotcha, this was definitely a sketch on how I thought shared memory could be incorporated into the RAJA programming model. The purpose of this PR was just to share and bounce around ideas.

DavidPoliakoff:
A suggestion: I know I've done similar things in the past, might be wise to have "demo" branches to communicate we aren't discussing a merge.

I'd also think an "experiments" folder could be a good idea, it'd give us more power to discuss things that are in rough states.

artv3:
Just to follow up, do we like this idea?  I can continue to carve this out.

ajkunen:
Sorry, I haven't had a chance to give it a look.  I will have some time tomorrow.

One comment:  Macros aren't going to work, especially if someone is trying to switch between CPU and GPU with the same code, and the same build.  

Is there a way to do this with templates?  and use a policy to determine if it grabs CUDA shared vs just an array on the CPU?

artv3:
@ajkunen, I hope so. This was just an initial sketch on how we would have to lay out loops in order to introduce "portable" shared memory.

artv3:
Hey everyone, @MrBurmark  and I chatted about this today. We made progress towards making it more RAJA-like, we observed a 1.6x  speed up on two-dimensional wave equation example by using shared memory on a grid size of 14400 x 14400. We expect that three-dimensional version would be more efficient as well as a higher-order version (Currently it's 2nd order in time 4th order in space).

ajkunen:
@artv3 I just pushed some updated (and merged in develop).

I added a SharedMemory class that worked with CUDA like we talked about this morning... and oddly enough it works :)

I put two examples in at the end of test/unit/nested.cpp  Please check them out and let me know what you think.

We would still need a multiple-lambda launcher to help get rid of the __syncthreads() and make it portable, but this at least demonstrates what you wanted to do

artv3:
@ajkunen, works for my example as well! I can start cleaning up the code a bit. Can we do something similar for the CPU case?  I.e. Allocate memory prior to parallel regions to create shared memory among threads?

ajkunen:
@artv3 there is a seq_shmem policy that makes a bare array.  I don't know the proper way to make this work for OpenMP and CPU threads.  Do you want to try and figure that one out?  We could make a shared_memory.hpp in the policy/openmp directory

artv3:
@ajkunen the bare array is actually what I had in mind. One thing I will work on next is adding thread private memory.

ajkunen:
@artv3 I have some improvements to shared memory, and a version of RAJA::nested::forall that allows a sequence of multiple loops.  Do you mind if I merge that into this branch today?  I have some unit tests, but I would like to see if you could adapt them for your example.

artv3:
@ajkunen sure go for it 👍

include/RAJA/pattern/shared_memory.hpp
```diff
@@ -38,38 +40,105 @@ namespace RAJA
...
-  std::shared_ptr<std::vector<T>> data;
+  /*!
+   * Constructs from a list of index sizes
+   */
+  template<typename ... SizeTypes>
+  RAJA_INLINE
+  SharedMemoryView(SizeTypes ... sizes) :
+    layout(sizes...),
+    shared_memory(layout.size())
+  {}
```
artv3: Does this work like the RAJA shared memory view?

PR #369, opened 2017-11-28
Bugfix/kunen1/cuda misc

ajkunen:
Fixed an iteration issue with the RAJA::nested::forall when a segment starts at anything other than 0, and has a stride anything other than 1.

Fixed a mismatch between forward declares using struct vs. class

Snubbed the remaining host/device warnings for CUDA


ajkunen:
@artv3 I think this PR fixes the issues you were seeing with nested::forall.  Could you give it a try and let me know.

PR #374, opened 2017-12-11
Fix: OpenMP scans broken when num threads didn't match requested num threads

ajkunen:
Building manually on rzhasgpu with nvcc+gcc-4.9.3 failes test-scan.exe when running more than 1 thread.

I don't see this building just gcc-4.9.3 w/ openmp.

Just checking that our CI is running w/ threads.


ajkunen:
The problem here:  it looks like the scans were implemented to request a number of threads using the num_threads(p) pragma clause.  The algorithm assumes that the OMP runtime gave us that many threads... however, in some cases, it doesn't.  I've modified the code to still request some number of threads, then check to see how many have been allocated.

Seems to cause a problem when building nvcc+gcc-4.9.3, where we're not running ctest.  In all other cases the runtime seems to obey the num_threads(p) request.

PR #390, opened 2018-01-09
added binning example

artv3:


artv3:
@rhornung67 , I added a binning example for RAJA atomics. I was wondering what number it should take?

PR #401, opened 2018-01-22
Unit test for ListSegment and IndexSet

artv3:
Hi everyone, I am currently putting together unit-test for TypedListSegment  and RAJA::nested loops. Currently RAJA::nested::forall does not work with TypedListSegments. The first error I get is the following: 

/g/g17/vargas45/Git-Repos/RAJA/unitTest/include/RAJA/pattern/nested/internal.hpp(83): error: name followed by "::" must be a class or namespace name.

I will be looking into this.


trws:
Interesting, we need a better error message for this case.  That error is because your policy, `Pol{}`, contains no For types nor any types with forlist members.  It would probably be good to add a static assert for that.

artv3:
@trws , The TypedFor's were recently removed. Is there a work around for this?

trws:
Found it.  The gcc7 error message makes it pretty clear, the list segment uses a pointer type directly as its iterator type, and we directly use container::iterator::value_type, which is invalid for a bare pointer.  All it needs is either a check for that case, or a switch to using std::iterator_traits<container::iterator>::value_type instead.

trws:
In `iterable_value_type_getter`.

artv3:
@trws , I wasn't 100% on what container should be, but the following works:
struct iterable_value_type_getter {  
  using type = typename std::iterator_traits<typename Iterator::iterator>::value_type;
 };

I was wondering if this is what you had in mind?

artv3:
Almost got through adding ListSegment unit tests but ran into an issue when combining CudaCollapse with ListSegment : 
error: An extended __device__ lambda cannot be defined in a member function that has private or protected access within its class ("Nested_ListSegments4_Test")

trws:
That's exactly what I meant yes.

trws:
Shouldn't that test be a `CUDA_TEST` rather than a `TEST`?  I think the error is because the normal test macro creates a private method on the test class.

PR #405, opened 2018-01-26
TBB forall:nested

jeffhammond:
I'm having trouble figuring out if the unit tests even cover this, but I figure the easiest way to answer that question is to submit a pull request 😄 

This changeset was created by pattern-matching the OpenMP implementation, in the same manner that TBB's `forallN` was created.

This does not implement grainsize.  I'll deal with that on a subsequent pass.

Changing `RAJA_SIMD` for the Intel compiler when OpenMP is disabled is not directly related but since that is most likely to appear when TBB is used, it is part of this.

### DIFF
--- a/include/RAJA/config.hpp.in
+++ b/include/RAJA/config.hpp.in
@@ -174,7 +174,7 @@ const int DATA_ALIGN = @RAJA_DATA_ALIGN@;
 #define RAJA_SIMD  _Pragma("omp simd")
 #define RAJA_NO_SIMD _Pragma("novector")
 #else
-#define RAJA_SIMD _Pragma("simd")
+#define RAJA_SIMD _Pragma("vector")
 #define RAJA_NO_SIMD  _Pragma("novector")
 #endif
 
--- a/include/RAJA/policy/tbb.hpp
+++ b/include/RAJA/policy/tbb.hpp
@@ -32,6 +32,7 @@
 
 #include "RAJA/policy/tbb/forall.hpp"
 #include "RAJA/policy/tbb/forallN.hpp"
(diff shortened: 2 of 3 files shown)


DavidPoliakoff:
Hey Jeff, is this PR targeting the same nested semantics as #378 ? Excuse the ignorance, I just want to make sure that parallel developments are working together. I'm not involved in the planning of this, but I'll ping @ajkunen to make sure everybody's on the same page

jeffhammond:
I haven’t looked at that PR yet. I was merely trying to make my code written to forall::nested (as found in develop) compile with the TBB backend.

I confess that, as an outsider, the whole nested transition is a lot of thrash on every level and I’m having a hard time keeping up. I’m basically starting over from scratch with the PRK ports.

Under different circumstances, I’d just wait until everything was done to make the switch, but as you probably know, there’s this CORAL-2 thing and I have to understand Kripke from top to bottom.

DavidPoliakoff:
That's completely understandable, we are rapidly changing that interface and I'm sorry if that's causing you any grief. I'm not sure about any of the project planning aspects, I'll defer that to wiser minds (hey @rhornung67 ), but I hope we find a development path that works for everybody

ajkunen:
@jeffhammond If you develop TBB for the nested::forall in develop, it shouldn't be too terribly hard to translate that to the reworked nested::forall that's under development.  So I wouldn't call it wasted work.

We're planning on having the reworked nested::forall finished over the next monthish, and at the same time deprecate forallN.

jeffhammond:
@ajkunen I don't consider it a waste.  I wrote the TBB version in part to warm up my brain for other developments.

PR #407, opened 2018-01-30
added reset method and default constructor to reducers

artv3:


trws:
Sync for cuda?

Also, do we need to lose the constexpr on the constructors? It may not be a big difference, but pre-computing those at compile time made a positive difference at least at one point.

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -203,13 +213,23 @@ class BaseCombinable
   T mutable my_data;
 
 public:
-  //! prohibit compiler-generated default ctor
-  BaseCombinable() = delete;
 
   RAJA_HOST_DEVICE
-  constexpr BaseCombinable(T init_val, T identity_ = T())
-      : identity{identity_}, my_data{init_val}
```
willkill07: Constructors should use member initializers wherever possible, especially if marked constexpr (otherwise static constexpr initialization isn't feasible).

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -203,13 +213,23 @@ class BaseCombinable
...
+  }
+
+  RAJA_HOST_DEVICE
+  BaseCombinable(T init_val, T identity_ = T())
+  {
+    reset(init_val, identity_);
+  }
+  
+  RAJA_HOST_DEVICE
+  void reset(T init_val, T identity_){
```
willkill07: this can be made constexpr by using the comma operator along with additional casts to void. At the cost of readability in this one function, we gain constexpr support. I know there are trade offs, but it might be a worthwhile one.

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -374,10 +398,14 @@ class BaseReduceMaxLoc
   using value_type = typename Base::value_type;
   using Base::Base;
 
-  //! constructor requires a default value for the reducer
+  BaseReduceMaxLoc()
+  {
+    Base::reset(value_type(T(),Index_type()));
```
willkill07: could be marked constexpr with change to member initialization list.

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -374,10 +398,14 @@ class BaseReduceMaxLoc
...
+  }
+
   BaseReduceMaxLoc(T init_val, Index_type init_idx)
-      : Base(value_type(init_val, init_idx))
```
willkill07: I would revert this back

include/RAJA/policy/cuda/reduce.hpp
```diff
@@ -603,20 +603,26 @@ struct Reduce_Data {
...
+  Reduce_Data(T initValue, T identity_)
   {
+    reset(initValue, identity_);
+  }
+
+  void reset(T initValue, T identity_ = T()){
+    value = initValue;
+    identity = identity_;
+    device_count = nullptr;
+    device = device;
```
willkill07: What's the purpose of the no-op?

include/RAJA/pattern/detail/reduce.hpp
willkill07: The tradeoff here is that we are assuming T has constexpr constructors -- since this isn't always true (here's looking at you, arrays), maybe it wouldn't matter and you can just ignore my comment here :)

include/RAJA/policy/cuda/reduce.hpp
```diff
@@ -603,20 +603,26 @@ struct Reduce_Data {
...
-        identity{identity_},
-        device_count{nullptr},
-        device{},
-        own_device_ptr{false}
+  Reduce_Data(T initValue, T identity_)
   {
+    reset(initValue, identity_);
+  }
+
+  void reset(T initValue, T identity_ = T()){
```
willkill07: I wasn't initially sure this would work -- so long as the CUDA tests pass, LGTM 👍

artv3:
Thanks @willkill07 , @trws . Once AppVeyor gets fixed I think the branch will be good to go.

artv3:
@davidbeckingsale, the gcc-6 Travis build is running out of virtual memory. Have you guys encountered this problem before?

rhornung67:
@artv3 you can try restarting the job in travis (I just did this). If that doesn't work, you can click on the 'more options' tab at the upper right and clear the cache associated with the PR and restart.

trws:
Only when the make process was being run with `-j` rather than `-j 3`, 
and then only with nvcc.  Did it somehow end up running too many 
processes again?

artv3:
I am pretty sure this good to go, but Travis keeps running out of virtual memory with the gcc-6 compiler.

trws:
This last try, it actually stalled on g++-6 for 10+ minutes.  It might be good to try and repro that before merging this in case we have a gcc-6 user somewhere.

rhornung67:
@artv3 you could grab the RAJA perf suite and build it with gcc6 on one of our toss3 machines. build script for this is included in the scripts directory.

artv3:
Ready to go :-)

trws:
Do I understand right that it's a minloc/maxloc issue that's causing gcc-6 to stall? Is it an issue with the new default constructor or something?

artv3:
I believe it is the minloc/maxloc and I do not believe it is the new constructors. I created a new pull request with a branch from develop illustrating the issue PR #415 .

PR #415, opened 2018-02-01
reporting bug with reducers (DO NOT MERGE)

artv3:
I think I found the issue. Stay tuned for Travis reports.....


artv3:
@ajkunen , @rhornung67  , it looks like if we have more than one type of reducer we run into a stalling issue with gcc-6. This branch was just made from develop.

artv3:
It also seems it is the ReduceMinLoc and ReduceMaxLoc that are the problem, not so much the ReduceSum variables.

davidbeckingsale:
Can this be closed? @artv3

artv3:
Yup! This doesn't happen with the new kernel framework.

PR #416, opened 2018-02-02
Layout/View example - WIP

artv3:


examples/ex-layout.cpp
```diff
@@ -0,0 +1,197 @@
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
```
artv3: @davidbeckingsale , @rhornung67  I was wondering if you might be able to checkout this example. I was wondering if it looked good, if so I can clean up the view/layout page around this example.

examples/ex-layout.cpp
```diff
@@ -0,0 +1,197 @@
...
+//
+// Allocate 1D arrays
+//
+  int array_len = Istride*Jstride*Mstride*Nstride;
+
+  int* A = new int[array_len];
+  int* B = new int[array_len];
+
+  //Wrap A in a view
+  RAJA::View<int, RAJA::Layout<dim> > A_IJMN_view(A, Istride, Jstride, Mstride, Nstride);
```
davidbeckingsale: This isn't a very good name for the view.

examples/ex-layout.cpp
```diff
@@ -0,0 +1,197 @@
...
+//
+  RAJA::RangeSegment Irange(0,Istride);
+  RAJA::RangeSegment Jrange(0,Jstride);
+  RAJA::RangeSegment Mrange(0,Mstride);
+  RAJA::RangeSegment Nrange(0,Nstride);
+
+//
+// Allocate 1D arrays
+//
+  int array_len = Istride*Jstride*Mstride*Nstride;
```
davidbeckingsale: Why i,j,m,n instead of i,j,k,l?

examples/ex-layout.cpp
```diff
@@ -0,0 +1,197 @@
...
+  RAJA::View<int, RAJA::Layout<dim> > A_IJMN_view(A, Istride, Jstride, Mstride, Nstride);
+
+  //Create layout objects assuming an IJMN indexing
+  RAJA::Layout<dim> IJMN_layout = RAJA::make_permuted_layout({Istride,Jstride,Mstride,Nstride}, RAJA::as_array<RAJA::Perm<0,1,2,3>>::get());
+  //Equivalent to : RAJA::Layout<dim> IJMN_Layout(Istride,JStride,MStride,Nstride);
+
+//----------------------------------------------------------------------------//
+//
+// In the following example we show how a layout
+// object may be used to convert a 1D index to a 4D index.
```
davidbeckingsale: I don't think the layout -> indices conversion is so useful inside of a loop. There should be a better way to highlight this.

artv3:
@davidbeckingsale , I took another pass at the example. Although, I am not sure how to use the view with multiple layouts.

PR #426, opened 2018-02-20
CUDA 9 Warnings

artv3:
I found warnings when compiling with CUDA 9. This PR removes the warnings.


davidbeckingsale:
Hey @artv3 - thanks for this. I'll update this PR to also build with CUDA 9 in Travis before merging.

davidbeckingsale:
Doesn't look like the CUDA 9 package is available yet so we can't add this to Travis. Let's just make a note to test this internally.

Issue #353, opened 2017-10-24
non-loop executors for nested, particularly OMP parallel


trws:
For an initial version, all we need is a specialization of `nested::Executor` for `OMP_Parallel`, or preferably `omp_parallel`, or even better yet `RAJA::omp::parallel`.  All the operator should need to do is this:

```c++
#pragma omp parallel
body();
```

artv3:
@trws , @ajkunen , should I hold off on RAJA region until the new nested framework is in or is this something I can do within the existing framework?

ajkunen:
I think you could go ahead with trying to make a RAJA::region() function..

This API call should be independent of how we end up working it into the nested loop stuff.

I would just stay away from any of the RAJA::nested::* until that work is merged in.


-Adam

rhornung67:
@artv3 Ditto on what @ajkunen said.  As a first test of this, we have two kernels in the RAJA Perf Suite that would benefit from opening an OpenMP parallel region and then launching the loops inside it with an omp for nowait policy.

trws:
One thing I’d say here, as a test of whatever you do, refactor the 
existing internal omp parallel to use it.  That should give it a decent 
sanity check for interactions with everything else.

ajkunen:
So I'm assuming we all agree this should look something like:
```

RAJA::region<policy>([=](){

 // Some code 

});
```

rhornung67:
That's what I am thinking.

artv3:
I think OMP should be in the namespace since this is an OMP specific feature. A user might try to use a CUDA policy.

rhornung67:
@artv3 I would think that we would have to do a specific specialization for each policy type. If we don't provide a generic implementation, then unsupported usage would not compile.

rhornung67:
@artv3 can this issue be closed?

artv3:
Yup, I'll close it.

PR #431, opened 2018-02-23
RAJA::Region

artv3:
I ended up introducing omp/seq_region mostly because I wasn't sure which of the existing execution policies would be be a good fit here.


include/RAJA/policy/sequential/policy.hpp
```diff
@@ -47,6 +47,12 @@ namespace sequential
 /// Segment execution policies
 ///
 
+struct seq_region : make_policy_pattern_launch_platform_t<Policy::sequential,
+                                                          Pattern::forall,
```
davidbeckingsale: `Pattern::forall` is incorrect, no?

include/RAJA/policy/openmp/policy.hpp
```diff
@@ -76,6 +76,10 @@ struct Distribute {
 //////////////////////////////////////////////////////////////////////
 //
 
+struct omp_region
+    : make_policy_pattern_t<Policy::openmp, Pattern::forall, omp::For> {
```
davidbeckingsale: This shouldn't be `Pattern::forall`.

include/RAJA/pattern/region.hpp
```diff
@@ -0,0 +1,68 @@
+/*!
+ ******************************************************************************
+ *
+ * \file
+ *
+ * \brief   Header file containing the RAJA Region API call
+ *
+ *          The templates for segments support the following usage pattern:
```
davidbeckingsale: "segments" looks like a copy-paste error?

include/RAJA/policy/openmp/policy.hpp
```diff
@@ -190,6 +194,7 @@ using policy::omp::omp_for_exec;
 using policy::omp::omp_for_nowait_exec;
 using policy::omp::omp_for_static;
 using policy::omp::omp_parallel_exec;
```
davidbeckingsale: I thought the idea was to use the omp_region (should probably be called `omp_parallel_region`) to implement some of the other `omp_parallel` policies?

artv3: I was aiming for something  mentioned in #353 .

include/RAJA/policy/openmp/policy.hpp
artv3: I can create a new pattern: Pattern::region.

davidbeckingsale: Yup, that makes sense.

include/RAJA/policy/openmp/policy.hpp
davidbeckingsale: Yup, and @trws suggested:
> One thing I’d say here, as a test of whatever you do, refactor the
existing internal omp parallel to use it.  That should give it a decent
sanity check for interactions with everything else.

artv3:
`template <typename Iterable, typename Func, typename InnerPolicy>
RAJA_INLINE void forall_impl(const omp_parallel_exec<InnerPolicy>&,
                             Iterable&& iter,
                             Func&& loop_body)
{
#pragma omp parallel
  {
    forall_impl(InnerPolicy{}, std::forward<Iterable>(iter), body.get_priv());
  }
}`

@trws is the code you had in mind?

include/RAJA/policy/openmp/policy.hpp
artv3: Oh I think know what you mean, internally have it do the following 
RAJA::region<omp_region>([=](){
    forall_impl(InnerPolicy{}, std::forward<Iterable>(iter), body.get_priv());
});

right?

include/RAJA/policy/openmp/forall.hpp
```diff
@@ -66,12 +67,15 @@ RAJA_INLINE void forall_impl(const omp_parallel_exec<InnerPolicy>&,
...
 {
-#pragma omp parallel
-  {
-    using RAJA::internal::thread_privatize;
-    auto body = thread_privatize(loop_body);
-    forall_impl(InnerPolicy{}, std::forward<Iterable>(iter), body.get_priv());
-  }
+
+  RAJA::Region<RAJA::omp_parallel_region>([&](){
+
```
artv3: @davidbeckingsale @trws , I hope this is what you meant. The lambda has to be pass by reference otherwise it breaks.

davidbeckingsale: I think that's what @trws meant. How does it break if you capture by value?

artv3: I get compilation errors, in particular the first error is:  
error: no matching function for call to 'forward(const RAJA::nested::IterableTiler<const RAJA::TypedRangeStrideSegment<long int>&>&)'
       forall_impl(InnerPolicy{}, std::forward<Iterable>(iter), body.get_priv());

davidbeckingsale: You can capture just the `iter` by reference and it works. I'd defer to @trws on what the most efficient implementation is.

include/RAJA/policy/openmp/forall.hpp
```diff
@@ -66,12 +67,15 @@ RAJA_INLINE void forall_impl(const omp_parallel_exec<InnerPolicy>&,
...
                              Func&& loop_body)
 {
-#pragma omp parallel
-  {
-    using RAJA::internal::thread_privatize;
-    auto body = thread_privatize(loop_body);
-    forall_impl(InnerPolicy{}, std::forward<Iterable>(iter), body.get_priv());
-  }
+    
+  RAJA::Region<RAJA::omp_parallel_region>([=, &iter](){
```
artv3: @trws , I had to capture 'iter' by reference in order to get the code to compile. I was wondering if you might have a suggestion for an alternative implementation.

trws: I'm not sure what accounts for that.  One note is that the thread_privatize should be moved into the parallel region code, it's a property of launching the threads rather than specifically a loop.  Maybe send me something with the error, I'm not sure why there would be an issue with copying the iterable at this point in the code...

davidbeckingsale: The error was coming from trying to `forward` the iter, which was no longer a universal reference after being captured by value.

trws: Oh.  Then ~~std::forward(~~ iter ~~)~~ and capture it by value.

davidbeckingsale: Do you think that will have performance implications?

trws: Only where regions are invoked many times inside loops, and nested[2] was architected to help make that not matter. Perhaps more importantly, if it's re-worked as I mentioned above, the reductions should work for parallel code inside regions like this even outside of explicit RAJA worksharing loops.

artv3:
Thanks @trws @davidbeckingsale ! I addressed the issue and RAJA region now captures by value internally. :-)

include/RAJA/policy/openmp/forall.hpp
```diff
@@ -66,12 +67,15 @@ RAJA_INLINE void forall_impl(const omp_parallel_exec<InnerPolicy>&,
...
-    auto body = thread_privatize(loop_body);
-    forall_impl(InnerPolicy{}, std::forward<Iterable>(iter), body.get_priv());
-  }
+    
+  RAJA::Region<RAJA::omp_parallel_region>([=](){
+
+      using RAJA::internal::thread_privatize;
+      auto body = thread_privatize(loop_body);
+      forall_impl(InnerPolicy{}, iter, body.get_priv());
+
```
artv3: @trws , removing std::forward didn't quite do the trick. The code didn't build.

trws: Ok, what's the error?

artv3: I got the following: 
/g/g17/vargas45/Git-Repo/RAJA/region/include/RAJA/pattern/detail/forall.hpp:23:39: error: no matching function for call to 'begin(const RAJA::nested::IterableTiler<const RAJA::TypedRangeStrideSegment<long int>&>&)'
   auto begin##SUFFIX = begin(CONTAINER);             \
                                       ^
/g/g17/vargas45/Git-Repo/RAJA/region/include/RAJA/pattern/detail/forall.hpp:27:40: note: in expansion of macro 'RAJA_EXTRACT_BED_SUFFIXED'
 #define RAJA_EXTRACT_BED_IT(CONTAINER) RAJA_EXTRACT_BED_SUFFIXED(CONTAINER, _it)
                                        ^
/g/g17/vargas45/Git-Repo/RAJA/region/include/RAJA/policy/openmp/forall.hpp:106:3: note: in expansion of macro 'RAJA_EXTRACT_BED_IT'
   RAJA_EXTRACT_BED_IT(iter);
   ^
/g/g17/vargas45/Git-Repo/RAJA/region/include/RAJA/pattern/detail/forall.hpp:23:39: note: candidates are:
   auto begin##SUFFIX = begin(CONTAINER);             \
                                       ^
/g/g17/vargas45/Git-Repo/RAJA/region/include/RAJA/pattern/detail/forall.hpp:27:40: note: in expansion of macro 'RAJA_EXTRACT_BED_SUFFIXED'
 #define RAJA_EXTRACT_BED_IT(CONTAINER) RAJA_EXTRACT_BED_SUFFIXED(CONTAINER, _it)
                                        ^
/g/g17/vargas45/Git-Repo/RAJA/region/include/RAJA/policy/openmp/forall.hpp:106:3: note: in expansion of macro 'RAJA_EXTRACT_BED_IT'
   RAJA_EXTRACT_BED_IT(iter);

trws: The relevant part of the error is that begin, and end, discard qualifiers.  `IterableTiler::begin` and end weren't marked const, even though they don't modify anything.

include/RAJA/pattern/region.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include "RAJA/util/types.hpp"
+
+#include "RAJA/util/chai_support.hpp"
+
+
+namespace RAJA
+{
+  
+  template<typename ExecutionPolicy, typename LoopBody>
+  void Region(LoopBody&& loop_body)
```
davidbeckingsale: Can we change this to `region` (lowercase)?

include/RAJA/pattern/region.hpp
```diff
@@ -0,0 +1,62 @@
...
+  {
+    Region_impl(ExecutionPolicy(), loop_body);
```
davidbeckingsale: Lowercase.

include/RAJA/policy/openmp/region.hpp
```diff
@@ -0,0 +1,71 @@
...
+namespace policy
+{
+namespace omp
+{
+
+
+//
+//////////////////////////////////////////////////////////////////////
+//
+// Add description here...
```
davidbeckingsale: Doxgen comment (see #439 )

include/RAJA/policy/sequential/region.hpp
```diff
@@ -0,0 +1,71 @@
...
+//////////////////////////////////////////////////////////////////////
+//
+// RAJA sequential region
+//
+//////////////////////////////////////////////////////////////////////
+//
+
+
+template <typename Func>
+RAJA_INLINE void Region_impl(const seq_region &, Func &&body)
```
davidbeckingsale: Doxygen.

include/RAJA/policy/sequential/policy.hpp
```diff
@@ -47,6 +47,12 @@ namespace sequential
 /// Segment execution policies
 ///
 
+struct seq_region : make_policy_pattern_launch_platform_t<Policy::sequential,
+                                                          Pattern::region,
+                                                          Launch::undefined,
```
davidbeckingsale: Launch for a region is synchronous?

include/RAJA/policy/openmp/region.hpp
```diff
@@ -0,0 +1,71 @@
...
+// For details about use and distribution, please read RAJA/LICENSE.
+//
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#ifndef RAJA_region_openmp_HPP
+#define RAJA_region_openmp_HPP
+
+#include "RAJA/config.hpp"
+
+#include "RAJA/util/types.hpp"
```
davidbeckingsale: Why all these includes?

include/RAJA/pattern/region.hpp
```diff
@@ -0,0 +1,62 @@
...
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#ifndef RAJA_region_HPP
+#define RAJA_region_HPP
+
+#include <functional>
+#include <iterator>
+#include <type_traits>
+
+#include "RAJA/config.hpp"
```
davidbeckingsale: Why all the includes?

include/RAJA/pattern/region.hpp
```diff
@@ -0,0 +1,62 @@
+/*!
+ ******************************************************************************
+ *
+ * \file
+ *
+ * \brief   Header file containing the RAJA Region API call
+ *
+ *             region<exec_policy>(loop body );
```
davidbeckingsale: This should be in `\code` and `\endcode` blocks.

artv3:
Hey all, just wanted to follow up on this PR.

davidbeckingsale:
@artv3 - you don't want to have `master` merged into this branch....

artv3:
Sorry! forgot we keep things in develop....

trws:
Did you lose my commit that changed the begin method on the tiler or the 
range segment or whatever it is to be a const method? This looks a lot 
like the error you were getting before that.

artv3:
@trws , thanks. Things got shifted around and I just updated kernel/Tile.hpp

include/RAJA/policy/openmp/forall.hpp
```diff
@@ -66,12 +67,16 @@ RAJA_INLINE void forall_impl(const omp_parallel_exec<InnerPolicy>&,
...
                              Func&& loop_body)
 {
-#pragma omp parallel
-  {
-    using RAJA::internal::thread_privatize;
-    auto body = thread_privatize(loop_body);
-    forall_impl(InnerPolicy{}, std::forward<Iterable>(iter), body.get_priv());
-  }
+
+  RAJA::region<RAJA::omp_parallel_region>([=, &loop_body](){
```
artv3: @trws , @ajkunen . I was able to fix the bugs, but in order to achieve correct results I have to capture loop_body by reference. If I don't, this branch fails certain unit tests in Kernel.cpp.

artv3:
@ajkunen , @trws , I was able to overcome the capture by reference issue by introducing thread privatizers in the #pragma omp for methods in policy/openmp/forall.hpp

artv3:
@trws , I found a couple of other things that led to correct results: 
Inside the region call in include/RAJA/policy/openmp/forall.hpp, one may  
1. make a local copy of loop_body inside region (ex. 
      auto region_body = loop_body ) and use the local copy. 
2. make the lambda mutable 
3. Capture loop_body by reference. 

I'm not 100% sure why this does the trick. I am guessing that OMP may need to modify the loop_body inside the lambda call.

artv3:
@trws , @davidbeckingsale , @rhornung67 , @ajkunen . I just pushed in a 4th possible fix for RAJA region. It seems the lambda is having trouble capturing R values. As such I made a second version of RAJA region to be used internally with Kernel. This variant takes R-Value loop-bodies as a second argument in RAJA region and serves as an inner body for RAJA region's main lambda.

trws:
Good work @artv3, thanks for pushing through on this!

PR #433, opened 2018-02-26
Added occupancy calculator for CUDA backend

ajkunen:
Two main things added here:

1) added occupancy calculator to cuda_exec to choose number of blocks to launch.  Uses grid-stride technique for large iteration spaces.

2) Added a cuda_occ_exec that uses occupancy calculator to choose number of threads and blocks to launch.  Also uses grid-stride.

With no other modification (no changing of exec policies, just grid-stride), unit tests on rzmanta.llnl.gov went from taking ~9.3 seconds to ~7.7 seconds.

@rhornung67 please let me know if there are performance implication on the perf suite.

### DIFF
--- a/include/RAJA/policy/cuda/forall.hpp
+++ b/include/RAJA/policy/cuda/forall.hpp
@@ -68,7 +68,7 @@ namespace impl
  ******************************************************************************
  */
 RAJA_INLINE
-dim3 getGridDim(size_t len, dim3 blockDim)
+unsigned getGridDim(size_t len, dim3 blockDim)
 {
   size_t block_size = blockDim.x * blockDim.y * blockDim.z;
 
@@ -76,6 +76,7 @@ dim3 getGridDim(size_t len, dim3 blockDim)
 
   return gridSize;
 }
+
 
 /*!
  ******************************************************************************
@@ -90,6 +91,13 @@ __device__ __forceinline__ unsigned int getGlobalIdx_1D_1D()
   unsigned int threadId = blockId * blockDim.x + threadIdx.x;
   return threadId;
 }
+
+__device__ __forceinline__ unsigned int getGlobalIdx_1D_1D(unsigned int blockId)
+{
+  unsigned int threadId = blockId * blockDim.x + threadIdx.x;
+  return threadId;
+}
+
 __device__ __forceinline__ unsigned int getGlobalNumThreads_1D_1D()
 {
   unsigned int numThreads = blockDim.x * gridDim.x;
@@ -130,7 +138,7 @@ __device__ __forceinline__ unsigned int getGlobalNumThreads_3D_3D()
 /*!
  ******************************************************************************
  *
- * \brief  CUDA kernal forall template for indirection array.
+ * \brief  CUDA kernel forall template for indirection array.
  *
  ******************************************************************************
  */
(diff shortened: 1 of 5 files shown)


rhornung67:
@ajkunen I'll run this through the perf suite later today or tomorrow morning and report my findings.

trws:
Looks good to me.  Barring issues from somewhere else LGTM.

include/RAJA/policy/cuda/forall.hpp
```diff
@@ -141,14 +149,54 @@ template <size_t BlockSize,
 __launch_bounds__(BlockSize, 1) __global__
```
rhornung67: @ajkunen did you say that there are two occurrences of the launch bounds thing? This is the only one I see. 

Do we want to remove this or leave it in? I thought the idea was to prevent folks from being hit with mysterious register pressure/spillage.

ajkunen: That's the only one.  I'm not sure what to do with it.  I would leave it in for the time being.

rhornung67:
So I think this is a useful option for some things, but it definitely can have negative performance impacts. @ajkunen and I have been looking at some data I generated from the RAJA Perf Suite (which showed a significant slowdown across the board using cuda_occ_exec policy vs. cuda_exec<256>) and plan to dig into this deeper. The slowdown may be due to the cost of the occupancy calculator. We'll report more when we spend some time on this tomorrow.

trws:
It might be worth a look to see how much of this is due to the occupancy 
calculation and how much the new while loop.  For relatively small trip 
counts especially a loop like that can be unfortunate for performance 
(or at least could the last time I tested it).

rhornung67:
@trws I ran the perf suite with and without the occupancy calculator and saw a roughly 50% to 3-4x slowdown in the kernels. I ran with kernel size factor of 20 so they all had at least 2M iterations, which should be enough - P-100 GPU has 56 SMs * 2048 threads/SM = 114,688 total threads. I did a separate run where I just timed the call to cudaOccupancyMaxPotentialBlockSize(). That showed roughly 10-30 microseconds for each call, depending on the kernel. I assume that the RAJA timer resolution is fine enough that that is reasonably accurate. When I can track @ajkunen down, we'll do some more digging.

trws:
Were both of those tests using the version in this branch that has the "logical_block" loop in it, or one with and one without? That's actually the information I was asking about, if not terribly well, before.

rhornung67:
I compared (1) develop branch (no logical_block loop) to (2) Adam's branch with cuda_exec policies (no logical_block loop) to (3) Adam's branch with cuda_occ_exec policies (has logical_block loop). 1 and 2 were about the same (average over 3 runs - so there is some minor variation). 3 showed the slowdown I mentioned. All runs were done on the same rzmanta node, same compilers, options, etc.

rhornung67:
@MrBurmark that's in line with what I saw.

include/RAJA/policy/cuda/forall.hpp
```diff
@@ -68,7 +68,7 @@ namespace impl
  ******************************************************************************
  */
 RAJA_INLINE
-dim3 getGridDim(size_t len, dim3 blockDim)
+unsigned getGridDim(size_t len, dim3 blockDim)
```
rhornung67: @ajkunen is returning 'unsigned' when gridSize is 'size_t' potentially problematic?

jonesholger:
I think we should add a policy to this PR which supports explicit compile-time grid and block dim setup, to support the case where you may have previously run the occupancy calculator, but now want to avoid the calculator overhead, but still reap the benefits of grid-stride. We may also want to revert the default policy too which would require a new policy mapping for the case where blocksize (or blockdim) is provided by the user.

rhornung67:
Is there any case where grid-stride looping would not be a win regardless of whether user specified blocksize or it was determined via occupancy calculation? If not, then I think we should use grid-stride in all CUDA variants. Based on what I've seen in the Perf Suite and what @MrBurmark observed, I think the default case should revert to what it was originally and add the cuda_occ_exec policy to make the use of the calculator explicit for users who want to try it out. @ajkunen and I discussed having some blocksize magic number (e.g., 0 or -1, constant named of course) that would trigger the occupancy calculator, but this turned out to be troublesome.

There were a few other things that @ajkunen and I discussed that may be useful "halfway choices" between fully user-specified blocksize (with # blocks determined via what we have now) and relying entirely on the occupancy calculator. And they would be cheap at runtime. I'll work this with @ajkunen and we'll run these through the perf suite and some other examples.

It seems that there is a bit more exploration to do, then discuss what we think is most beneficial to users. I would like to avoid having a proliferation of exec policies.

jonesholger:
I think the counter example to grid stride would be branchy kernels where in grid stride mode the probability to impact a participating thread is higher, while spreading the pain via non grid stride is a win. But, I don't have this counter example in my back pocket to prove this. This was probably the one sticking point in earlier work involving grid-stride.

rhornung67:
@jonesholger that's an interesting case to know about. Does the IF_QUAD kernel in the perf suite cover this adequately? It has a potential branch divergence at each iterate, but I believe that the way the arrays are initialized it doesn't diverge nearly as much.

jonesholger:
That would be awesome if IF_QUAD covered this, otherwise we should keep an eye out for a perf suite candidate that does. I can check this later this week.

ajkunen:
I don’t think grid-stride has any adverse effect with heavily branch divergent code.

The divergence is re-joined at the point of the branch.  You shouldn’t see any branch divergence propagate past the grid-stride loop.


For example:

If(blah){    // Some threads execute here, while other stall
  …
}else{       // other threads execute here, while the prior stall
  …
}
// All threads have rejoined here


-Adam

trws:
I’m not sure I see what case that would be exactly.  Is the question 
just whether there should be a loop or not? The case that comes to mind 
where you would want a loop, and want it not to be grid-stride, is where 
the data isn’t laid out in thread order such that it allows 
coalescing, am I missing something?

trws:
That’s correct, though it actually does point out the other point 
where you might want to control ordering.  If at all possible you want 
runs of threads to take the same branch, but again that’s a property 
of the data more than it is the iteration ordering.

jonesholger:
Agree to bifurcating code where runs of threads that take the same branch are together (reorganize data).   Also, I guess I'm thinking cases where the balance isn't so heavily divergent, to avoid every block impacted. And to the case where the architecture allows a warp swap among active blocks due to a stall.

jonesholger:
Nonetheless Tom, you're providing impetus for a kernel design to highlight these trades

trws:
Fair enough, it certainly doesn’t hurt to test it.

ajkunen:
@rhornung67 You want to give this a try again?  I added a "better" occupancy calculator for cuda_exec which just gets a wag at the number of blocks.

PR #434, opened 2018-02-28
Set RAJA_LOADED in parent scope, remove CACHE

### LINKED ISSUE #430: force cache for RAJA_LOADED breaks things
https://github.com/LLNL/RAJA/blob/d329c9e91b86dfdb4f62bf91464e8b2faa60c03c/CMakeLists.txt#L25

So if we put RAJA_LOADED into cache, then it breaks if CMake ever decides to reconfigure... like after using ccmake... or a change to a CMakeLists.txt file.

Is this necessary?

davidbeckingsale:
This should fix #430


davidbeckingsale:
@MrBurmark - does this work for you?

PR #435, opened 2018-03-02
Fix for a bug in the xl compiler on power8.

koning:
This was necessary to get hypre to compile with xl/beta-2018.02.05 on power8.


davidbeckingsale:
Thanks @koning! Do you need this in 0.5.3, or would you be happy if we get out another release including this fix?

koning:
A new version is fine with me. We are updating the support libs often. 
Joe

davidbeckingsale:
Okay, great - I've changed the base to `develop`.

PR #436, opened 2018-03-05
fix new travis setup for external PRs

trws:
I accidentally broke the external PR check for intel, basically the secure info isn't there, so the openssl command fails and kills everything else.  This should fix it, and probably #435 also.


trws:
Yup, me too.  Hopefully this will just take care of it.

trws:
Ok... travis seems to be cranky, any idea what could cause it to just not respond @davidbeckingsale?

davidbeckingsale:
No good ideas, we can probably just wait.

davidbeckingsale:
@trws still looks like it's picking up Intel: https://travis-ci.org/LLNL/RAJA/jobs/349588648

trws:
I see that.  The odd part is that travis on that branch doesn't seem to be running the final updated version, or the new version I posted up there, but something between the two.  I think the update I posted over there will take care of it, but if it isn't picked up we may have to cherry-pick it over to develop to make everything work.

PR #440, opened 2018-03-09
Fixed tuple get-by-type, and make sure operators have returns

ajkunen:
I was getting warnings for no return value on operators.

Also, get compile errors for get-by-type with clang+nvcc.

This seems to fix things.


trws:
@davidbeckingsale, has something changed with appveyor that you know of? It looks like that test is dying on warnings here, and I don't recall that being an issue before.

willkill07:
There was a new MSVC release within the past two weeks. Perhaps the image was updated.

trws:
I think this has been subsumed since it was posted, all that seems to be left is a blt update.  Closing pending someone finding that I missed something.

PR #445, opened 2018-03-13
add clang 6 dockerfile and add to travis matrix

trws:
The subject says it all.


trws:
I have to say, I’m seriously fond of how easy that was to do.  Change 
the URL in the dockerfile, build, push, PR.  Whole process maybe five 
minutes… Really nice to not have to poke at travis and pray quite so 
much anymore.

PR #446, opened 2018-03-19
Intel - SIMD Bugs  (DO NOT MERGE)

artv3:
We have found bugs when paring RAJA::Kernels with OMP parallelism, and simd compiler hints. 
In particular: 
1. Incorrect output when using RAJA view 
2. Incorrect output when using the nested branch on a SW4 Kernel 

Source code for issues are found in simd-issues folder: 
view.cpp 
sw4Kernel.cpp 

To tryout  (assuming CMAKE 3.3 or higher is available) : 
git clone -b issue/Intel-bug https://github.com/LLNL/RAJA.git intel-bug
cd intel-bug 
bash ./scripts/intel-build.sh

cd build-intel

to run code: 
./simd-issues/bin/sw4Kernel
./simd-issues/bin/view

Comments: 
1. We find that a traditional nesting of RAJA forall statements will generate the correct output, but    using the RAJA::Kernel construct will not. 

2. Each .cpp file includes a KERNEL macro. When defined the code will use the RAJA::Kernel framework. When the macro is undefined the code will use a series of nested RAJA::forall loops.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,7 @@ if (NOT RAJA_LOADED)
   option(ENABLE_WARNINGS "Enable warnings as errors for CI" Off)
   option(ENABLE_DOCUMENTATION "Build RAJA documentation" Off)
   option(ENABLE_COVERAGE "Enable coverage (only supported with GCC)" Off)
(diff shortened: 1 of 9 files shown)


trws:
Closing to get it off the list, keeping history and branch for the moment.

PR #451, opened 2018-04-03
Remove vectorization hint when intel and omp is being used

artv3:
We have evidence of Intel compilers generating incorrect code when RAJA Kernel is paired with OMP and simd execution policies. 

Q: Should omp simd be supported in RAJA when using the Intel compiler?


davidbeckingsale:
Is correct code generated using `#pragma simd`? If so, we should use that even if OpenMP is available.

artv3:
@davidbeckingsale , simd will generate incorrect output. I just tried Intel's #pragma vector instead of #pragma omp simd and that seems to be closer to correct results.

davidbeckingsale:
Does it break for a trivial example, or is this for the sw4 stuff?

artv3:
@davidbeckingsale , it breaks @ajkunen 's view example where he does point-wise multiplication. 
Cview(k,j,i) = Aview(k,j,i)*Bview(k,j,i) ;

CMakeLists.txt
```diff
@@ -91,6 +91,12 @@ if (NOT RAJA_LOADED)
   if (ENABLE_OPENMP)
     set (raja_depends
       openmp)
+
+  if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
+    message(WARNING "RAJA::simd_exec not supported ")
```
davidbeckingsale: Can you make the message say something like "simd_exec not supported when OpenMP is enabled with the Intel compiler"

artv3:
Hi @davidbeckingsale and @rhornung67 can I get a re-approval on this branch? The updates I pushed dismissed the approval.

artv3:
@ajkunen , @trws , @rhornung67 , I tried out the proposed solution by Xinmin and Nikolay and it works for adding two vectors with the simd clause. Should we close this branch?

PR #454, opened 2018-04-06
deduplicate as_list alias and fix header include order

### LINKED ISSUE #453: duplicate aliases
Older Intel compilers (Intel 15) fail with the following error.

/include/camp/camp.hpp(227): error: "as_list" has already been declared in the current scope
  using as_list = typename as_list_s<T>::type;
        ^
 
As far as I can tell, the using as_list is declared in both tuple.hpp and list.hpp, both of which are included from camp_aliases.hpp.

Obviously the Intel 15 compiler is pretty old (three versions behind).  However, it seems like the declaration of this alias twice was not the original intent and points this alias being in the wrong place.

trws:
fixes #453 

If someone has icc15 handy, maybe test this, but the aliases are deduplicated.

### DIFF
--- a/include/camp/camp.hpp
+++ b/include/camp/camp.hpp
@@ -4,11 +4,11 @@
 #include <array>
 #include <type_traits>
 
+#include "camp/size.hpp"
 #include "camp/defines.hpp"
 #include "camp/helpers.hpp"
 #include "camp/lambda.hpp"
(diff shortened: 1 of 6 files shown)


trws:
Anyone on the merge button here? @davidbeckingsale, @rhornung67?

PR #457, opened 2018-04-11
Ensure consistency between libRAJA.a and host code

DavidPoliakoff:
@ajkunen @davidbeckingsale 

This seems a sane way to go about what we talked about at the meeting, now if people are using OpenMP in a file that isn't in policy/openmp/... that file includes a [file which checks for whether they enabled OpenMP in RAJA's config but not in their build](https://github.com/LLNL/RAJA/compare/develop...DavidPoliakoff:safe_openmp?expand=1#diff-18e058589d384a2ad501b9e3dd0b6abdR33).

I didn't want to put this at the point where RAJA_ENABLE_OPENMP is defined, as I could see somebody who isn't using CUDA going down this road, and they wouldn't get the linker error. This does impose a developer burden, if you use OpenMP outside of policy/openmp you need to include my file. 

Any thoughts on whether we want to do it this way, or whether we want to be more strict?


artv3:
I like this 👍

davidbeckingsale:
The check is good, but I'm not sure why it needs to be in a separate file, and why you include that file in `reduce` and `Memutils_CUDA`. Why not just do the check in `config.hpp`?

davidbeckingsale:
I guess I understand that a user might want to just use sequential from an OpenMP enabled build, in which case, I think the check should go in `RAJA/policy/openmp.hpp`. The supported way of selectively using features is to _not_ use `RAJA/RAJA.hpp` which includes everything, and instead to include the necessary `pattern` and `policy` headers manually.

DavidPoliakoff:
I could look at moving this there,I just don't know whether everything that includes OpenMP code includes RAJA/policy/openmp.hpp. Let me play with that a bit

DavidPoliakoff:
If I put the check in policy/openmp.hpp

```
#include "../include/RAJA/policy/cuda/forall.hpp"
#include "../include/RAJA/pattern/forall.hpp"
#include "../include/RAJA/policy/cuda/MemUtils_CUDA.hpp"

int main(){
  RAJA::forall<RAJA::cuda_exec<512>>(0,1,[=]__device__(int i){

  });
}
```

Will hit the linker error, not the compile time check.

DavidPoliakoff:
@davidbeckingsale and I had a conversation in a side channel, he seems to be leaning towards "put the check in config.hpp." This would make it a bug to build RAJA with OpenMP, and then in an application which only uses loop_exec include RAJA. 

Wanted to open that up to @trws, @artv3, @ajkunen , and @rhornung67 , I want at least three voices before I change RAJA in a way that could break user code.

DavidPoliakoff:
Because there's really no problem there. Essentially, our only bugs are when host code thinks a data member is process-wide, and libRAJA.a thinks it's thread-local. The number of places we define such data members is vanishingly small (policy/cuda/MemUtils_CUDA.hpp), so I'm leery of breaking builds.

DavidPoliakoff:
@davidbeckingsale : cool, people have had a chance to comment if they had problems with this. Moved the check to config.hpp

davidbeckingsale:
Hey @DavidPoliakoff , this looks good. My only comment would be shrinking the message. Maybe `RAJA configured with ENABLE_OPENMP, but OpenMP not supported by current compiler` or something?

PR #461, opened 2018-04-19
Fixed performance issue with Intel 16/17/18 compilers and Views 

ajkunen:
by removing the StrideOneDim optimization if RAJA_COMPILER_INTEL is defined

Also simplified ConditionalMultiply class to take two scalars instead of a scalar and array... this may have helped deconfuse icpc


trws:
Each of the two changes makes sense by itself, but am I reading correctly that this makes a fix to help intel in ConditionalMultiply, then turns off ConditionalMultiply if intel is detected?

Issue #464, opened 2018-04-24
Non-const -> Const View copy-constructor


timrlaw:
Hi there,

Is it possible to add a copy-constructor to the View class that allows construction of a View with a const `value_type` from one with identical layout, but a non-const `value_type`. As in the following:

```
RAJA_INLINE constexpr View(NonconstView const &rhs)
    : layout(rhs.layout), data(rhs.data)
{
}
```

where NonconstView is defined something like:

```
using nonconst_value_type = typename std::remove_const<value_type>::type;
using nonconst_pointer_type = typename std::add_pointer<typename std::remove_const<std::remove_pointer<pointer_type>::type>::type>::type;

using NonconstView = View<nonconst_value_type, layout_type, nonconst_pointer_type>;
```

The lack of this is the only thing preventing me from using RAJA views as a drop-in replacement in my application.

Cheers,
Tim

davidbeckingsale:
Hey @timrlaw, that sounds reasonable. Have you tried this in your local RAJA repo?

timrlaw:
Hi David (hope you're doing well!),

Yes appears to work fine. I"ve made a PR (#466 ).

My compiler does *not* delete the implicit copy constructor when adding the new one (since its not technically a copy constructor I suppose). I'm not sure I trust all compilers to do this though, so I've made it explicit with the `= default` cctor. Any thoughts on this? Doing so means the new cctor needs to be `std::enable_if`'d away when the current `value_type` is non-const, as the constructors have identical signatures in that case.

Cheers,
Tim

davidbeckingsale:
Closed by #466

PR #466, opened 2018-04-25
Allow construction of const Views from their nonconst analogues.

timrlaw:
Cf. issue #464


timrlaw:
Done. There were no existing View tests that I could find, so I created a new file.

PR #478, opened 2018-05-09
Reproducer/nvidia/offset - (DO NOT MERGE)

artv3:
Reproducer for NVIDIA. We found a bug when views are used in the device with offset layouts. In particular the generated copy constructor does not actually copy-construct the object. 

The follows compilers generate run time errors : 
CUDA 8, 9.0.176, 9.0.184, 9.1.76, 9.1.85 . 

In particular the following test fails /test/unit/cuda/test/test-cuda-forall-view.exe 

By defining FIX_NVCC_BUG in nvidia-bug/include/RAJA/util/View.hpp we are able to fix the bug, i.e. specifying the copy constructor. Removing FIX_NVCC_BUG macro  will cause run-time errors. 

 git clone -b reproducer/nvidia/offset https://github.com/LLNL/RAJA.git nvcc-bug
cd nvcc-bug
git submodule init && git submodule update
mkdir build
cmake -DENABLE_CUDA=On -DCUDA_TOOLKIT_ROOT_DIR=$(PATH_TO_CUDA) ../
make -j
./test/unit/cuda/test/test-cuda-forall-view.exe

### DIFF
--- a/include/RAJA/util/View.hpp
+++ b/include/RAJA/util/View.hpp
@@ -36,6 +36,9 @@
 #include "chai/ManagedArray.hpp"
 #endif
 
+#define FIX_NVCC_BUG
+#undef FIX_NVCC_BUG
+
 namespace RAJA
 {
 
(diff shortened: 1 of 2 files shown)


davidbeckingsale:
@artv3 - do we need to keep this PR open?

artv3:
@davidbeckingsale , I am not sure. I shared the issue on the COE compiler issue page and with an NVIDIA rep. I can check with him.

davidbeckingsale:
We don't need to delete the branch, but we can just close the PR.

artv3:
Oh gotcha. Sure we can close it. I can let him know it will still be there.

PR #486, opened 2018-05-23
ListSegment support for Kernel

artv3:
Modified ListSegment in order to be usable in Kernel.


include/RAJA/index/ListSegment.hpp
```diff
@@ -223,6 +234,24 @@ class TypedListSegment
...
+   */
+  RAJA_HOST_DEVICE RAJA_INLINE TypedListSegment slice(Index_type begin,
+                                                      Index_type length) const
+  {
+
+    //ListSegType idx_list(&idx[0], idx.size());
+    //auto start = m_data[0];
+    //auto end = start + length > m_end[0] ? m_end[0] : start + length;
+
+    return TypedListSegment(&m_data[0],length);
```
MrBurmark: should start at &m_data[begin]

include/RAJA/index/ListSegment.hpp
```diff
@@ -223,6 +281,23 @@ class TypedListSegment
...
+  /*!
+   * \return A new instance spanning *begin() + begin to *begin() + begin +
+   * length
+   */
+  RAJA_HOST_DEVICE RAJA_INLINE TypedListSegment slice(Index_type begin,
+                                                      Index_type length) const
+  {
+    Index_type end = begin+length > m_size ? (m_size-begin) : length;
+#if !defined(__CUDA_ARCH__)
+    return TypedListSegment(&m_data[begin], end, Owned);
```
davidbeckingsale: Is `Owned` correct?

include/RAJA/index/ListSegment.hpp
```diff
@@ -148,14 +148,29 @@ class TypedListSegment
...
   ///
-  /// By default the ctor performs deep copy of array elements.
+  /// By default the ctor on the host performs deep copy of array elements.
+  /// By default the ctor on the device does not perform a copy of array of elements. 
+  ///
+  RAJA_HOST_DEVICE TypedListSegment(const value_type* values,
+                                    Index_type length) 
+    : m_data{const_cast<value_type*>(values)}, m_size{length}, m_owned{Unowned}
+  {
+#if !defined(__CUDA_ARCH__)
```
davidbeckingsale: Should we make a `RAJA_DEVICE_CODE` define instead of using `__CUDA_ARCH__` directly? This seems like the kind of code change we would need for other devices too.

include/RAJA/index/ListSegment.hpp
artv3: I had initially tried Unowned but it did not work with CPU compilers. I think there might be a thread privatizing issue if its not owned.

davidbeckingsale: Semantically it's wrong because the slice is not supposed to be owning the data right.

artv3: I see, I think there may be a bug in tiling. Currently tiling relies on the slice method to do bounds checking. I changed the data ownership to Unowned and made a note on the .hpp file. Furthermore, the unit test are testing basic mechanics.

include/RAJA/util/defines.hpp
```diff
@@ -36,6 +36,9 @@
 // We need a better solution than this as it is a pain to manage
 // this stuff in an application.
 //
+#if defined(__CUDA_ARCH__)
```
davidbeckingsale: I would make this `#if defined(RAJA_ENABLE_CUDA)`

include/RAJA/index/ListSegment.hpp
```diff
@@ -223,6 +281,25 @@ class TypedListSegment
   //! accessor to retrieve the total number of elements in a TypedListSegment
   RAJA_HOST_DEVICE Index_type size() const { return m_size; }
 
+  //! Create a slice of this instance as a new instance but with no ownership
+  //! of the data it points to.
+  /*!
+   * \return A new instance spanning *begin() + begin to *begin() + begin +
+   * length
+   * 
+   * Current status: Tiling may not be distributing data correctly
```
davidbeckingsale: We should fix it... 😄

trws: Is it a "program might be slower than intended" or a "program will give wrong results silently" problem?

artv3: The program will segfault if data is not owned when combining tiling with Kernel.

trws:
Yeah, I’m thinking that’s something we should fix before this is 
merged either by making it so that wont happen or making it a 
compile-time error to use tiling on an index that maps to  ListSegment.

artv3:
@trws , How would we check if a user is using a ListSegment for tiling? To generate the compile time error?

include/RAJA/index/ListSegment.hpp
```diff
@@ -148,14 +148,30 @@ class TypedListSegment
...
-  /// By default the ctor performs deep copy of array elements.
+  /// By default the ctor on the host performs deep copy of array elements.
+  /// By default the ctor on the device does not perform a copy of array of elements. 
+  ///
+  RAJA_HOST_DEVICE TypedListSegment(const value_type* values,
+                                    Index_type length) 
+    : m_data{const_cast<value_type*>(values)}, m_size{length}, m_owned{Unowned}
+  {
+    
+#if !defined(RAJA_DEVICE_CODE) && defined(RAJA_ENABLE_CUDA)
```
artv3: @trws , @davidbeckingsale , @MrBurmark , I think this was the missing piece. I have this working working now with Tiling. The data is un-owned when compiling on the CPU.

artv3: But its owned when CUDA is enabled, I think there is a corner case that is missing  to get it to work correctly.

ajkunen:
@artv3 the iterable wrapper stuff is great.  We need to compare register usage with CUDA backend for things like RangeSegment.

artv3:
@ajkunen, credit to @MrBurmark, he coached me with it. OpenMP was destroying ListSegments early and it was causing some issues. This was the workaround.

include/RAJA/index/ListSegment.hpp
```diff
@@ -142,23 +142,28 @@ class TypedListSegment
...
   /// \brief Construct list segment from given array with specified length.
-  ///
-  /// By default the ctor performs deep copy of array elements.
   /// If 'Unowned' is passed as last argument, the constructed object
   /// does not own the segment data and will hold a pointer to given data.
   /// In this case, caller must manage object lifetimes properly.
-  ///
-  TypedListSegment(const value_type* values,
+  /// This constructor is host only since ownership may only be specified 
+  /// on the host.
```
MrBurmark: Comment no longer relevant.

artv3:
@davidbeckingsale , would you have some time to revisit this PR?

davidbeckingsale:
I think this is looking good now.

include/RAJA/internal/Span.hpp
```diff
@@ -60,26 +61,50 @@ struct Span {
...
-  return {data, size};
+RAJA_HOST_DEVICE RAJA_INLINE
+Span<ValueType, IndexType> make_span(ValueType begin, IndexType size)
+{  
+  return Span<ValueType,IndexType>(begin, begin + size);
+}
+
+template <typename ValueType, typename IndexType>
+RAJA_HOST_DEVICE RAJA_INLINE
+Span<ValueType, IndexType> make_span(ValueType begin, ValueType end)
```
MrBurmark: This maker function doesn't deduce IndexType.
It should be made to, or removed.

artv3:
@davidbeckingsale can I get a reapproval?

davidbeckingsale:
Don't think you need one, everything looks green 😄

Issue #372, opened 2017-12-06
OpenMP target code does not compile


jeffhammond:
# Summary

I cannot complete a build of RAJA with any compiler of interest (to me).

# Compilers Tested

```sh
$ gcc-7 --version
gcc-7 (GCC) 7.1.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ icc --version
icc (ICC) 18.0.1 20171018
Copyright (C) 1985-2017 Intel Corporation.  All rights reserved.

$ clang-4.0 --version
clang version 4.0.0 (tags/RELEASE_400/final)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /opt/llvm/4.0.0/bin
```

# Errors

## GCC

```
/home/jrhammon/Work/DOE/LLNL/RAJA/test/unit/omp-target/test-reductions.cpp: In member function ‘virtual void gtest_case_ReductionCorrectnessTestTargetOMP_::ReduceSum<gtest_TypeParam_>::TestBody()’:
/home/jrhammon/Work/DOE/LLNL/RAJA/test/unit/omp-target/test-reductions.cpp:143:9: error: ‘#pragma omp target data’ must contain at least one ‘map’ clause
 #pragma omp target data use_device_ptr(array)
```

Details: [gcc.log](https://github.com/LLNL/RAJA/files/1533416/gcc.log)

## Intel

```
[ 76%] Building CXX object test/unit/omp-target/CMakeFiles/test-omp-target-reductions.exe.dir/test-reductions.cpp.o
/home/jrhammon/Work/DOE/LLNL/RAJA/include/RAJA/policy/openmp/target_forall.hpp(67): error: variable "body" used in this "omp target" region is not bitwise copyable
      schedule(static, 1) map(to : body)
```

Details: [intel.log](https://github.com/LLNL/RAJA/files/1533417/intel.log)

## Clang

```
/home/jrhammon/Work/DOE/LLNL/RAJA/test/unit/omp-target/test-reductions.cpp:143:9: error: expected at least one map clause for '#pragma omp target data'
#pragma omp target data use_device_ptr(array)
```

Details: [clang.log](https://github.com/LLNL/RAJA/files/1533415/clang.log)

jeffhammond:
[...]

PR #489, opened 2018-05-30
Add OpenMP 4.5 support to `kernel`

davidbeckingsale:


trws:
This reminds me, icc 16 has openmp offload support, at least enough to 
compile the code and check syntax and so-forth.  We may need to add MPSS 
into the docker image, but that should be a much lighter lift than 
getting a GCC built.

davidbeckingsale:
@rhornung67 @trws @ajkunen  - do you think we should break the openmp-target stuff out into it's own directory under `policy`?

davidbeckingsale:
@trws -  it doesn't look like we have the equivalent of `implicit-declare-target` when using Intel, so I just hit the following:

`error: function "blah blah blah" called in "omp target" region must appear in prior "omp declare target" pragma`

@jeffhammond does such a compile flag exist?

davidbeckingsale:
My guess would be no. It basically lets the compiler pretend that everything has been declared as `target`.

What flags were you using to compile in #372 ?

rhornung67:
@davidbeckingsale moving openmp target stuff into its own subdirectories sounds like a good idea -- less confusion....

davidbeckingsale:
Okay great, I will do it in this PR.

jeffhammond:
@davidbeckingsale I don't remember but likely something like `-O3 -xHOST -qopenmp -qopenmp-offload=host`.

IIRC Intel doesn't insist on target declarations when the host and offload target are the same arch but I'll have to run tests to be sure.

davidbeckingsale:
Ah I can try that. Does it still generate a different codepath when the offload target is host, or just ignore the target stuff?

jeffhammond:
@davidbeckingsale I don't know the details but I know that I was able to call `sched_yield()` from an offload region with target=host but linking failed with target=gfx, and obviously this function isn't declared as target in the system header.

It sounds like I need to update my RAJA repo and redo all of these tests with the Intel 19 compiler.

rhornung67:
@davidbeckingsale the destination branch for this PR should be the 0.6.0rc3 branch, I believe, since that has most of the new kernel stuff.  It will also make it easier to try this out in the perf suite.

davidbeckingsale:
Okay, I can do that. Are we going to merge `develop` into `0.6.0` when we do the final release?

rhornung67:
@davidbeckingsale Yes, I think that is the plan, unless I'm missing something. I've been periodically merging develop into that branch to avoid big merge conflicts to resolve. But I haven't done it in a couple of weeks.

trws:
Clearly I’m coming to this late, but implicit declare target is a required default feature in 4.5 IIRC. If I remember incorrectly, then it is as of 5. A newer intel version (I would expect 18 at least) should certainly have it.


---
Sent from VMware Boxer

On June 1, 2018 at 11:08:26 AM PDT, Rich Hornung <notifications@github.com> wrote:

@davidbeckingsale<https://github.com/davidbeckingsale> Yes, I think that is the plan, unless I'm missing something. I've been periodically merging develop into that branch to avoid big merge conflicts to resolve. But I haven't done it in a couple of weeks.

davidbeckingsale:
@trws - per @jeffhammond's errors in #372 it looks like the implicit declare target isn't working there, and there is still the non-mappable type error for the lambda bodies. I'll play around with this some more locally.

rhornung67:
@davidbeckingsale and @trws this stuff is working with the clang-coral compilers we have here for all RAJA perf suite kernels.  My guess is that some compiler implementations are WIPs.

davidbeckingsale:
@rhornung67 - shall we get this merged, and then expand on testing via #543 ?

rhornung67:
@davidbeckingsale yes, let's get it in. I'm using it extensively in the Perf Suite and everything seems to be working. We can beef up the tests as part of the testing reorg.

PR #497, opened 2018-06-25
Examples - matrix transpose (WIP)

artv3:
New example in progress; testing travis builds. I may have found a corner case that breaks Kernel.


artv3:
Folks, the ex-shared-matrix-tranpose.cpp has the base version of the shared memory matrix transpose example. Ultimately we would like to have an abstraction such that developers do not have to do manual index calculations.

Ideally the code would like like this: 

     `` [=](int row, int col, seq_shmem_t &RAJA_SEQ_TILE) {
        RAJA_SEQ_TILE(row, col) = Aview(row, col);
      }``

I'll be working off this branch and introduce a third example which accomplishes the task and then merge the examples together.

PR #516, opened 2018-08-08
tuple forwarding and cat fixes

trws:
This is a set of fixes for tuple storage that specifically apply to forwarding and catting tuples.  Mainly reported by @DavidPoliakoff.


include/camp/camp.hpp
```diff
@@ -151,19 +151,23 @@ namespace test
 }
 #endif
 
+
+/**
+ * @brief Get the index of the first instance of T in L
```
davidbeckingsale: Woooo, documentation 😄

DavidPoliakoff:
Thanks for putting this together, be nice to have these merged so I can operate off of RAJA/develop again

include/camp/camp.hpp
trws: Yup.  Much more is needed, but it seemed like a place to start.  CAMP just got its release today, so I'll probably be taking a pass over all of that in the near future.

trws:
Ironically, given #514, the travis issue here is icc16 not generating an implicit copy constructor that should exist.

PR #521, opened 2018-08-27
add gcc8, update clang debug target, remove old clang and intel targets

trws:


trws:
I'm going to re-start the tests, the new gcc8 image is an ubuntu bionic base, completely new base image, that will include the gcc offload support on nvptx in case we want to try that.

davidbeckingsale:
Awesome!

trws:
Ouch, looks like the clang 6 updated address sanitizer is finding something suspicious in kernel collapse... will look into that shortly.

trws:
Ok, newly found "bug" fixed.  New gcc8 in, icc16 and clang4 out.  LGTM?

PR #541, opened 2018-11-20
Enable warnings as errors for GCC 4.9 & nvcc

davidbeckingsale:
This fixes RAJA-188 by enabling warnings as errors for GCC 4.9 and nvcc


davidbeckingsale:
Anyone got any ideas what's going on here: https://travis-ci.org/LLNL/RAJA/jobs/457692390#L593

`std::aligned_alloc` should be in `cstdlib` which is included in that header... 😞 

@trws @ajkunen @rhornung67 @artv3

trws:
`std::aligned_alloc` is from C++17, maybe should be `std::aligned_storage` (from C++11)?

davidbeckingsale:
What's weird is that we are using it, or at least attempting too, previously....

trws:
Maybe it's a gnu extension that happens to be available wherever we've been testing?  I'm not sure what's going on there, or maybe it's part of C11 but not c++11 or something?

davidbeckingsale:
Yeah. As far as I can see, it's working on our current develop/master builds. The only change in this branch is enabling warnings as errors. Maybe that tightens up some stuff...

trws:
Maybe it also switches the language default from the gnu++ variant to the c++ variant?

davidbeckingsale:
Hopefully this works... I've updated check_function_exists to use the namespace.

davidbeckingsale:
So this is all passing now, but it doesn't actually detect any of the functions (memalign, mm_alloc, or aligned_alloc) - I don't think the CMtests are set up correctly since they don't use any header files.

davidbeckingsale:
Can I get a review? @trws @rhornung67

PR #542, opened 2018-12-03
Rework of CUDA policies for RAJA::kernel

ajkunen:
This replaces a lot of the crud and cruft involved in the implementation of RAJA::kernel with CUDA.
The indexing scheme is more directly understandable by people familiar with CUDA, and it allows for much simpler/faster implementation.


artv3:
@ajkunen , what do the _loop policies do? For example:  using cuda_thread_x_loop = cuda_thread_xyz_loop<0, 1>; ?

ajkunen:
@artv3 they implement a block-stride loop... so you can have more iterations than threads... whereas the _direct policies are direct-mapped to threads.  If you are using shmem you probably want the direct-mapped.

examples/ltimes.cpp
```diff
@@ -50,8 +50,8 @@
  */
```
artv3: @ajkunen , I think #include<cmath> is missing in this file. I suspect that is causing ambiguity in the gcc6 build.

ajkunen: that was it, thanks!

ajkunen:
I think this is ready for some review

trws:
I feel like I've asked this before, but why support the loop variant on both and the direct variant only on threads?

ajkunen:
@trws It wasn't clear there was a use case for direct mapping blocks.  It could easily be added. 
The number of threads is fixed either by hardware, or by the launching methods, so it's easy to develop algorithms that depend on a fixed number of threads.  Direct mapping threads is then possible, and allows for things like thread-private values to be passed between lambdas.  There doesn't seem to be a benefit to direct mapping blocks, and the number of blocks you get is much more variable, especially if we eventually want guarantees like you get with cooperative groups: we may need to restrict the number of blocks artificially, so it may not be clear to the user how many blocks they can use.  This is all hidden with grid-stride looping.

If you have a use case in mind, I'm open to adding those policies.

trws:
The only real use-case I had in mind was the ability to disable the grid-stride loop.  It isn't frequently worthwhile, but every once in a while I run into a kernel that runs better without it.  Usually it's a load-balancing issue or something like that.  That said, I don't know that it's terribly important.  As you said, direct thread mapping is much more directly useful, where this would just be an occasional optimization option.

ajkunen:
I don't disagree with what you're saying.  Why don't we discuss this at our next developer's meeting?  My argument is assuming we add grid-sync or other grid-XX functions that don't exist yet.

We can always add direct block policies later, but it's harder to remove them if people start using them.  Could we agree to hold off on those for now?

PR #548, opened 2018-12-10
Removed Shmem and replace with LocalArray

ajkunen:
Excised the ShmemXXX stuff and replaced with LocalArray
Added permutation parameter to the StaticLayout and TypedStaticLayout
Update the examples and unit tests to use LocalArray
Added invert_permutation<> helper

Depends on PR #547


ajkunen:
I think this is ready for some scrutiny

examples/ltimes.cpp
```diff
@@ -312,19 +313,22 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
...
   constexpr size_t tile_g = 0;
 
+  using RAJA::statement::Param;
+
   using EXECPOL = 
     RAJA::KernelPolicy<
 
+      // Create memory tiles
+      statement::InitLocalMem<RAJA::cpu_tile_mem, RAJA::ParamList<0,1,2>,
+
```
artv3: Shouldn't this go after a set of loops?

ajkunen: not in this case, this is just sequential loops with some fixed array work buffers

PR #552, opened 2018-12-11
build and test on forks that don't have intel key

trws:
This is what caused the issue with PR #549, since forks don't have the encrypted key, they were having both DO_BUILD and DO_TEST set to no.


trws:
As a note, I put this in a branch so we could check on travis here to see if the environment is set correctly, as expected the intel build and test did not run, but all others did.

davidbeckingsale:
I think you mean in a fork?

trws:
Quite so.

PR #553, opened 2018-12-12
Remove RAJA::forallN

ajkunen:
Not really much to say.... except that it's gone


davidbeckingsale:
😢

ajkunen:
@davidbeckingsale Ok, it looks good now.

examples/waveShared.cpp
```diff
@@ -111,9 +111,19 @@ void innerLoop(const serialLoop &, int Ny, int Nx, Func &&innerLoop){
   }
 #endif
```
artv3: @ajkunen , can we delete this file?

ajkunen: @artv3 , yeah, those were your examples... I started converting one of them, but then realized they weren't even in the CMakeLists files anymore.
Do you want to remove them as an ancillary PR?

PR #554, opened 2018-12-13
Only set RAJA_LOADED where there actually is a PARENT_SCOPE

ajkunen:
This squashes an annoying warning you get during the CMake configure step


DavidPoliakoff:
This is really solid, nice. If @cyrush doesn't already know this pattern, I want to ping him, it seems like a good practice for BLT-style codes

ajkunen:
Oh Joy!

cyrush:
`RAJA_FULLY_LOADED` 🥇

PR #571, opened 2019-01-30
Update Dockerfile

artv3:
Builds and installs RAJA using the gcc8 compiler. Excellent for hands on activities.


Dockerfile
```diff
@@ -1,10 +1,16 @@
-FROM nvidia/cuda:8.0-devel-ubuntu16.04
+FROM nvidia/cuda:9.0-devel-ubuntu16.04
```
davidbeckingsale: I think you can build on top of our existing cuda image?

```suggestion
FROM rajaorg/compiler:nvcc9
```

Dockerfile
```diff
@@ -1,10 +1,16 @@
...
 MAINTAINER RAJA Development Team <raja-dev@llnl.gov>
 
 RUN apt-get update -y
 RUN apt-get install -y git cmake gdb
 
-RUN cd /opt/ && git clone https://github.com/LLNL/RAJA.git
+ADD https://cmake.org/files/v3.11/cmake-3.11.0-Linux-x86_64.sh /cmake-3.11.0-Linux-x86_64.sh
```
davidbeckingsale: This should be unnecessary if you are using our nvcc9 image

davidbeckingsale:
@artv3 - maybe add a comment in the Dockerfile about what the image is (e.g. builds and installs RAJA), we can get this merged.

Dockerfile
```diff
@@ -1,13 +1,12 @@
...
+COPY --chown=raja:raja . /home/raja/src
 
-RUN cd /opt/ && git clone https://github.com/LLNL/RAJA.git
+WORKDIR /home/raja/src
 
-WORKDIR /opt/RAJA
+RUN  mkdir build && cd build && cmake -DENABLE_CUDA=OFF ..
 
-RUN mkdir build && cd build && cmake -DENABLE_CUDA=ON ..
+RUN cd build && sudo make -j 3
```
davidbeckingsale: Don't we want to do `make install` too?

artv3: done.

Dockerfile
```diff
@@ -1,13 +1,16 @@
...
+WORKDIR /home/raja/src
 
-WORKDIR /opt/RAJA
+RUN  mkdir build && cd build && cmake -DENABLE_CUDA=OFF ..
 
-RUN mkdir build && cd build && cmake -DENABLE_CUDA=ON ..
+RUN cd build && sudo make -j 3 && sudo make install
 
-RUN cd build && make -j && make install
+CMD ["bash"]
```
davidbeckingsale: @artv3 - can you verify which directory this drops into? I'm assuming `~/`, but if not we might need to update the tutorial accordingly.

artv3: It drops you at /home/raja/src

davidbeckingsale: Okay, I think that works best for the tutorial right?

artv3: It might be good to change the src folder name to work-space or something like that.  
As is /home/raja/src contains the raja files and folders, one of them is being src.

davidbeckingsale: Sure, that would make sense.

artv3:
@davidbeckingsale , @rhornung67 .  I think I got all changes. Could I please get another review?

davidbeckingsale:
Nice job @artv3 👍

PR #578, opened 2019-02-20
Fixed for cuda performance

ajkunen:
Eliminated a bunch of cases that caused inadvertent conversions between 32 and 64 bit integers


DavidPoliakoff:
Any numbers yet? I'm curious.

PR #580, opened 2019-02-25
Remotes/origin/bugfix/chen59/omptargetfixes

rchen20:
omp_target header files were not merged & tested correctly in previous PR. Moved all OpenMP target to openmp_target directory. Removed an assert in omp_target reducer to play nicely with nvcc. This PR can only be compiled with clang-coral 18.08.08 at the moment.

### DIFF
--- a/include/RAJA/policy/openmp_target/reduce.hpp
+++ b/include/RAJA/policy/openmp_target/reduce.hpp
@@ -20,8 +20,6 @@
 
 #if defined(RAJA_ENABLE_TARGET_OPENMP)
 
-#include <cassert>
-
 #include <algorithm>
 
 #include <omp.h>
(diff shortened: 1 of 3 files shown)


include/RAJA/policy/openmp_target/reduce.hpp
```diff
@@ -196,7 +194,6 @@ struct TargetReduce
   //! apply reduction on device upon destruction
   ~TargetReduce()
   {
-    assert ( omp_get_num_teams() <= omp::MaxNumTeams );
```
davidbeckingsale: Why is this assertion removed? Does it still need to be enforced?

davidbeckingsale: We don't really have a policy on asserts, but that sounds like a good idea. If you can comment them out and leave a note about the XL issue that would be perfect.

rchen20: It's not really necessary at that point because we have performed checks on the number of teams in include/RAJA/policy/openmp_target/forall.hpp lines 54-69.

If you add the asserts back in, you'll see that clang coral cannot nvlink those asserts. The asserts used to work with XL, however, after ~mid-January XL builds started to hang during test-atomic. I can put the asserts back in as comments, until compiler issues are fixed in the future?

PR #594, opened 2019-03-26
Fixed TypedRangeSegment, TypedRangeStrideSegment

ajkunen:
Mostly fixes for defaults for difference_type, and remove accidental type conversions (like assuming Index_type)

Had to add a helper class to strip strongly typed indices down to their value type.
Reverted some of the changes to TypedRangeStrideSegment to fix correctness issues.

### DIFF
--- a/include/RAJA/index/IndexValue.hpp
+++ b/include/RAJA/index/IndexValue.hpp
@@ -338,6 +338,31 @@ constexpr RAJA_HOST_DEVICE RAJA_INLINE
   return val;
 }
 
+namespace internal{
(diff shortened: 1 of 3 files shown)


artv3:
@ajkunen , did you run this through the RAJA Perf Suite?

ajkunen:
@artv3 no, i didn't... I ran it through my code, and it was slightly faster.  I suppose I should be more thourough and try it in the perf suite

ajkunen:
There appears to be little to no effect on the perf suite, generally things seem about 1% faster, which is probably not statistically significant.

PR #598, opened 2019-04-02
Guards for cuda enabled RAJA

artv3:
By adding a few checks, this PR enables users to maintain one build of RAJA and be compatible with strict device code and CUDA code. This was motivated by an application which compiled packages which depended on RAJA with both nvcc and host compilers.

### DIFF
--- a/include/RAJA/index/ListSegment.hpp
+++ b/include/RAJA/index/ListSegment.hpp
@@ -38,7 +38,7 @@
 #include "RAJA/util/macros.hpp"
 #include "RAJA/util/types.hpp"
 
-#if defined(RAJA_ENABLE_CUDA)
+#if defined(__NVCC__) && defined(RAJA_ENABLE_CUDA)
 #include "RAJA/policy/cuda/raja_cudaerrchk.hpp"
 #else
 #define cudaErrchk(...)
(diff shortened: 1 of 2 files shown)


davidbeckingsale:
@artv3 - looks like this breaks the clang-cuda builds, I think you might need a different macro. See [this](https://llvm.org/docs/CompileCudaWithLLVM.html#detecting-clang-vs-nvcc-from-code).

include/RAJA/index/ListSegment.hpp
```diff
@@ -38,7 +38,7 @@
 #include "RAJA/util/macros.hpp"
 #include "RAJA/util/types.hpp"
 
-#if defined(RAJA_ENABLE_CUDA)
+#if defined(__NVCC__) && defined(RAJA_ENABLE_CUDA)
```
davidbeckingsale: Probably want `__CUDACC__` rather than `__NVCC__`

artv3: Ah even better! Thanks.

davidbeckingsale: Your solution looks good too.

artv3: Not sure why, but the  ``#if defined(__CUDACC__) && defined(RAJA_ENABLE_CUDA)``  combination didn't work. I reverted to the previous commit.

davidbeckingsale: Works for me.

PR #599, opened 2019-04-08
 Kernel lambda statement API

artv3:
This is a work in progress, but with a few modifications to kernel internals we can remove the requirement that all lambdas should have the same number of arguments. kernel-api.cpp illustrates the existing and potential API.


davidbeckingsale:
Cool!

include/RAJA/pattern/kernel/Lambda.hpp
```diff
@@ -64,6 +64,18 @@ struct Lambda : internal::Statement<camp::nil> {
   const static camp::idx_t loop_body_index = BodyIdx;
 };
 
+//Base case
```
ajkunen: This is really great work.

Here are some suggestions:

It would be great if tLambda got merged into Lambda...
template<camp::idx_t BodyIndex, typename ... Args>

Where the current Lambda implementation is a specialization with zero Args.

It would also be great if you could say
Lambda<0, Seg<0>, Param<0>, Seg<1>>
Which would pass to a lambda the following arguments: value of segment 0, parameter 0, value of segment 1.
(Clearly we would need to create a Seg<camp::idx_t> type)

Or,
Lambda<0, Seg<0>, Offset<0>>
Which would pass the arguments: the value of segment 0, the offset of segment 0.
This would obviate the need for the ForICount, and eliminate the need for defining a param to hold the ICount value since it would just be passed in by argument.

include/RAJA/pattern/kernel/internal.hpp
```diff
@@ -258,10 +258,9 @@ RAJA_HOST_DEVICE RAJA_INLINE void invoke_lambda_expanded(
     camp::idx_seq<ParamIdx...> const &,
     Data &&data)
 {
-  camp::get<LoopIndex>(
-      data.bodies)((camp::get<OffsetIdx>(data.segment_tuple)
-                        .begin()[camp::get<OffsetIdx>(data.offset_tuple)])...,
-                   camp::get<ParamIdx>(data.param_tuple)...);
+  camp::get<LoopIndex>(data.bodies)
+    ((camp::get<OffsetIdx>(data.segment_tuple).begin()[camp::get<OffsetIdx>(data.offset_tuple)])...,
```
artv3: @ajkunen , by OffsetIdx this corresponds to the iteration space index right?  For example it can be a RangeSegment or ListSegment.

artv3: I used SegIdx in other places, I can make it consistent though.

ajkunen: That's right, it's an offset into the segment tuple... So it can be anything segment type

ajkunen: SegIdx would be more descriptive, feel free to change OffsetIdx to SegIdx to make things more consistent.

examples/kernel-api.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+
+  std::cout<<"----------------------------------------------------\n \n"<<std::endl;
+  std::cout<<"Testing new kernel lambda statement types \n"<<std::endl;
+
+  //Lambda statement format : lambda idx, segment indices, parameter indices (to be tested...)
+  using NEW_POLICY =
+    RAJA::KernelPolicy<
+      RAJA::statement::For<0,RAJA::loop_exec,
+        RAJA::statement::tLambda<0, camp::idx_seq<0>, camp::idx_seq<>>
```
rhornung67: I assume that 'tLambda' is a temporary name for this to avoid stepping on the 'lambda' statement type for now.

examples/kernel-api.cpp
```diff
@@ -0,0 +1,179 @@
...
+      >,
+      RAJA::statement::For<1,RAJA::loop_exec,
+        RAJA::statement::tLambda<1, camp::idx_seq<1>, camp::idx_seq<>>
```
rhornung67: We will want this to be a variadic template so one only provides parameters for the lambda args to be used, correct? Also, we will want to have clearly-named types for the parameters as we discussed; e.g., Seg<#>, SegList<#>, Param<#>, etc. What do you think of using SegList and ParamList only, which would take 1 or more arguments?

examples/kernel-api.cpp
```diff
@@ -0,0 +1,179 @@
...
+        match = false;
+      }
+    }
+  }
+  if ( match ) {
+    std::cout << "\n\t result -- PASS\n";
+  } else {
+    std::cout << "\n\t result -- FAIL\n";
+  }
+};
```
rhornung67: Have you tried this using strongly-typed indices (i.e., TypedRangeSegment) and associated lambda args? The compiler should be able to help with correctness.

include/RAJA/pattern/kernel/Lambda.hpp
rhornung67: I agree with @ajkunen 's first comment. I assume you would want to do that too, but wanted to get some early feedback before doing more surgery to the code.

@ajkunen second comment is similar to the one I made above.

We should probably discuss @ajkunen 's third comment with others to see what they think. While the API may be leaner, it may be too opaque for some folks.

examples/kernel-api.cpp
```diff
@@ -0,0 +1,179 @@
...
+          RAJA::statement::For<2, RAJA::loop_exec,
+            RAJA::statement::tLambda<1, camp::idx_seq<0,1,2>, camp::idx_seq<0>> // inner loop: dot += ...
+          >,
+          RAJA::statement::tLambda<2, camp::idx_seq<0,1>, camp::idx_seq<0>>   // set C(row, col) = dot
+        >
+      >
+    >;
+
+  RAJA::kernel_param<EXEC_POL6>(
+    RAJA::make_tuple(RAJA::RangeSegment(0, N),
```
rhornung67: We also talked about making RAJA::make_seg_tuple, and RAJA::make_param_tuple methods that generate distinct types. While (slightly) more verbose, I think it would make the code easier to understand. @artv3, @ajkunen thoughts?

ajkunen: Why not RAJA::make_tuple_of_segment_looking_thingies(...) and RAJA::make_tuple_of_parameter_looking_thingies(...)?

rhornung67: :-B

artv3: What about having a single tuple consisting of segments and parameters?

rhornung67: Personally, I prefer two distinct tuples for code readability and less disruption to current usage. Also, going with one tuple will require more rework of RAJA internals.

examples/kernel-api.cpp
artv3: Right, I think starting with Seg<>, Param<> would be a good start and then we can bootstrap to the list variants.

examples/kernel-api.cpp
artv3: I can try that next, with the second iteration of the API.

include/RAJA/pattern/kernel/Lambda.hpp
artv3: With next iteration, I've made Lambda variadic so the new variation is just an extension. I'm currently working on getting something like Lambda<0, Seg<0>, Param<0>, Seg<1>> to work. I think I'm not too far.

examples/kernel-api.cpp
rhornung67: 👍

examples/kernel-api.cpp
rhornung67: 👍

examples/kernel-api.cpp
```diff
@@ -0,0 +1,133 @@
...
+  using RAJA::statement::Param;
+  using POLICY_V2 =
+    RAJA::KernelPolicy<
+    RAJA::statement::For<0,RAJA::loop_exec,
+                         RAJA::statement::Lambda<0, Param<0>, Seg<0>>
+                         >,
+    RAJA::statement::For<1,RAJA::loop_exec,
+                         RAJA::statement::Lambda<1, Seg<1>>
+    >
+    >;
```
artv3: @ajkunen , @rhornung67 I think this bring us closer to the described API. Here I can choose to set the Param<0> to be the first argument in Lambda 0.

include/RAJA/pattern/kernel/Lambda.hpp
```diff
@@ -79,6 +91,89 @@ struct StatementExecutor<statement::Lambda<LoopIndex>> {
...
+
+//Iteration 1
+template <camp::idx_t LoopIndex,typename... Args>
+struct StatementExecutor<statement::Lambda<LoopIndex, Args...>> {
+
+  template <typename Data>
+  static RAJA_INLINE void exec(Data &&data)
+  {
+    const int size = sizeof...(Args); //Total number of arguments
+    auto myTuple = inspector<size, Args...>::make_args(camp::make_tuple(),data);
```
artv3: In a nutshell, I'm doing a gather of the desired arguments and their specified order, storing them in a tuple then feeding them into the lambda. I'm wondering if there is a cleaner way to do this...

ajkunen: it might be cleaner to do something that looked like:

auto myTuple = camp::make_tuple( inspector<Args>(data)::extract_arg() ... );

Where you use the helper class (inspector) to extract each argument as a single value... this eliminates the need for peeling off each argument and using tuple_cat_pair.
It would also make the implementation a lot simpler.

ajkunen: sorry, "data" above should be the argument to extract_arg, like this:

`auto myTuple = camp::make_tuple( inspector<Args>::extract_arg(data) ... );`

include/RAJA/pattern/kernel/Lambda.hpp
```diff
@@ -59,11 +59,23 @@ namespace statement
  * RAJA::kernel<exec_pol>(make_tuple{s0, s1, s2}, lambda0, lambda1);
  *
  */
-template <camp::idx_t BodyIdx>
+template <camp::idx_t BodyIdx, typename... Args >
```
ajkunen: Nice!

examples/kernel-api.cpp
```diff
@@ -0,0 +1,133 @@
...
+  std::cout<<"\n \n----------------------------------------------------"<<std::endl;
+  std::cout<<"Kernel API Iteration 2"<<std::endl;
+
+  using RAJA::statement::Seg;
+  using RAJA::statement::Param;
+  using POLICY_V2 =
+    RAJA::KernelPolicy<
+    RAJA::statement::For<0,RAJA::loop_exec,
+                         RAJA::statement::Lambda<0, Param<0>, Seg<0>>
+                         >,
```
ajkunen: This looks great!

include/RAJA/pattern/kernel/Lambda.hpp
artv3: Sweet! It works. I'll do some clean up on this branch next.

include/RAJA/pattern/kernel/Lambda.hpp
```diff
@@ -79,6 +91,63 @@ struct StatementExecutor<statement::Lambda<LoopIndex>> {
...
+};
+
+//Iteration 1
+template <camp::idx_t LoopIndex,typename... Args>
+struct StatementExecutor<statement::Lambda<LoopIndex, Args...>> {
+
+  template <typename Data>
+  static RAJA_INLINE void exec(Data &&data)
+  {
+    const int size = sizeof...(Args); //Total number of arguments
```
ajkunen: nice work!

include/RAJA/pattern/kernel/Conditional.hpp
```diff
@@ -69,6 +69,27 @@ struct Param : public internal::ParamBase {
...
 };
 
+/*!
+ * An expression that returns the value of the specified RAJA::kernel
+ * segment.
+ *
+ * This allows run-time values to affect the control logic within
+ * RAJA::kernel execution policies.
+ */
+template <camp::idx_t SegId>
```
ajkunen: Seg (and Param) should get moved out of Conditional.hpp and into their own files...  Seg.hpp and Param.hpp

artv3: Okay I can work on adding the Seglist, Paramlist methods in there too.

ajkunen: Also, Offset should be easy to implement now that you have Seg implemented!

artv3: Should the offset struct be in its own file? or should it be in Seg.hpp?

examples/kernel-api.cpp
rhornung67: 👍

include/RAJA/pattern/kernel/Lambda.hpp
```diff
@@ -86,6 +87,17 @@ template<typename Head, typename...Tail>
 struct extractor
 {};
 
+template<camp::idx_t id, typename...Tail>
+struct extractor<RAJA::statement::OffSet<id>, Tail...>
+{
+  template<typename Data>
+  static auto extract_arg(Data &&data) ->
+    decltype(camp::get<id>(data.offset_tuple))
+  {
```
artv3: @ajkunen , I added support for offsets. Just wanted to double check if this is what you had in mind. If so I may just need some unit test now.

artv3:
The CPU version seems to work. The CUDA variant is work in progress.

include/RAJA/pattern/kernel/ArgHelper.hpp
```diff
@@ -0,0 +1,293 @@
...
+// LLNL-CODE-689114
+//
+// All rights reserved.
+//
+// This file is part of RAJA.
+//
+// For details about use and distribution, please read RAJA/LICENSE.
+//
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
```
artv3: @rhornung67 , @ajkunen , I have this working now. The downside of my current implementation is that there are quite a number of helpers needed. I'll try to shrink this down, please feel free to comment.

include/RAJA/pattern/kernel/ArgHelper.hpp
```diff
@@ -0,0 +1,293 @@
...
+  template<typename Data>
+  RAJA_HOST_DEVICE
+  static auto extract_arg(Data &&data)->
+    decltype(camp::get<id>(data.param_tuple))
+  {
+    return camp::get<id>(data.param_tuple);
+  }
+};
+
+//Helper to concatante list [Move to camp?]
```
ajkunen: camp::list is a type, and has no data.... so catList should be implemented using types not functions that return objects.

That is, the usage should look like:`catList<...>::type`
instead of  `catList<...>::makeList`

This will make implementation a lot cleaner.

Also, it may make sense to rename catList to catListHelper (maybe in an internal namespace) and then define:
`template<typename A, typename B>
using catList = typename internal::catListHelper<A,B>::type`

include/RAJA/pattern/kernel/ArgHelper.hpp
```diff
@@ -0,0 +1,293 @@
...
+  RAJA_HOST_DEVICE
+  static auto makeList(camp::list<itemsA...> const &, camp::list<> const &) ->
+    camp::list<itemsA...>
+  {
+    return camp::list<itemsA...> {};
+  }
+
+};
+
+template<typename Arg>
```
ajkunen: Same comment for listMaker as for catList... convert to type based calculation

include/RAJA/pattern/kernel/ArgHelper.hpp
```diff
@@ -0,0 +1,293 @@
...
+
+    return camp::list<Param<id>>{};
+  }
+};
+
+//Helper structs to parse through template arguments
+//Takes a pre-processing step convert SegList, ParamList
+//into a list of Seg and Param's
+template<typename List>
+struct parser{};
```
ajkunen: Same comment again: convert to type based calculation

include/RAJA/pattern/kernel/ArgHelper.hpp
```diff
@@ -0,0 +1,293 @@
...
+
+template<typename ...Args>
+struct call_extractor<camp::list<Args...>>
+{
+  template<typename Data>
+  RAJA_HOST_DEVICE
+  static auto make_tuple(Data &&data)
+    -> decltype(camp::make_tuple(extractor<Args>::extract_arg(data) ...))
+  {
+    return camp::make_tuple(extractor<Args>::extract_arg(data) ...);
```
ajkunen: You could probably move call_extractor into the invoke_lambda_with_args function.  This is probably an unnecessary class.

include/RAJA/pattern/kernel/Lambda.hpp
```diff
@@ -79,6 +80,32 @@ struct StatementExecutor<statement::Lambda<LoopIndex>> {
...
+
+    //Convert SegList, ParamList into Seg, Param types, and store in a list
+    auto targList = parser<camp::list<Args...>>::checkArgs();
+
+    //Create a tuple with the appropriate lambda arguments
+    auto argTuple = call_extractor<decltype(targList)>::make_tuple(data);
+
+    //Invoke the lambda with custom arguments
+    const int tuple_size = camp::tuple_size<decltype(argTuple)>::value;
+    invoke_lambda_with_args<LoopIndex>(std::forward<Data>(data),
```
ajkunen: it would be nice if this call looked like:
`invoke_lambda_with_args<LambdaIndex, targList>(data)`

the invoke function should be able to pull out the appropriate information from the LoopData object that's passed in.

if invoke_lambda_width_args needs the make_idx_seq_t, then it should call a invoke_lambda_with_args_expanded with the appropriate index sequence... that way the caller doesn't have to know the special sauce that's needed to call it.

this would cleanup the call sites of invoke_lambda_with_args, and make it easier to maintain in the future

include/RAJA/pattern/kernel/Lambda.hpp
```diff
@@ -79,6 +80,32 @@ struct StatementExecutor<statement::Lambda<LoopIndex>> {
   }
 };
 
+
+/*!
+ * A RAJA::kernel statement that invokes a lambda function
+ * with user specified arguments.
+ */
+template <camp::idx_t LoopIndex,typename... Args>
```
ajkunen: LoopIndex should be LambdaIndex,  since it's selecting a Lambda

include/RAJA/pattern/kernel/OffSet.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+/*!
+ * An expression that returns the value of the specified RAJA::kernel
+ * OffSet.
+ *
+ * This allows run-time values to affect the control logic within
+ * RAJA::kernel execution policies.
+ */
+template<camp::idx_t offId>
+struct OffSet : public internal::OffSetBase
```
ajkunen: Nice!

include/RAJA/pattern/kernel/internal.hpp
```diff
@@ -278,6 +281,21 @@ RAJA_INLINE RAJA_HOST_DEVICE void invoke_lambda(Data &&data)
       std::forward<Data>(data));
 }
 
+//Lambda with custom arguments
+RAJA_SUPPRESS_HD_WARN
+template<camp::idx_t LoopIndex, typename Data, typename T, camp::idx_t... Args>
+RAJA_INLINE RAJA_HOST_DEVICE void invoke_custom_lambda(Data &&data,T myTuple)
+{
+  camp::get<LoopIndex>(data.bodies)(camp::get<Args>(myTuple)...);
```
ajkunen: I think if you merge in the call_extractor logic here, you'll never have to build the tuple in the first place... the argument would just drop into the expansion

include/RAJA/util/LocalArray.hpp
```diff
@@ -41,7 +41,7 @@ namespace RAJA
 
 
 template<camp::idx_t ... Sizes>
-using ParamList = camp::idx_seq<Sizes...>;
+using InitList = camp::idx_seq<Sizes...>;
```
ajkunen: Can you explain the need for InitList?
The LocalArray's are indeed Params, so wouldn't ParamList be appropriate for InitLocalMem?
Or was there some disambiguation reason?

artv3: There was some ambiguity issues, I developed a Paramlist for the new kernel interface but its different than the existing version. Thanks for the comments! I'll address these and do some clean up.

examples/tut_matrix-transpose-local-array.cpp
```diff
@@ -207,17 +214,17 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
       RAJA::statement::Tile<1, RAJA::statement::tile_fixed<TILE_DIM>, RAJA::loop_exec,
         RAJA::statement::Tile<0, RAJA::statement::tile_fixed<TILE_DIM>, RAJA::loop_exec,
 
-          RAJA::statement::InitLocalMem<RAJA::cpu_tile_mem, RAJA::ParamList<2>,
-
-          RAJA::statement::ForICount<1, RAJA::statement::Param<0>, RAJA::loop_exec,
-            RAJA::statement::ForICount<0, RAJA::statement::Param<1>, RAJA::loop_exec,
-              RAJA::statement::Lambda<0>
+          RAJA::statement::InitLocalMem<RAJA::cpu_tile_mem, RAJA::InitList<0>,
```
rhornung67: Do we think it's confusing for users if we have too many different types for these things?  For example, I need to know when to use InitList vs. ParamList vs. SegList, etc.  It would probably be a good idea to run some of this by some users to get their feedback.

examples/tut_matrix-transpose-local-array.cpp
```diff
@@ -261,17 +268,17 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
       RAJA::statement::Tile<0, RAJA::statement::tile_fixed<TILE_DIM>, RAJA::loop_exec,
         // This statement will initalize local array memory inside a
         // kernel. The cpu_tile_mem policy specifies that memory should be
-        // allocated on the stack. The entries in the RAJA::ParamList
+        // allocated on the stack. The entries in the RAJA::InitList
         // identify RAJA local arrays in the parameter tuple to intialize.
-        RAJA::statement::InitLocalMem<RAJA::cpu_tile_mem, RAJA::ParamList<2>,
+        RAJA::statement::InitLocalMem<RAJA::cpu_tile_mem, RAJA::InitList<0>,
```
rhornung67: We're going to have to go through the documentation carefully to make sure the usage of these things is clear.

examples/tut_matrix-transpose-local-array.cpp
```diff
@@ -261,17 +268,17 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
...
           // (1) Execution policies for the first set of inner
           // loops. These loops copy data from the global matrices
           // to the local tile.
           //
-          RAJA::statement::ForICount<1, RAJA::statement::Param<0>, RAJA::loop_exec,
-            RAJA::statement::ForICount<0, RAJA::statement::Param<1>, RAJA::loop_exec,
-                                       RAJA::statement::Lambda<0>
+          RAJA::statement::For<1, RAJA::loop_exec,
+            RAJA::statement::For<0, RAJA::loop_exec,
+              RAJA::statement::Lambda<0, Seg<0>, Seg<1>, OffSet<0>, OffSet<1>, Param<0> >
```
rhornung67: Same comment as above...

examples/tut_matrix-transpose-local-array.cpp
```diff
@@ -421,9 +428,9 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
             //     swapped! This enables us to swap which
             //     index has unit stride.
             //
-            RAJA::statement::ForICount<0, RAJA::statement::Param<1>, RAJA::cuda_thread_y_direct,
-              RAJA::statement::ForICount<1, RAJA::statement::Param<0>, RAJA::cuda_thread_x_direct,
-                                            RAJA::statement::Lambda<1>
+            RAJA::statement::For<0, RAJA::cuda_thread_y_direct,
+              RAJA::statement::For<1, RAJA::cuda_thread_x_direct,
+               RAJA::statement::Lambda<1, Seg<0>, Seg<1>, OffSet<0>, OffSet<1>, Param<0> >
```
rhornung67: Do we have any SegList, etc. examples?

include/RAJA/pattern/kernel/ArgHelper.hpp
```diff
@@ -0,0 +1,293 @@
...
+#include "RAJA/pattern/kernel/Param.hpp"
+#include "RAJA/pattern/kernel/Seg.hpp"
+
+namespace RAJA
+{
+
+namespace internal
+{
+
+using RAJA::statement::SegList;
```
rhornung67: This collection of types is what we need users to weigh in on....

include/RAJA/pattern/kernel/OffSet.hpp
```diff
@@ -0,0 +1,65 @@
+/*!
```
rhornung67: These are the 'helpers' you mentioned above, correct??  I know why you did it this way, but the implementations are all the same. Have you thought about making one implementation that is templated on the "index" type?  For example, instead of ParamList<0, 1>, what about something like List<Param>(0, 1)? List is a bad name here, but we could come up with something better.  @ajkunen what do you think about this?

test/unit/test-new-kernel-api.cpp
```diff
@@ -0,0 +1,605 @@
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
```
rhornung67: Would this file and tests inside it be named better if it were something like test-kernel-lambda-args.cpp?? In other words break out the tests that do contortions of the lambda arguments into a separate test source file.

include/RAJA/pattern/kernel/OffSet.hpp
artv3: I think we need to have the variants to identify from which tuple the data should come from, Segment tuple vs Param tuple.

ajkunen:
@Ram3sh can you take a look at this PR?  It has the Lambda rework that we had talked about a while ago.   It's still being cleaned up, but the basic interface is there.

Any feedback would be welcome.

examples/tut_matrix-transpose-local-array.cpp
```diff
@@ -207,17 +214,17 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
...
+              RAJA::statement::Lambda<0, Seg<0>, Seg<1>, OffSet<0>, OffSet<1>, Param<0> >
             >
           >,
 
-          RAJA::statement::ForICount<0, RAJA::statement::Param<1>, RAJA::loop_exec,
-            RAJA::statement::ForICount<1, RAJA::statement::Param<0>, RAJA::loop_exec,
-              RAJA::statement::Lambda<1>
+          RAJA::statement::For<0, RAJA::loop_exec,
+            RAJA::statement::For<1, RAJA::loop_exec,
+              RAJA::statement::Lambda<1, Seg<0>, Seg<1>, OffSet<0>, OffSet<1>, Param<0> >
```
artv3: @Ram3sh , I would suggest looking here. We can also combine multiple Seg/Offset/Param to make *Lists . I can add an example in a bit.

Ram3sh: Looking at the example, I am not sure what Seg<X>,Offset<Y> and Param<0> are supposed to do.

artv3: @Ram3sh , I'll put together a more detailed example. I'll send it over once its ready.

include/RAJA/pattern/kernel/ArgHelper.hpp
artv3: Thanks! That is much better!

examples/kernel-api.cpp
```diff
@@ -0,0 +1,432 @@
...
+        RAJA::statement::Tile<0, RAJA::statement::tile_fixed<TILE_DIM>, RAJA::loop_exec,
+
+          //InitList identifies memory within the tuple which needs to be intialized
+          RAJA::statement::InitLocalMem<RAJA::cpu_tile_mem, RAJA::InitList<0>,
+
+          RAJA::statement::For<1, RAJA::loop_exec,
+            RAJA::statement::For<0, RAJA::loop_exec,
+              //The additional types passed into the lambda statement
+              //are used to specify lambda arguments.
+             RAJA::statement::Lambda<0, LambdaArgs<seg_t, 0, 1>, LambdaArgs<offset_t, 0,1>, LambdaArgs<param_t, 0> >
```
artv3: @ajkunen @rhornung67 , I have the alternative container working now for lambda arguments. We can change the name of the container, but functionally its doing the job. I was wondering if we liked this option? It does reduce some of the internals.

ajkunen: Great.  I think hiding LambdaArgs from the user would be the best.  This allows the internals to be templated, but present the user with a cleaner interface.  All you need to do is define some aliases like:
`template<camp::idx_t ... args> using Segs = LambdaArgs<seg_t, args...>;`

rhornung67: @artv3 I agree with @ajkunen about hiding LambdaArgs from the user and using aliases as he described.

artv3: 👍

include/RAJA/pattern/kernel/ArgHelper.hpp
```diff
@@ -0,0 +1,179 @@
...
+{};
+
+template<typename...itemsA, typename...itemsB>
+struct merge_list<camp::list<itemsA...>, camp::list<itemsB...>>
+{
+  using type = typename camp::list<itemsA...,itemsB...>;
+};
+
+template<typename...itemsA>
+struct merge_list<camp::list<itemsA...>, camp::list<>>
```
willkill07: Is this overload necessary? itemsB in the prior specialization could be zero-or-more

include/RAJA/pattern/kernel/ArgHelper.hpp
```diff
@@ -0,0 +1,179 @@
...
+struct listMaker
+{
+  using type = typename camp::list<>;
+};
+  
+//Convert LambdaArgs<T, 1, 2, 3> - > camp::list<LambdaArgs<T, 1>, LambdaArgs<T, 2>, LambdaArgs<T, 3> >
+template<typename T, camp::idx_t head, camp::idx_t... tail>
+struct listMaker<LambdaArgs<T, head, tail...>>
+{
+  using type = typename merge_list<camp::list<LambdaArgs<T, head>>,
```
willkill07: The additional types/expansions generated will make compilers cry. The following should make them happier:

```cpp
template <typename T, camp::idx_t... elems>
struct listMaker<LambdaArgs<T, elems...>>
{
  using type = camp::list<LambdaArgs<T, elems>...>;
};
```

Note: this also removes the need for the specialization below

include/RAJA/pattern/kernel/ArgHelper.hpp
```diff
@@ -0,0 +1,179 @@
...
+template<typename List>
+struct parser{};
+
+template<>
+struct parser<camp::list<>>
+{
+  using type = camp::list<>;
+};
+
+template <typename Head, typename... Tail>
```
willkill07: Similar change as above for parser... we should really consider adding a `camp::map` for `camp::list`... I'll probably work on that

```cpp
template <typename... Elems>
struct parser<camp::list<Elems...>>
{
  using type = camp::list<typename listMaker<Elems>::type...>;
};
```

include/RAJA/pattern/kernel/LambdaArgs.hpp
```diff
@@ -0,0 +1,86 @@
...
+template<camp::idx_t ... Args>
+struct LambdaArgs<param_t, Args...>
+{
+};
+
+template<camp::idx_t ... args> 
+using Segs = LambdaArgs<seg_t, args...>;
+
+template<camp::idx_t ... args> 
+using OffSets = LambdaArgs<offset_t, args...>;
```
willkill07: Should be Offsets instead of OffSets (nitpick/naming)

include/RAJA/pattern/kernel/ArgHelper.hpp
artv3: Thanks Will! Fixed

include/RAJA/pattern/kernel/ArgHelper.hpp
artv3: I don't think we can do the same for the parser since we have the merge list step.

include/RAJA/pattern/kernel/OffSet.hpp
artv3: Fixed with latest iteration.

test/unit/test-new-kernel-api.cpp
artv3: Yup, I just renamed the file. The new source file just looks at testing the different API.

include/RAJA/pattern/kernel/internal.hpp
artv3: Here I got a bit stuck. I wasn't sure how to further merge the code together.

artv3:
@ajkunen @rhornung67 , I think the code is ready to review. I just need to update the docs now. 
In the matrix multiply and matrix transpose with local array examples I added additional examples which use the new API.

include/RAJA/pattern/kernel/ArgHelper.hpp
trws: Would `camp::transform` and `camp::extend`, essentially map and concat, not work for this?

include/RAJA/pattern/kernel/ArgHelper.hpp
trws: Maybe we should chat about some of this, I have a feeling there are camp components that could slim this down quite a bit.  Hazard of the fact I haven't documented camp well, need to make some time for that.

include/RAJA/pattern/kernel/LambdaArgs.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+struct offset_t
+{};
+
+template<typename T, camp::idx_t ...>
+struct LambdaArgs
+{
+};
+
+template<camp::idx_t ... Args>
```
trws: Why the explicit LambdaArgs instantiations? Aren't these all equivalent to the generic one?

include/RAJA/pattern/kernel/internal.hpp
trws: Here too, there's nothing wrong with this or call_extractor, but we might be able to make it a bit easier to follow.

include/RAJA/pattern/kernel/ArgHelper.hpp
artv3: @trws , should I send an invite for this week? Or feel free to prose something via email.

trws:
Sounds good to me, maybe tomorrow afternoon? I have a lunch thing that isn't on my calendar, but the afternoon is open.

artv3:
Hi all, I met with @trws  earlier today. The internals are now much cleaner and simpler. Just need to update the sphinx docs.

artv3:
@rhornung67 , I think this PR is ready for review. I extended the docs to discuss the new kernel statement, I can also extend the docs more if more discussion is needed. We will also need a follow up PR to fix the line numbers for other examples, I wasn't sure if fixing it here would be the right place.

rhornung67:
@artv3 Please go ahead and fix all the broken the line numbers in the code examples in this PR. Thanks.  I'll review this PR later today.

docs/sphinx/user_guide/feature/loop_basic.rst
```diff
@@ -174,16 +174,40 @@ Here, the innermost type in the kernel policy is a
...
+                                RAJA::statement::Lambda<0, RAJA::statement::Segs<N,...,0>>
+                              >
+                            ...
+                          > 
+                        >;
+
+The template parameter ``RAJA::statement::Segs`` is used to identify elements from the
+segment tuple to be used as arguments for a lambda. RAJA offers other statements
+such as ``Offsets``, and ``Params`` to identify offsets and parameters in segments and 
+param tuples respectively to be used as lambda arguments.
```
rhornung67: @artv3 please put a reference here to the discussion of the example that shows how to use the offsets and params statements.  Do we have a code example of the offsets usage?

artv3: Sure thing! The matrix transpose example uses the offsets.

trws:
Looks good overall.

trws:
One comment here, the number of line-number changes required in the docs is pretty shocking.  Maybe next time something like this comes along we can switch to `start-after` and `end-before` instead of `lines` for these?

rhornung67:
@trws agreed. We didn't think ahead. I'll address this soon.

ajkunen:
I approved, but I also agree with @trws statement about line number referencing

artv3:
👍 Should we merge this now and save the line number referencing for the next PR?

trws:
Yeah, lets not delay the feature.  I'd rather have people trying it than waiting, and the docs are fine as they are it just shouldn't hurt so much to update them.  Thanks for the good work @artv3!

PR #609, opened 2019-05-30
Update getting_started.rst

metalberkeley:
I added a small section for compilation of RAJA with both CUDA and MPI enabled.

### DIFF
--- a/docs/sphinx/user_guide/getting_started.rst
+++ b/docs/sphinx/user_guide/getting_started.rst
@@ -89,34 +89,18 @@ RAJA uses CMake to configure a build. A basic configuration looks like::
   $ cmake -DCMAKE_INSTALL_PREFIX=/path/to/install ../
 
 
-MPI+CUDA Compilation
----------------------
+MPI+CUDA Compilation with ``nvcc``
+-----------------------------------
 
-When trying to compile RAJA with MPI and CUDA make sure that both the CMAKE 
-flags of RAJA and BLT are in sync. The BLT cmake variables are easily 
-recognizable because they have the ``BLT_`` prefix. This is not always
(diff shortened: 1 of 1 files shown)


docs/sphinx/user_guide/getting_started.rst
```diff
@@ -88,6 +88,33 @@ RAJA uses CMake to configure a build. A basic configuration looks like::
...
+When trying to compile RAJA with MPI and CUDA make sure that both the CMAKE 
+flags of RAJA and BLT are in sync. The BLT cmake variables are easily 
+recognizable because they have the ``BLT_`` prefix. 
+Next, you want to change the target architecture variables to the compute 
+capability of your hardware, in order for the ``nvcc`` compiler to generate 
+GPU code native for your GPU hardware. 
+`Click here <https://developer.nvidia.com/cuda-gpus>`_ to check which is the
+right version for your card. For example, if you have a Geforce GTX 1080, 
+its compute capability is 6.1. The cmake options to change in RAJA are 
+``BLT_CLANG_CUDA_ARCH`` and ``CUDA_ARCH``. Both should be set to ``sm61``.
```
davidbeckingsale: Why would both `BLT_CLANG_CUDA_ARCH` and `CUDA_ARCH` need to be set? BLT doesn't add any arch flag to `nvcc`, since it's not a required option. The `BLT_CLANG_CUDA_ARCH` option exists because when compiling CUDA code with clang, you _have_ to specify it.

metalberkeley: Got it. I saw that they are both generated when I run the cmake config. I did not look into whether that was always used. I just assumed that it would be a good practice to have them both in sync to avoid potential conflicts. But hey, I can change this addition to incorporate your comment! :)

PR #615, opened 2019-06-10
Error when trying to compile CUDA-enabled RAJA without CUDA

davidbeckingsale:


davidbeckingsale:
@gzagaris - can you give this a try?

davidbeckingsale:
@rhornung67 - this is the PR for the issue we discussed in the meeting today.

davidbeckingsale:
After the email thread a couple of weeks ago, I think there is a little more we need to do here.

PR #622, opened 2019-06-20
Update RAJA.hpp

corbett5:
This way you can include `RAJA.hpp` in translation units that aren't compiled with nvcc.

### DIFF
--- a/include/RAJA/RAJA.hpp
+++ b/include/RAJA/RAJA.hpp
@@ -62,7 +62,7 @@
 #include "RAJA/policy/tbb.hpp"
 #endif
 
-#if defined(RAJA_ENABLE_CUDA)
+#if defined(RAJA_ENABLE_CUDA) && defined(__CUDACC__)
 #include "RAJA/policy/cuda.hpp"
 #endif
 


include/RAJA/RAJA.hpp
```diff
@@ -62,7 +62,7 @@
 #include "RAJA/policy/tbb.hpp"
 #endif
 
-#if defined(RAJA_ENABLE_CUDA)
+#if defined(RAJA_ENABLE_CUDA) && defined(__CUDACC__)
```
davidbeckingsale: This may lead to unintended consequences if you are trying to use a RAJA-enabled CUDA without CUDA. We were actually planning to add another guard for this: #615

corbett5: I'm building RAJA with cuda in GEOS but currently none of our code uses cuda and therefore doesn't need to compile with nvcc. In fact compiling with nvcc raises errors because all compiler implemented methods (default constructors and whatnot) are declared host-device which gives rise to lots of `error: calling host function from host-device function`. So Adding in that check in #615 would actually break things for me.

davidbeckingsale: You should use a non-CUDA RAJA if you aren't using CUDA.

corbett5: Well I'm in the process of using cuda, the first step was to build RAJA with cuda. The second step was to port over a single solver while leaving the rest of the code unchanged. Without this guard every translation unit that includes RAJA (basically our entire code base) needs to be compiled with nvcc.

PR #626, opened 2019-07-08
Task/mdavis36/raja 293 segment tests

mdavis36:
Parameterized tests for TypedRangeSegment and TypedListSegment.

One small issue is that the `TypedRangeSegment::slice` function throws a warning during build for 4 byte or less data types. The warning is a narrowing conversion generated by `stripIndexType()`. The warning could be suppressed, however I assume there should be a more permanent fix to ensure this works correctly for end users?


trws:
That should only be happening there if the type it's coming from is four bytes or more, which is dangerous. I don't know that there's really a fix there other than ensuring that the test passes in appropriately sized inputs.

On 8 Jul 2019, at 10:58, Michael Davis wrote:

Parameterized tests for TypedRangeSegment and TypedListSegment.

One small issue is that the TypedRangeSegment::slice function throws a warning during build for 4 byte or less data types. The warning is a narrowing conversion generated by stripIndexType(). The warning could be suppressed, however I assume there should be a more permanent fix to ensure this works correctly for end users?

You can view, comment on, or merge this pull request online at:

  https://github.com/LLNL/RAJA/pull/626

Commit Summary

  *   TypedRangeSegment Constructor Testing
  *   TypedRangeSegment tests warnings
  *   Adding ListSegmment tests

File Changes

  *   M test/unit/index/CMakeLists.txt<https://github.com/LLNL/RAJA/pull/626/files#diff-0> (4)
  *   A test/unit/index/test-listsegment.cpp<https://github.com/LLNL/RAJA/pull/626/files#diff-1> (111)
  *   M test/unit/index/test-rangesegment.cpp<https://github.com/LLNL/RAJA/pull/626/files#diff-2> (67)

Patch Links:

  *   https://github.com/LLNL/RAJA/pull/626.patch
  *   https://github.com/LLNL/RAJA/pull/626.diff

rhornung67:
@trws Should we remove the tests for the potentially problematic index types? Although exhaustive unit testing is good, other than a few of these (int, long, signed and unsigned), I think users attempting the other types would be very rare. @mdavis36 what do you think?

ajkunen:
There are three issues that I see here:
1. In the TypedRangeSegment::slice() function, https://github.com/LLNL/RAJA/blob/develop/include/RAJA/index/RangeSegment.hpp#L158, the "+" operator will promote things smaller than an "int" to an int... so calculating the sliced begin and end will result in an int (even if your type is a short, etc.).  We probably just need to fix the slice() function to not use "auto" for begin and end, but rather DiffT.... or do a static_cast to DiffT.
2. The unit tests are going over a suite of signed and unsigned types, however the tests themselves have negative values.  This isn't causing the compilation errors, but will result in runtime problems.
3. We should really have a static_assert on RangeSegment and RangeStrideSegment segment that require DiffT to be signed... in these cases we get a RangeSegment<unsigned> will have a DiffT=unsigned

mdavis36:
@ajkunen
1. Switching to DiffT rather than auto did the trick, no more warnings during build, thank you!

2. Good point, I'll add some flow control for those instances of signed and unsigned types.

3. Are you referring to DiffT within just the slice functions of Range and RangeStride segments?

trws:
As long as that works, it's probably fine.  The remaining worry for me
is whether we can reasonably prevent overflow without allowing promotion
there, but I suppose the user will get what they asked for.

rhornung67:
@mdavis36, for the sake of code simplicity and ease of use, I think we should force the underlying index type to be signed. Allowing unsigned types is problematic as it will result in runtime issues (the distance between two iterates may need to be negative; for example, when using negative strides). So, I think we should add static assertion checks were it makes sense to catch user misuse as early as possible (e.g., segment constructors) and remove the unit tests with unsigned index types. If we add the assertion checks, they won't compile.  @ajkunen and @trws do you agree?

ajkunen:
@rhornung67 I agree, we should static_assert out on std::is_signed

trws:
I haven't dug into this far enough to be sure, but my normal response is that only difference_type needs to be signed.  It is generally a good idea to use signed types, but there are times when what you get for whatever reason is size_t for example, and it would be a narrowing conversion to even make that ssize_t.  Do we currently have a split between index_type and difference_type?

rhornung67:
@trws scanning the code, it looks like we distinguish between index type and difference type -- looks like someone thought ahead 🥇 Using std::make_unsigned to ensure that the difference type is always unsigned seems reasonable. It's not clear to me what the implications are for this in user code, but it would ensure correctness of RAJA internal operations, I believe. We still have to make sure we are covering the bases w.r.t. the type narrowing since the C++ standard says that the result type for arithmetic applied to types smaller than int will always be int (due to hardware registers, I think). @mdavis36 can you look into this approach? @trws should weigh in if I got this wrong.

trws:
I may have written it wrong, but I meant the other way around, difference_type should be signed whether the index type is or not.  In fact it's part of the documented requirements for iterator_traits: https://en.cppreference.com/w/cpp/iterator/iterator_traits

The value can be any type, as can the index, but the difference_type is signed.

rhornung67:
You're right. I mistakenly reversed signed/unsigned. @mdavis36 replace with std::make_signed in what I said earlier.

mdavis36:
@rhornung67 @trws Just to ensure I'm understanding this correctly. We want to ensure that the difference_type passed to the iterator within RangeSegment and RangeStrideSegment is always signed. This therefor allows the user to use both signed and unsigned types when constructing a Segment object.

This can be done instead of restricting the user to only signed types and adding assertion checks to constructors etc.

rhornung67:
@mdavis36 yes, I believe that's what we want to do. The implementation of the Segment classes will need to change a bit. I am probably missing some details, but I think we want to define DiffT to be std::make_signed(StorageT). And we probably need to change the use of DiffT in class methods, especially the arguments, to be StorageT in several places. @ajkunen may also have some thoughts on these details.

ajkunen:
I would say: 

1. make the default for DiffT be the std::make_signed<StorageT>, instead of just StorageT
AND
2. static_assert that DiffT is signed (in case the user specifies it)

rhornung67:
@ajkunen what about the arguments for a bunch of the segment methods that are DiffT type?  If StorageT is unsigned, then some of those methods won't work properly.

ajkunen:
I think that if it makes sense for an argument to a function to take a signed value (regardless of SotrageT's signedness), then it should be DiffT, otherwise the argument should be StorageT.

rhornung67:
@ajkunen I agree with that statement. I was hoping you would provide some insight into which method args should be changed to StorageT.... :-)

ajkunen:
@rhornung67 The question about which arguments that should take StorageT vs. DiffT is left as an exercise for the reader.

davidbeckingsale:
I think all the methods should take `StorageT` args, and `size()` needs to return a `DiffT`.

rhornung67:
Thanks @davidbeckingsale. 👍  That's what I was thinking too, but I didn't want to stick my neck out too far for fear of incurring ridicule from my colleagues.

rhornung67:
BTW, I encountered a valid use case recently where a TypedRangeStrideSegment with negative stride returned a negative value from size(). This change would fix that.

mdavis36:
Editing the RangeSegment Constructor to take StorageT arguments throws a narrowing warning when constructing the begin and end iterators as they use the difference_type to store their value, causing issues with unsigned types again. It would seem the Iterator class will need to be modified to allow for the use of the value_type instead of difference_type for storing its value.

mdavis36:
@rhornung67 @ajkunen @trws  @davidbeckingsale I believe I have addressed all of the issues we have  discussed, let me know what you think of the changes.

include/RAJA/index/RangeSegment.hpp
```diff
@@ -69,8 +69,18 @@ namespace RAJA
  *
  ******************************************************************************
  */
-template <typename StorageT, typename DiffT = strip_index_type_t<StorageT>>
+
+template<typename T>
+using make_signed_t = typename std::conditional <
```
davidbeckingsale: Is the conditional needed? Also, this should be moved to the same file as `strip_index_type_t`

include/RAJA/index/RangeSegment.hpp
```diff
@@ -87,8 +97,9 @@ struct TypedRangeSegment {
    * \param[in] begin the starting value (inclusive) for the range
    * \param[in] end the ending value (exclusive) for the range
    */
-  RAJA_HOST_DEVICE constexpr TypedRangeSegment(DiffT begin, DiffT end)
-      : m_begin(iterator{begin}), m_end(iterator{end})
+  using StripStorageT = strip_index_type_t<StorageT>;
+  RAJA_HOST_DEVICE constexpr TypedRangeSegment(StripStorageT begin, StripStorageT end)
```
davidbeckingsale: Should these arguments be the stripped type, or the actual storage type? @ajkunen

include/RAJA/index/RangeSegment.hpp
```diff
@@ -145,18 +156,18 @@ struct TypedRangeSegment {
...
+  RAJA_HOST_DEVICE RAJA_INLINE DiffT size() const { return m_end - m_begin; }
 
   //! Create a slice of this instance as a new instance
   /*!
    * \return A new instance spanning *begin() + begin to *begin() + begin +
    * length
    */
-  RAJA_HOST_DEVICE RAJA_INLINE TypedRangeSegment slice(DiffT begin,
-                                                       DiffT length) const
+  RAJA_HOST_DEVICE RAJA_INLINE TypedRangeSegment slice(StorageT begin,
```
davidbeckingsale: Same question as above - should these be the stripped type?

include/RAJA/index/RangeSegment.hpp
```diff
@@ -145,18 +156,18 @@ struct TypedRangeSegment {
   /*!
    * \return the range (end - begin) of this Segment
    */
-  RAJA_HOST_DEVICE RAJA_INLINE StorageT size() const { return m_end - m_begin; }
+  RAJA_HOST_DEVICE RAJA_INLINE DiffT size() const { return m_end - m_begin; }
```
davidbeckingsale: 👍

include/RAJA/index/RangeSegment.hpp
```diff
@@ -265,8 +279,9 @@ struct TypedRangeStrideSegment {
    * \param[in] end the ending value (exclusive) for the range
    * \param[in] stride the increment value for the iteration of the range
    */
-  RAJA_HOST_DEVICE TypedRangeStrideSegment(DiffT begin,
-                                           DiffT end,
+  using StripStorageT = strip_index_type_t<StorageT>;
+  RAJA_HOST_DEVICE TypedRangeStrideSegment(StripStorageT begin,
```
davidbeckingsale: Same questions as above.

include/RAJA/index/RangeSegment.hpp
```diff
@@ -350,12 +365,12 @@ struct TypedRangeStrideSegment {
    * \return A new instance spanning *begin() + begin * stride to *begin() +
    * (begin + length) * stride
    */
-  RAJA_HOST_DEVICE TypedRangeStrideSegment slice(DiffT begin,
-                                                 DiffT length) const
+  RAJA_HOST_DEVICE TypedRangeStrideSegment slice(StorageT begin,
+                                                 StorageT length) const
```
davidbeckingsale: One more storage vs. stripped.

include/RAJA/internal/Iterators.hpp
```diff
@@ -41,13 +43,14 @@ class numeric_iterator
 {
 public:
   using value_type = Type;
+  using s_value_type = strip_index_type_t<Type>;
```
davidbeckingsale: Why is this change needed?

include/RAJA/index/RangeSegment.hpp
mdavis36: conditional is needed in order to allow for the use of floating types. std::make_signed cannot natively handle floats as it is only designed to be used with integral types. However a case could be made that floating types should not be allowed to be used with RangeSegments as they are not compatible with RangeStrideSegment objects.

include/RAJA/index/RangeSegment.hpp
mdavis36: the iterator expects to receive a stripped type (it was previously receiving a DiffT value), therefore using RAJA generated types such as those created by RAJA_INDEX_VALUE etc. causes errors

include/RAJA/index/RangeSegment.hpp
mdavis36: Thesse don't need to be stripped type as that was already handled on the return statement on 172 `return TypedRangeSegment{stripIndexType(start), stripIndexType(end)};`

include/RAJA/index/RangeSegment.hpp
mdavis36: Same as previous answer, iterators.

include/RAJA/internal/Iterators.hpp
mdavis36: > Editing the RangeSegment Constructor to take StorageT arguments throws a narrowing warning when constructing the begin and end iterators as they use the difference_type to store their value, causing issues with unsigned types again. It would seem the Iterator class will need to be modified to allow for the use of the value_type instead of difference_type for storing its value. 

Value needed to be stripped a well, similar to before.

davidbeckingsale: I don't understand why enforcing a signed DiffT has required this change to the underlying iterator.

mdavis36: Because the value the iterator currently holds needs to be the same as the StorageT, storing the value as the DiffT can cause narrowing issues if the StorageT given to iterator is unsigned. DiffT is still used to modify the iterators value, it just no longer stores it as that.

include/RAJA/index/RangeSegment.hpp
davidbeckingsale: Could we still move this to a more centralized place?

include/RAJA/index/RangeSegment.hpp
davidbeckingsale: That makes sense to me. I wonder if we should strip them internally though, rather than accepting the stripped type. I'd be interested in what Adam thinks.

include/RAJA/index/RangeSegment.hpp
davidbeckingsale: I think length should be a `DiffT`, because it's a difference, not an element.

include/RAJA/index/RangeSegment.hpp
davidbeckingsale: Same additional question 😄

include/RAJA/index/RangeSegment.hpp
davidbeckingsale: I think `length` should be a `DiffT`

include/RAJA/internal/Iterators.hpp
davidbeckingsale: I don't think that `make_signed` should narrow though, if the signed type cannot represent the maximum distance between two `StorageT` then we could have a problem.

include/RAJA/internal/Iterators.hpp
```diff
@@ -178,7 +181,7 @@ class numeric_iterator
   }
 
 private:
-  difference_type val;
+  s_value_type val;
```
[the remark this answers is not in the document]
davidbeckingsale: We might want to come up with type naming conventions? In this case, I think this is the stripped type, so a longer and more descriptive name might help: `stripped_value_type`.

trws:
Given all the discussion around this, it might interest/amuse folks to read [this paper](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1522r0.pdf) from the current WG21 mailing. Basically trying to lay the groundwork to make it possible to deal with overflow in difference_type, gotta love these 20-year-old design problems. ><

include/RAJA/internal/Iterators.hpp
trws: @mdavis36 and I chatted about this a bit this afternoon.  While you're quite right @davidbeckingsale, it's a problem essentially every stl container, and every pair of pointers, has.  The `intptr_t` and `ptrdiff_t` types are frequently the same size, just with one being signed, and taking the difference of two pointers farther apart than PTRDIFF_MAX is UB... yay c++...

Anyway, my take on it is we could put in some debugging help if we want to do our due-diligence and catch the issue happening, but it's something we kinda just have to deal with allowing if the user asks for it.

davidbeckingsale:
Can you add a test case for the overflow?

mdavis36:
@rhornung67 @davidbeckingsale When you get a chance could you please review the latest changes, thank you.

mdavis36:
@rhornung67 I have updated the copyright information on all the new test files.

PR #628, opened 2019-07-11
More portable pragma macro. (courtesy of Al Nichols)

robinson96:
This implements a more portable version of the pragma macro that we're trying to promote in all of our code's dependencies.


trws:
I have no issue with this.  Vaguely surprised  we didn't have a stringify macro already actually.  Go for it.

PR #649, opened 2019-08-14
Replace all typedefs with using directives...

rhornung67:


rhornung67:
@davidbeckingsale now folks have less of a reason to call RAJA antedeluvian...

PR #652, opened 2019-08-14
Fix for seg fault in OMP Target when running PerfSuite. 

rchen20:
When running Polybench_ADI and Polybench_GESUMMV, RAJA OpenMP Target seg faults when the thread_limit(tperteam) clause does not match the number of thread iterations. Removing the clause staves off the "Warp Illegal Address" errors and allows them to run to completion, albeit with faulty checksums. A single-file non-RAJA reproducer is in the works, but it is difficult to reproduce in a non-RAJA environment. Will iterate with Johann and Roy on this (separately from this PR).

### DIFF
--- a/include/RAJA/policy/openmp_target/forall.hpp
+++ b/include/RAJA/policy/openmp_target/forall.hpp
@@ -60,9 +60,9 @@ RAJA_INLINE void forall_impl(const omp_target_parallel_for_exec<ThreadsPerTeam>&
     numteams = tperteam;
   }
 
+// thread_limit(tperteam) unused due to XL seg fault (when tperteam != distance)
 #pragma omp target teams distribute parallel for num_teams(numteams) \
-    thread_limit(tperteam) schedule(static, 1) map(to              \
-                                                    : body)
+    schedule(static, 1) map(to : body)
   for (Index_type i = 0; i < distance; ++i) {
     Body ib = body;
     ib(begin[i]);


rchen20:
@rhornung67 I have been doing exactly that; writing only the parallel for pragma with/without thread_limit(), and with direct array accesses (no RAJA). Unfortunately, this does not reproduce the error in normal runs. However, it does run into "Warp Illegal Address" when run through cuda-gdb or cudamemcheck. I'm tinkering with it, and may end up submitting more than one reproducer to IBM.

rhornung67:
@rchen20 what if you pull out the RAJA view/layout stuff and use that with the straight for-loop implemetation?

rchen20:
@rhornung67 If I eliminate RAJA View & Layout and use a normal array, the seg fault disappears. However, I performed bounds checking throughout the View and Layout code, and no accesses appear to exceed the bounds. I also tried an atomic reference count of accesses to View/Layout, and nothing out of the ordinary happened.

rhornung67:
@rchen20 I mean use a RAJA View in the c-style for-loop implementation.

rchen20:
@rhornung67 I just wrote a test for this in RAJA/test/unit/test-view.cpp and it worked (surprised we didn't have a test for this before). Putting View into a reproducer for IBM would probably not be tenable though . . .

trws:
When you say “worked” do you mean “caused the failure?” If so that’s great news.  IIRC we saw something related to (or at least like) this at the hackathon, but couldn’t reproduce a reasonable version of it.  Even with view this might be a great way to help them track it down.


On Aug 15, 2019, at 2:55 PM, Robert Chen <notifications@github.com<mailto:notifications@github.com>> wrote:


@rhornung67<https://github.com/rhornung67> I just wrote a test for this in RAJA/test/unit/test-view.cpp and it worked (surprised we didn't have a test for this before). Putting View into a reproducer for IBM would probably not be tenable though . . .

rchen20:
@trws I wrote a for loop which used a RAJA::View and it returned a correct result; so it did not reproduce the error. Rich meant for me to try a non-RAJA View in a loop (templated indexing but without RAJA), which I will do shortly. 

Yes, reproducing this during the Brookathon was quite difficult. I am not even certain that the fault is within View. Last week, I instrumented every constructor and operator call within View and Layout, and all calls were within expected bounds. I started instrumenting Kernel but didn't finish, and may instrument CAMP if I don't find any problems in Kernel.

PR #662, opened 2019-09-23
Suppress omp target warnings and cleaned up forall_impl.

rchen20:
Added OMP Target warning suppression for "can't inline" and "nostrict -O3". Macro'ed iterables in target forall_impl's for clarity.


rchen20:
@davidbeckingsale Done; also added the comments to Rich's previous host-config changes.

PR #663, opened 2019-09-24
Add note to docs about find_package(camp)

davidbeckingsale:


docs/sphinx/user_guide/using_raja.rst
```diff
@@ -21,16 +21,19 @@ CMake Configuration File
...
-capability to import RAJA into your CMake project.
+capability to import RAJA into your CMake project. RAJA also relies on the
+``camp`` library. This is bundled with the RAJA repository, and will install a
+``camp-config.cmake`` file that also works with CMake's find_package.
 
-To use the configuration file, you can add the following command to your CMake
+To use the configuration files, you can add the following command to your CMake
 project::
 
+  find_package(camp)
```
trws: Does the user have to do this one? The rest looks great but unless they're using camp themselves this should be optional right?

trws: Actually, if it installed correctly, why isn't RAJA's installed cmake file finding it? Are we missing a path we need to store in the FindRAJA.cmake?

rhornung67:
@davidbeckingsale are we going to merge this PR or does it still need work?

davidbeckingsale:
I was waiting for @trws to comment more. I believe we do need the `find_package(camp)`

davidbeckingsale:
@trws - this is the error I can reproduce:

```
CMake Error at CMakeLists.txt:22 (find_package):
  Found package configuration file:

    <snip>/raja/code/raja/install/share/raja/cmake/raja-config.cmake

  but it set RAJA_FOUND to FALSE so package "RAJA" is considered to be NOT
  FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing: camp
```

trws:
I agree there needs to be a find_package(camp), or equivalent, somewhere but why is that not taken care of by the raja-config.cmake itself? We really don’t want the user to get a different camp than the one raja was installed with do we?


On Oct 9, 2019, at 9:59 AM, David Beckingsale <notifications@github.com> wrote:

﻿

@trws<https://github.com/trws> - this is the error I can reproduce:

CMake Error at CMakeLists.txt:22 (find_package):
  Found package configuration file:

    <snip>/raja/code/raja/install/share/raja/cmake/raja-config.cmake

  but it set RAJA_FOUND to FALSE so package "RAJA" is considered to be NOT
  FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing: camp

davidbeckingsale:
I see what you mean – let me try and handle it there.

Cheers,
--David

davidbeckingsale:
@trws - I think this can be superseded by #669 . Let me know what you think.

davidbeckingsale:
Closed by #669

PR #666, opened 2019-09-27
Simplify bringing in RAJA with CHAI and Umpire into a CMake project

ajkunen:
Cleans up import of chai and umpire.
Also, makes it easier to bring in RAJA in a CMake project with add_subdirectory

(I realize that @davidbeckingsale is going to change the ways this works... but need this now for Kripke)


davidbeckingsale:
@ajkunen - can you point me at how you are using this in Kripke?

ajkunen:
@davidbeckingsale sure, check out https://github.com/LLNL/Kripke/blob/develop/CMakeLists.txt#L75 
We are bringing in RAJA, Umpire, Chai, etc. into Kripke as submodules, and including them with add_subdirectory
Note that this is in our develop branch

davidbeckingsale:
Got it. Do you think this branch should also have a check to see whether `umpire` is a target - since CHAI will break if Umpire is not found.

ajkunen:
@davidbeckingsale, I've made a similar change to CHAI.  CMake should pull in CHAI, which pulls in the dependency for Umpire.  CMake "does the right thing" and chains the dependencies correctly.  So if RAJA depends on CHAI, then CHAI is already handling the dependency check for Umpire.

I don't believe that RAJA directly touches Umpire... is that correct?  otherwise we should bring in CHAI and Umpire directly into RAJA.

davidbeckingsale:
It doesn't, so this should work just fine. Thanks for the clarification.

PR #667, opened 2019-09-27
Prevent user from declaring non-RAJA-policy CudaSyncs.

rchen20:
Privatized the CudaSyncThreads and Warp constructors to prevent misuse in lambdas.

### DIFF
--- a/include/RAJA/policy/cuda/kernel/Sync.hpp
+++ b/include/RAJA/policy/cuda/kernel/Sync.hpp
@@ -43,16 +43,12 @@ namespace statement
  * A RAJA::kernel statement that performs a CUDA __syncthreads().
  */
 struct CudaSyncThreads : public internal::Statement<camp::nil> {
-  private:
-  CudaSyncThreads() = default;
 };
 
 /*!
  * A RAJA::kernel statement that performs a CUDA __syncwarp().
  */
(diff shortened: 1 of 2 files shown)


include/RAJA/policy/cuda/kernel/Sync.hpp
```diff
@@ -43,12 +43,16 @@ namespace statement
  * A RAJA::kernel statement that performs a CUDA __syncthreads().
  */
 struct CudaSyncThreads : public internal::Statement<camp::nil> {
+  private:
+  CudaSyncThreads() = default;
```
ajkunen: Instead of deleting this constructor, delete the constructor in internal::Statement.
That will force the issue for all Statements

rchen20:
Regardless of whether it's privatized or deleted, this only issues a compiler warning. Will that be sufficient?

include/RAJA/policy/cuda/kernel/Sync.hpp
rchen20: Done. Still simply gives a warning in gcc though.

ajkunen:
@rchen20  deleting the constructor should be a compilation error if someone tries to construct an object from any of the derived statement types

rchen20:
@ajkunen You're right, my test case did not call the constructor. It definitely gives an error when I do now.

PR #696, opened 2019-11-15
Improving cache use in CudaKernel launcher

MrBurmark:
# Summary

- This PR improves caching of cuda occupancy calculator results in the CudaKernel launcher.


artv3:
How often does the occupancy calculator get invoked?

MrBurmark:
@artv3 Unfortunately the answer is somewhat complicated.
Ideally the occupancy calculator is called only once.
This PR makes that true when actual_threads equals the num_threads in the policy.

PR #697, opened 2019-11-15
Add "-Xcompiler -mno-float128" to CMAKE_CUDA_FLAGS when GCC 8 is used

davidbeckingsale:
# Summary

- This PR is a bugfix
- It does the following:
  - Adjusts the `CMAKE_CUDA_FLAGS` when CUDA is enabled and the host compiler is GCC 8+. It is required to make GCC 8 work as the host compiler.


CMakeLists.txt
```diff
@@ -84,6 +84,11 @@ if (ENABLE_CUDA)
     message(STATUS "CUDA compute architecture set to RAJA default sm_35 since it was not specified")
     set(CUDA_ARCH "sm_35" CACHE STRING "Set CUDA_ARCH to RAJA minimum supported" FORCE)
   endif()
```
davidbeckingsale: Note that this change is here (before BLT is loaded and outside of our `cmake/SetupCompilers.cmake` file because the flag needs to be set _before_ `enable_language(CUDA)` is called.

PR #702, opened 2019-12-04
Make plugin caller functions noexcept

davidbeckingsale:
# Summary

- This PR is a feature, that makes the two functions that call plugins `noexcept`


davidbeckingsale:
Just as a note, this will abort if the plugin throws an exception. @trws and I discussed adding a custom error handler - I will leave that for future work 😄

PR #704, opened 2019-12-06
Remove volatile from sequential atomic methods.

### LINKED ISSUE #703: Could the sequential atomic methods take a non volatile pointer?
TLDR: I'd like `RAJA::atomicAdd< RAJA::seq_atomic >` to work with a `std::string`.

Currently the sequential atomic methods such as `atomicAdd` take a volatile pointer
```c++
T atomicAdd(seq_atomic, T volatile *acc, T value)
```
I'm writing a general purpose container similar to the following
```c++
template< typename T >
struct Container
{
  template< typename Policy=RAJA::seq_atomic >
  void add( int position, T const & value )
  {
    RAJA::atomicAdd< Policy >( m_data + position, value );
  }

  T * m_data;
}
```
But the use of `atomicAdd` limits the container to only holding types for which T can be copy constructed from a volatile value, even if `Container< T >::add` is never instantiated with a non-sequential atomic policy.

It seems to me that the sequential atomic methods such as `atomicAdd` should have a signature of 
```c++
T atomicAdd(seq_atomic, T * acc, T value)
```
or better yet to eliminate a copy
```c++
T atomicAdd(seq_atomic, T * acc, T const & value)
```

corbett5:
Closes #703

### DIFF
--- a/include/RAJA/pattern/atomic.hpp
+++ b/include/RAJA/pattern/atomic.hpp
@@ -79,6 +79,31 @@ namespace RAJA
  *
  */
 
+/*!
+ * @brief Atomic store
+ * @param acc Pointer to location of result value
+ * @param value Value to store in *acc
+ * @return Returns value that was just stored at acc.
+ */
+RAJA_SUPPRESS_HD_WARN
+template <typename Policy, typename T>
+RAJA_INLINE RAJA_HOST_DEVICE void atomicStore(T * const acc, T const & value)
+{
+  RAJA::atomicStore(Policy{}, acc, value);
+}
+
+/*!
+ * @brief Atomic load
+ * @param acc Pointer to location of result value
+ * @return Returns value that was stored at acc.
+ */
+RAJA_SUPPRESS_HD_WARN
+template <typename Policy, typename T>
+RAJA_INLINE RAJA_HOST_DEVICE T atomicLoad(T * const acc)
+{
+  return RAJA::atomicLoad(Policy{}, acc);
+}
+
 
 /*!
  * @brief Atomic add
@@ -309,14 +334,14 @@ public:
   RAJA_HOST_DEVICE
   void store(value_type rhs) const
   {
-    *m_value_ptr = rhs;
+    RAJA::atomicStore<Policy>(m_value_ptr, rhs);
   }
 
   RAJA_INLINE
   RAJA_HOST_DEVICE
   value_type operator=(value_type rhs) const
   {
-    *m_value_ptr = rhs;
+    RAJA::atomicStore<Policy>(m_value_ptr, rhs);
     return rhs;
   }
 
@@ -324,14 +349,14 @@ public:
   RAJA_HOST_DEVICE
   value_type load() const
   {
-    return *m_value_ptr;
+    return RAJA::atomicLoad<Policy>(m_value_ptr);
   }
 
   RAJA_INLINE
   RAJA_HOST_DEVICE
   operator value_type() const
   {
-    return *m_value_ptr;
+    return RAJA::atomicLoad<Policy>(m_value_ptr);
   }
 
   RAJA_INLINE
--- a/include/RAJA/pattern/atomic.hpp
+++ b/include/RAJA/pattern/atomic.hpp
@@ -88,7 +88,7 @@ namespace RAJA
  */
 RAJA_SUPPRESS_HD_WARN
 template <typename Policy, typename T>
-RAJA_INLINE RAJA_HOST_DEVICE T atomicAdd(T volatile *acc, T value)
+RAJA_INLINE RAJA_HOST_DEVICE T atomicAdd(T * const acc, T const & value)
 {
   return RAJA::atomicAdd(Policy{}, acc, value);
 }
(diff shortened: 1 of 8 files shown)


artv3:
I think this is reasonable. ping @ajkunen

include/RAJA/pattern/atomic.hpp
```diff
@@ -80,6 +80,13 @@ namespace RAJA
  */
 
 
+namespace
+{
+  template <typename T>
+  using remove_volatile_t = typename std::remove_volatile< T >::type;
+}
+
+
```
MrBurmark: Should this be just defined once in a util header?
I thought unnamed namespaces were discouraged in headers.

ajkunen: agreed.  Seems like it should go in camp

corbett5: I removed this in my update.

include/RAJA/pattern/atomic.hpp
```diff
@@ -307,29 +307,29 @@ class AtomicRef
...
-  value_type load() const
+  value_type load() const volatile
   {
     return *m_value_ptr;
   }
 
   RAJA_INLINE
   RAJA_HOST_DEVICE
-  operator value_type() const
+  operator value_type() const volatile
```
MrBurmark: Should these methods be volatile.

corbett5: I changed `m_value_pointer` to be non-volatile since the atomic methods now accept a non-volatile pointer. The volatile qualification is recovered once you get to the specific policy implementation (except the serial implementation). However when `AtomicRef` access `m_value_pointer` directly I believe it should be volatile which is accomplished through the method qualification.

MrBurmark: I see what you mean, however adding volatile to the method will make the pointer instead of what it points to volatile.
Just removing volatile may be the way to go here.

corbett5: Duh. Would that be preferable to changing `m_value_ptr` to a reference in which case the `volatile` would propagate correctly?

corbett5: @MrBurmark let me know which you'd like to go and I'll make the fix.

MrBurmark: My current thinking is that volatile should be removed because atomic is not the same thing as volatile. The intention of this class is to be non-atomic in the seq case, so it should be subject to seq optimization in that case.
So my recommendation is to add policied atomic load and store functions and use them here instead of volatile.

corbett5: Sounds good to me.

corbett5: I removed the volatile qualifier from the methods but I didn't add policy specializations for the load and store. After reading a bit about volatile memory it's not clear to me that it does anything useful when implementing the atomics. However I'd be happy to add in policy specializations.

MrBurmark: This looks good. It makes sense for each function in atomicRef to be implemented via a policy specialized function.

corbett5: Done.

include/RAJA/policy/atomic_builtin.hpp
```diff
@@ -225,6 +225,19 @@ RAJA_INLINE T builtin_atomic_CAS_oper_sc(T volatile *acc,
 
 }  // namespace detail
 
+template <typename T>
+RAJA_DEVICE_HIP
+RAJA_INLINE void atomicStore(builtin_atomic, T volatile *acc, T value)
+{
+  *acc = value;
```
MrBurmark: This should use something like atomicExchange to ensure actual atomicity.

include/RAJA/policy/atomic_builtin.hpp
```diff
@@ -225,6 +225,19 @@ RAJA_INLINE T builtin_atomic_CAS_oper_sc(T volatile *acc,
...
+}
+
+template <typename T>
+RAJA_DEVICE_HIP
+RAJA_INLINE T atomicLoad(builtin_atomic, T volatile *acc)
+{
+  return *acc;
```
MrBurmark: This should also probably use an atomic to read it to ensure atomicity.
I've used something like this before.
`T dummy_for_cas;`
`return AtomicCAS(acc, dummy_for_cas, dummy_for_cas);`

include/RAJA/pattern/atomic.hpp
MrBurmark: This looks good, but I'm concerned that the atomic load and store don't actually use atomics to perform their function. I've used atomic exchange for store and atomicCAS with the same value for load before. Does that seem reasonable to you @trws.

include/RAJA/policy/atomic_builtin.hpp
trws: Yeah, this should definitely not be a bare store.

include/RAJA/policy/atomic_builtin.hpp
trws: I usually prefer `atomicAdd(acc, 0);` for performance, but that or an atomic load would work.

include/RAJA/policy/cuda/atomic.hpp
```diff
@@ -559,6 +559,22 @@ struct cuda_atomic {
  *
  * These are atomic in cuda device code and non-atomic otherwise
  */
+RAJA_SUPPRESS_HD_WARN
+template <typename T>
+RAJA_INLINE RAJA_HOST_DEVICE void atomicStore(cuda_atomic,
+                                              T volatile *acc,
+                                              T value)
+{
+  *acc = value;
```
trws: Same here.

trws:
The load and store ones need to be reworked, as it is they're non-atomic and don't even yield a compile barrier so they're not usable.  The sequential rework to non-volatile is probably fine, but I'm not 100% sure we shouldn't have a compile barrier in there even in the sequential case to prevent incorrect reordering if you want to take the volatile off.

MrBurmark:
This could be another place that seq and loop policies could be used to allow both of those sequential use cases.

PR #721, opened 2020-01-03
Removed RAJA_HOST_DEVICE modifier from defaulted constructors.

rchen20:
# Summary

- This PR is a bugfix.
- It does the following:
  - Removes RAJA_HOST_DEVICE from defaulted constructors to prevent NVCC warnings.

### DIFF
--- a/cmake/SetupCompilers.cmake
+++ b/cmake/SetupCompilers.cmake
@@ -70,7 +70,7 @@ endif()
 
 if (ENABLE_CUDA)
   set(CMAKE_CUDA_STANDARD 11)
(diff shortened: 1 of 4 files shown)


rchen20:
I will test it out with CUDA 8-10 soon, but there are a handful of one-off warnings of a similar type (in the example and tutorial code) that I'm trying to squash. When that's done I will ask for re-approval.

trws:
I would somewhat prefer that it be a single point ifdef to define something like `RAJA_DEVICE_HIP` for that file and use it rather than have the conditions everywhere, but functionally this works for me.

rchen20:
This should be ready for final approval. I've macro'ed the builtin atomic file, and removed the warning suppression. Leaving in the CUDA error/warning numbers for good measure. This has been checked with CUDA 9 and 10.

trws:
Looks good to me. Thanks!

PR #747, opened 2020-02-06
Fix Cuda atomicExchange

MrBurmark:
Support cuda atomicExchange in the same manner as other cuda atomics.
Add test for atomicExchange and atomicCAS.

# Summary

- This PR is a bugfix
- It does the following:
  - Fixes atomicExch\<cuda\>(double)
  - Fixes atomicExch\<hip\>(double)

### DIFF
--- a/test/unit/test-atomic.cpp
+++ b/test/unit/test-atomic.cpp
@@ -24,12 +24,12 @@ void testAtomicFunctionBasic()
 // initialize an array
 #if defined(RAJA_ENABLE_CUDA)
   T *dest = nullptr;
-  cudaErrchk(cudaMallocManaged((void **)&dest, sizeof(T) * 8));
+  cudaErrchk(cudaMallocManaged((void **)&dest, sizeof(T) * 10));
 
   cudaErrchk(cudaDeviceSynchronize());
 
 #else
-  T *dest = new T[8];
+  T *dest = new T[10];
 #endif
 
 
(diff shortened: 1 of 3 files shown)


test/unit/test-atomic.cpp
```diff
@@ -42,6 +42,8 @@ void testAtomicFunctionBasic()
   dest[5] = (T)0;
   dest[6] = (T)N + 1;
   dest[7] = (T)0;
+  dest[8] = (T)N;
+  dest[9] = (T)0;
```
rchen20: Does the size of the new and cudaMalloc above need to become 10?

MrBurmark: good catch, fixed.

test/unit/test-atomic.cpp
```diff
@@ -68,6 +72,10 @@ void testAtomicFunctionBasic()
   EXPECT_EQ((T)4, dest[5]);
   EXPECT_EQ((T)1, dest[6]);
   EXPECT_EQ((T)13, dest[7]);
+  EXPECT_LE((T)0, dest[8]);
+  EXPECT_GT((T)N, dest[8]);
+  EXPECT_LT((T)0, dest[9]);
+  EXPECT_GE((T)N, dest[9]);
```
rchen20: If you have a moment, would you mind also mirroring these changes to the HIP version of the test (testAtomicFunctionBasic_gpu() lines 263-302)? Looks great, and thanks again!

MrBurmark: that's a good idea.

rchen20:
Thank you! 👍🥇

PR #757, opened 2020-02-21
[WIP] CAMP resource manager integration

artv3:
# Summary

This PR integrates the camp resource manager.


artv3:
@trws @davidbeckingsale , I integrated the resource manager in the first example. I was wondering if you guys had some time to glance over my usage to see if this is what we had in mind?

examples/tut_add-vectors.cpp
```diff
@@ -166,36 +184,38 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
 #if defined(RAJA_ENABLE_HIP)
   std::cout << "\n Running RAJA HIP vector addition...\n";
 
-  int *d_a = memoryManager::allocate_gpu<int>(N);
-  int *d_b = memoryManager::allocate_gpu<int>(N);
-  int *d_c = memoryManager::allocate_gpu<int>(N);
+  Resource hipRc{Cuda()};
```
davidbeckingsale: We need to add a HIP resource?

davidbeckingsale:
Hey @artv3 - this looks fine. I think in the context of the test reorg we will need a way to map an execution policy to a Resource.

examples/tut_add-vectors.cpp
trws: We have one, is it not sufficient? https://github.com/LLNL/camp/blob/master/include/camp/resource/hip.hpp

trws:
> Hey @artv3 - this looks fine. I think in the context of the test reorg we will need a way to map an execution policy to a Resource.

We should probably add a helper for this, but the resource_from_platform type trait on the camp side is meant to ease that somewhat: https://github.com/LLNL/camp/blob/7ed94963456041e7a38232758b401c9cedc0f4a9/include/camp/resource.hpp#L116

```c++
template<typename Pol>
auto get_resource(int group = -1) {
    return typename camp::resource_from_platform<RAJA::platform_of<Policy>::value>::type(group);
}
Resource r(get_resource<cuhiompbb>());
...
```

davidbeckingsale:
Ah yes, and we should get #725 merged so we are using camp Platform underneath our policies.

examples/tut_add-vectors.cpp
artv3: Oops, copy paste error. Just pushed a fix.

artv3:
@trws @mdavis36 I posted up a bug fix for camp here https://github.com/LLNL/camp/pull/21. This should fix the travis build.

PR #759, opened 2020-02-21
Forall Functional RangeSegment tests

mdavis36:
# Summary

- This is a first pass at handling forall functional tests with RangeSegments. I believe this is close to what we discussed for structuring these tests.  Please let me know what you think.


test/functional/forall/test-forall-rangesegment.hpp
```diff
@@ -0,0 +1,73 @@
...
+#include "RAJA/RAJA.hpp"
+#include "camp/resource.hpp"
+
+template<typename T>
+void ForallRangeSegmentFunctionalTest_host(T first, T last)
+{
+  RAJA::TypedRangeSegment<T> r1(first, last);
+  RAJA::Index_type N = r1.end() - r1.begin();
+
+  camp::resources::Resource res{camp::resources::Host()};
```
artv3: Rather than having _host and _cuda variants, one solution I discussed with @rchen20 is always having two resource objects. Once for the host and the other for a device, here a device can be again the host, cuda or hip. You can then pass in the device resource manager as an argument in the functional test function. Inside the actual function we would still go through the motions of copying to and from the device. Does this make sense?

davidbeckingsale: I think you always want a host copy for checking the data, then you allocate a "working" copy based on the Platform/Policy - you copy the data back to the "check" array for doing the ASSERT.

The Umpire tests are structured in a similar way.

rhornung67: I agree with @artv3, especially in the header files that define the tests. I think our goal should be to have no programming model or architecture specific code in those files and use the source files to specialize those tests for the different back-end cases.

mdavis36: Okay thanks all, I'll have another go at this.

test/functional/forall/test-forall-rangesegment.hpp
```diff
@@ -0,0 +1,73 @@
...
+  RAJA::forall<RAJA::seq_exec>(r1, [=](RAJA::Index_type idx){
+    ASSERT_EQ(ref_array[idx-*r1.begin()], idx);
+  });
+
+  res.deallocate(ref_array);
+}
+
+#if defined(RAJA_ENABLE_CUDA)
+template<typename T>
+void ForallRangeSegmentFunctionalTest_cuda(T first, T last)
```
rchen20: As an example of combining @artv3's and @davidbeckingsale's suggestions, we would want the test case call for this function to look something like:
`ForallRangeSegmentFunctionalTest<TypeParam>(first, last, camp::resources::Resource temp{camp::resources::Cuda()});`

The function signature would change to something like:
`void ForallRangeSegmentFunctionalTest(T first, T last, camp::resources::Resource working)`

And finally, `dev_res` would become `working`, and the `host_res` would become `checking`.

test/functional/forall/test-forall-rangesegment.hpp
```diff
@@ -0,0 +1,73 @@
...
+template<typename T>
+void ForallRangeSegmentFunctionalTest_host(T first, T last)
+{
+  RAJA::TypedRangeSegment<T> r1(first, last);
+  RAJA::Index_type N = r1.end() - r1.begin();
+
+  camp::resources::Resource res{camp::resources::Host()};
+  T * ref_array = res.allocate<T>(N);
+
+  for(RAJA::Index_type i=0; i < r1.size(); i++)
```
rhornung67: If we include a parameter for the index type so we can easily test alternatives; e.g., int, long, etc. does this cause issues? I think we need to test cases representative of how users use these things ans assume that they are not all using RAJA::Index_type.

mdavis36: I was using these loops to iterate over the check arrays rather than the actual test so I didn't think it was as important for them to iterate over the testing-type, but I will change this type to the testing-type then it will reflect the testing loop.

test/functional/forall/test-forall-rangesegment.hpp
mdavis36: I want to try building up a list of testing TypeParams which contain information on our iteration type, execution policy and resource architecture, so for now I am going to try and pass the resource type during compile time, but if that doesn't work out I'll go ahead and revert back to passing it how you've suggested like this. Thanks @rchen20 !

mdavis36:
@rhornung67 @rchen20 @artv3 @davidbeckingsale Thank you for your earlier suggestions, I believe this is now ready for another review.

rhornung67:
@mdavis36 Checking in. I see you made a commit separating the back-end test instantiations into separate source files.  Is this ready to review?

mdavis36:
@rhornung67 Yes it is.

rhornung67:
@mdavis36 what do you think about deleting the old tests when we have them converted to the new structure?  I would like to avoid rework due to folks adding new tests in the old-tests directory.

mdavis36:
@rhornung67 I think we should once all of the forall segment type tests are done. The old forall tests combines all of the segment types within a single test file so its a bit tricky to remove the old ones individually.

rhornung67:
@mdavis36 OK. We just need to pay close attention to testing changes folks make so that we don't lose anything.

test/functional/forall/test-forall-cuda.cpp
```diff
@@ -0,0 +1,22 @@
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+// Copyright (c) 2016-20, Lawrence Livermore National Security, LLC
+// and RAJA project contributors. See the RAJA/COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#include "RAJA/RAJA.hpp"
```
rhornung67: @mdavis36 why not move the include statements into the test-forall.hpp file here and in the other *.cpp files below.

test/functional/forall/test-forall-rangesegment.hpp
```diff
@@ -0,0 +1,76 @@
...
+  using WORKING_RESOURCE = typename at<TypeParam, num<1>>::type;
+  using EXEC_POLICY      = typename at<TypeParam, num<2>>::type;
+
+  ForallRangeSegmentFunctionalTest<INDEX_TYPE, WORKING_RESOURCE, EXEC_POLICY>(0,5);
+  ForallRangeSegmentFunctionalTest<INDEX_TYPE, WORKING_RESOURCE, EXEC_POLICY>(1,5);
+  ForallRangeSegmentFunctionalTest<INDEX_TYPE, WORKING_RESOURCE, EXEC_POLICY>(1,255);
+
+  if(std::is_signed<INDEX_TYPE>::value){
+#if !defined(__CUDA_ARCH__) && !defined(RAJA_ENABLE_TBB)
+    ForallRangeSegmentFunctionalTest<INDEX_TYPE, WORKING_RESOURCE, EXEC_POLICY>(-5,0);
```
rhornung67: @mdavis36 do these tests using negative indices not work with CUDA or TBB?

mdavis36: They do work but they throw warnings just like with the rangesegment unit tests.

rhornung67: And we don't have a way to suppress them, correct?

trws: That sounds like a possible bug in our TBB and CUDA backends to me, if we allow negatives and throw warnings when they're used.  Do you happen to have the warnings handy?

mdavis36: /usr/workspace/wsrzd/davis291/raja2/test/functional/forall/test-forall-rangesegment.hpp(66): warning #68-D: integer conversion resulted in a change of sign
          detected during:
            instantiation of "void gtest_suite_ForallFunctionalTest_::RangeSegmentForall<gtest_TypeParam_>::TestBody() [with gtest_TypeParam_=camp::list<unsigned short, camp::resources::v1::Host, RAJA::policy::sequential::seq_exec>]"
/usr/WS2/davis291/raja2/blt/thirdparty_builtin/googletest-master-2020-01-07/googletest/include/gtest/internal/gtest-internal.h(470): here
            implicit generation of "gtest_suite_ForallFunctionalTest_::RangeSegmentForall<gtest_TypeParam_>::~RangeSegmentForall() [with gtest_TypeParam_=camp::list<unsigned short, camp::resources::v1::Host, RAJA::policy::sequential::seq_exec>]"
/usr/WS2/davis291/raja2/blt/thirdparty_builtin/googletest-master-2020-01-07/googletest/include/gtest/internal/gtest-internal.h(470): here
            instantiation of class "gtest_suite_ForallFunctionalTest_::RangeSegmentForall<gtest_TypeParam_> [with gtest_TypeParam_=camp::list<unsigned short, camp::resources::v1::Host, RAJA::policy::sequential::seq_exec>]"
/usr/WS2/davis291/raja2/blt/thirdparty_builtin/googletest-master-2020-01-07/googletest/include/gtest/internal/gtest-internal.h(470): here
            implicit generation of "gtest_suite_ForallFunctionalTest_::RangeSegmentForall<gtest_TypeParam_>::RangeSegmentForall() [with gtest_TypeParam_=camp::list<unsigned short, camp::resources::v1::Host, RAJA::policy::sequential::seq_exec>]"
/usr/WS2/davis291/raja2/blt/thirdparty_builtin/googletest-master-2020-01-07/googletest/include/gtest/internal/gtest-internal.h(470): here
            instantiation of class "gtest_suite_ForallFunctionalTest_::RangeSegmentForall<gtest_TypeParam_> [with gtest_TypeParam_=camp::list<unsigned short, camp::resources::v1::Host, RAJA::policy::sequential::seq_exec>]"
/usr/WS2/davis291/raja2/blt/thirdparty_builtin/googletest-master-2020-01-07/googletest/include/gtest/internal/gtest-internal.h(470): here
            [ 8 instantiation contexts not shown ]

trws: Yup, those are accurate warnings.  Ideally we would just make the `if` check an `if constexpr`, but we can't do that because it takes C++17.  We can probably make a workaround version in C++11 that would let you wrap it in a lambda and give the condition to a template, I'll look into that after I grab some lunch.  Alternately it could be handled with some sfinae on the ForallRangeSegmentFunctionalTest template.

test/functional/forall/test-forall-rangesegment.hpp
```diff
@@ -0,0 +1,76 @@
...
+#define __TEST_FORALL_RANGESEGMENT_HPP__
+
+#include "RAJA/RAJA.hpp"
+#include "camp/resource.hpp"
+
+#include "test-forall.hpp"
+
+using namespace camp::resources;
+using namespace RAJA;
+using namespace camp;
```
rchen20: @mdavis36 I think you've scoped everything in this file properly. Do we need these namespaces?

mdavis36: we do not, these are left over from refactoring into separate .cpp files, thank you

test/functional/forall/test-forall-seq.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+#include "test-forall-rangesegment.hpp"
+
+// Generate Sequential Type List
+using SequentialTypes = list< RAJA::seq_exec, 
+                              RAJA::loop_exec,
+                              RAJA::simd_exec
+                            >;
+
+using SequentialForallTypes = Test<cartesian_product< IdxTypes, ListHost, SequentialTypes >>::Types;
```
rchen20: I didn't know cartesian_product existed. This makes enumeration much cleaner; I'll refactor my tests to use it. 👍

trws: FWIW, before this it didn't.  @mdavis36 asked for a way to do something like that, and I'd had it on my to-do list for a while, so we have it now.  Expansion is probably a bit slow for now, but it's there.

test/functional/forall/test-forall-rangesegment.hpp
trws: Unfortunately there's nothing to be done about a helper for C++11, for 14 I can whip something up with a bit of work, but for 11 it will likely have to be tag dispatch.  Just add a function, `runNegativeTests` or something with one variant for true_type and one for false_type and dispatch on it.

test/functional/forall/test-forall-rangesegment.hpp
```diff
@@ -0,0 +1,76 @@
...
+  INDEX_TYPE N = r1.end() - r1.begin();
+
+  camp::resources::Resource working_res{WORKING_RES()};
+  camp::resources::Resource check_res{Host()};
+
+  INDEX_TYPE * working_array = working_res.allocate<INDEX_TYPE>(N);
+  INDEX_TYPE * check_array   = check_res.allocate<INDEX_TYPE>(N);
+  INDEX_TYPE * test_array    = check_res.allocate<INDEX_TYPE>(N);
+
+  for(INDEX_TYPE i=0; i < N; i++)
```
rhornung67: For brevity, you could change this loop to std::iota( test_array, test_array + N, *r1.begin() );

Now that I read this, it doesn't look much shorter :-)

test/functional/forall/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# SPDX-License-Identifier: (BSD-3-Clause)
+###############################################################################
+
+raja_add_test(
+  NAME test-forall-seq
+  SOURCES test-forall-seq.cpp)
+
+raja_add_test(
+  NAME test-forall-openmp
+  SOURCES test-forall-openmp.cpp)
```
mdavis36: @rhornung67 All tests will run successfully if their target architecture is enabled or not. Do we want to still build these tests even when they aren't enabled?

rhornung67: I think it's best to only build the tests when the corresponding back-ends are enabled. @davidbeckingsale @MrBurmark @trws do you agree?

trws: I agree with @rhornung67, we only want to build them if the backend is there.  I'm vaguely surprised they will build without the appropriate backend, what happens do they just build a null test?

mdavis36: Okay brilliant, I just wanted to check. Essentially yes, the tests just never gets instantiated with the specific back-end execution types.

test/functional/forall/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+###############################################################################
+# Copyright (c) 2016-19, Lawrence Livermore National Security, LLC
```
rhornung67: Update copyright year to '2016-20'

test/functional/forall/test-forall.hpp
```diff
@@ -0,0 +1,52 @@
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+// Copyright (c) 2016-20, Lawrence Livermore National Security, LLC
+// and RAJA project contributors. See the RAJA/COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#ifndef __TEST_FORALL_HPP__
+#define __TEST_FOARLL_HPP__
```
rhornung67: 'FOARLL' -> 'FORALL'

test/functional/forall/test-forall.hpp
```diff
@@ -0,0 +1,52 @@
...
+                      long int,
+                      unsigned long int,
+                      long long,
+                      unsigned long long>;
+
+using ListHost = list< camp::resources::Host >;
+
+TYPED_TEST_SUITE_P(ForallFunctionalTest);
+
+#endif //__TEST_FORALL_TYPES_HPP__
```
rhornung67: comment label doesn't match #define label above

trws: clang-format should have caught and/or fixed that, has it been run on these files?

mdavis36: No I will do that

test/functional/forall/test-forall-openmp.cpp
```diff
@@ -0,0 +1,22 @@
...
+// and RAJA project contributors. See the RAJA/COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#include "test-forall-rangesegment.hpp"
+
+// Generate OMP Type List
+#if defined(RAJA_ENABLE_OPENMP)
+using OMPTypes = list<RAJA::omp_parallel_exec<RAJA::seq_exec>,
```
rhornung67: @mdavis36 what is this policy? Does it create an openmp parallel region and then do a sequential loop inside it?  If so, I didn't know we had that.

Please also add RAJA::omp_parallel_for_exec policy to list.

mdavis36: That is exactly what it does. Added.

trws:
The outer part came of, I think Will’s(?), refactoring of the OpenMP 
policies a while back.  It put the parallel region in one place then we 
can parametrize it with different worksharing or sequential loops on the 
interior.  I think that one also works with region, but Arturo would 
know for sure.

rhornung67:
@mdavis36 I'm starting to work on the forall list segment functional tests. Should these be in the same directory as your range segment tests? If so, then you should rename the directory to 'forall-segment'. If not, then you should rename the directory to 'forall-rangesegment'.  When we get the segment tests done, we will move on to forall indexset tests, forall tests with reductions, etc. If we jam too much in a single directory, it will be hard to tell what's going on. Do you agree?

mdavis36:
@rhornung67 For each new set of forall tests that we add we should only need to add a new header file defining the test cases as we will still be running the same .cpp files for each backend. What do you think about having all of the headers for each type of forall test in a subdirectory of forall rather than splitting up forall across multiple top level directories and duplicating the backend .cpp files?

rhornung67:
@mdavis36 We can try that. My concern is that it may be easier to run a subset of tests if we can see a different executable for each reasonable subset. We can get input from others on the overall organization in the PR to merge into develop. 

Do you have a version of your tests organized like that I'll try to follow your lead and see how that looks as I'm working through the list segment tests.

rhornung67:
Also, we will need some ability to segregate tests since, for example, we can' do CPU forall-reduction tests with a SIMD exec policy. 

We will also need a set of reduction policies for reduction tests. I can see how we make those work in what you have now.

rhornung67:
@mdavis36 I think I see how you plan to organize these tests.  I'll let you know if I have more questions.

mdavis36:
@rhornung67 I just updated the structure to what I mentioned above if you want to go ahead and work off of this for the list-segments

rhornung67:
@mdavis36 thanks.  I'm working through how I want to test the list segments now. I will work that into your new structure.

I probably won't have anything for you to look at until tomorrow.  I have a concall 4-5pm today.

test/functional/forall/tests/test-forall-rangesegment.hpp
```diff
@@ -49,23 +49,32 @@ void ForallRangeSegmentFunctionalTest(INDEX_TYPE first, INDEX_TYPE last)
...
-  }
+  using INDEX_TYPE  = typename at<TypeParam, num<0>>::type;
+  using WORKING_RES = typename at<TypeParam, num<1>>::type;
+  using EXEC_POLICY = typename at<TypeParam, num<2>>::type;
+
+  ForallRangeSegmentFunctionalTest<INDEX_TYPE, WORKING_RES, EXEC_POLICY>(0, 5);
+  ForallRangeSegmentFunctionalTest<INDEX_TYPE, WORKING_RES, EXEC_POLICY>(1, 5);
+  ForallRangeSegmentFunctionalTest<INDEX_TYPE, WORKING_RES, EXEC_POLICY>(1, 255);
+
+  runNegativeTests<INDEX_TYPE, WORKING_RES, EXEC_POLICY>();
```
rhornung67: Nice 🥇

test/functional/forall/tests/test-forall.hpp
```diff
@@ -9,74 +9,26 @@
...
-  host_res.deallocate(test_array);
-}
-
-TYPED_TEST_SUITE_P(ForallFunctionalTest);
+#include "test-forall-rangesegment.hpp"
+REGISTER_TYPED_TEST_SUITE_P(ForallFunctionalSegmentTest,
+                            RangeSegmentForall);
+
+
+//TYPED_TEST_SUITE_P(ForallFunctionalReductionTest);
```
rhornung67: Please remove these commented out lines.

rhornung67:
@mdavis36 is this the PR you want a review on. If so, the base branch needs to be changed to RAJA-302-...

mdavis36:
#780 pulls the changes from this so I'm going to go ahead and close this out.

PR #804, opened 2020-05-02
Segments for Atomic Functional Forall Tests

rchen20:
# Summary

- This PR is a refactoring of the atomic forall test
- It does the following:
  - Adds simple TypedRangeStrideSegment and TypedListSegment to forall-atomic-basic tests.
  - Adds framework for TypedIndexSet, but not activated due to incompatible execution policies. This can be removed from the PR if so desired.

### DIFF
--- a/test/functional/forall/atomic-view/tests/test-forall-atomic-view.hpp
+++ b/test/functional/forall/atomic-view/tests/test-forall-atomic-view.hpp
@@ -32,12 +32,11 @@ void testAtomicViewBasic( RAJA::Index_type N )
   RAJA::TypedRangeSegment<RAJA::Index_type> seg(0, N);
   RAJA::TypedRangeSegment<RAJA::Index_type> seg_half(0, N / 2);
 
-  camp::resources::Resource src_res{WORKINGRES()};
-  camp::resources::Resource dest_res{WORKINGRES()};
+  camp::resources::Resource work_res{WORKINGRES()};
   camp::resources::Resource host_res{camp::resources::Host()};
 
-  T * source = src_res.allocate<T>(N);
-  T * dest = dest_res.allocate<T>(N/2);
+  T * source = work_res.allocate<T>(N);
+  T * dest = work_res.allocate<T>(N/2);
   T * check_array = host_res.allocate<T>(N/2);
 
 #if defined(RAJA_ENABLE_CUDA)
@@ -68,7 +67,7 @@ void testAtomicViewBasic( RAJA::Index_type N )
     sum_atomic_view(i / 2) += vec_view(i);
   });
 
-  dest_res.memcpy( check_array, dest, sizeof(T) * N/2 );
+  work_res.memcpy( check_array, dest, sizeof(T) * N/2 );
 
 #if defined(RAJA_ENABLE_CUDA)
   cudaErrchk(cudaDeviceSynchronize());
@@ -82,8 +81,8 @@ void testAtomicViewBasic( RAJA::Index_type N )
     EXPECT_EQ((T)2, check_array[i]);
   }
 
-  src_res.deallocate( source );
-  dest_res.deallocate( dest );
+  work_res.deallocate( source );
+  work_res.deallocate( dest );
   host_res.deallocate( check_array );
 }
 
--- a/test/functional/forall/atomic-view/tests/test-forall-atomic-view.hpp
+++ b/test/functional/forall/atomic-view/tests/test-forall-atomic-view.hpp
@@ -14,6 +14,7 @@
 
 #include <RAJA/RAJA.hpp>
 #include "RAJA_gtest.hpp"
+#include "camp/resource.hpp"
 
 TYPED_TEST_SUITE_P(ForallAtomicViewFunctionalTest);
 
@@ -33,9 +34,11 @@ void testAtomicViewBasic( RAJA::Index_type N )
 
   camp::resources::Resource src_res{WORKINGRES()};
   camp::resources::Resource dest_res{WORKINGRES()};
+  camp::resources::Resource host_res{camp::resources::Host()};
 
   T * source = src_res.allocate<T>(N);
   T * dest = dest_res.allocate<T>(N/2);
+  T * check_array = host_res.allocate<T>(N/2);
 
 #if defined(RAJA_ENABLE_CUDA)
   cudaErrchk(cudaDeviceSynchronize());
@@ -65,6 +68,8 @@ void testAtomicViewBasic( RAJA::Index_type N )
     sum_atomic_view(i / 2) += vec_view(i);
   });
 
+  dest_res.memcpy( check_array, dest, sizeof(T) * N/2 );
+
 #if defined(RAJA_ENABLE_CUDA)
   cudaErrchk(cudaDeviceSynchronize());
 #endif
@@ -74,11 +79,12 @@ void testAtomicViewBasic( RAJA::Index_type N )
 #endif
 
   for (RAJA::Index_type i = 0; i < N / 2; ++i) {
-    EXPECT_EQ((T)2, dest[i]);
+    EXPECT_EQ((T)2, check_array[i]);
   }
 
   src_res.deallocate( source );
   dest_res.deallocate( dest );
+  host_res.deallocate( check_array );
 }
 
 TYPED_TEST_P(ForallAtomicViewFunctionalTest, AtomicViewFunctionalForall)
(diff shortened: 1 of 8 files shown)


test/functional/forall/test-forall-atomic-utils.hpp
```diff
@@ -86,4 +88,73 @@ using AtomicDataTypeList =
...
+  RAJA::TypedListSegment<Index>
+  makeseg( Index N )
+  {
+    std::vector<Index> temp(N);
+    std::iota( std::begin(temp), std::end(temp), 0 );
+    return RAJA::TypedListSegment<Index>( &temp[0], static_cast<size_t>(temp.size()) );
+  }
+};
+
+template< typename Index >
```
rhornung67, line 128: @rchen20 you can remove the index set stuff. I think the segment stuff is sufficient.

test/functional/forall/test-forall-atomic-utils.hpp
```diff
@@ -86,4 +88,73 @@ using AtomicDataTypeList =
...
+struct RSMultiplexer < Index, RAJA::TypedListSegment<Index> >
+{
+  RAJA::TypedListSegment<Index>
+  makeseg( Index N )
+  {
+    std::vector<Index> temp(N);
+    std::iota( std::begin(temp), std::end(temp), 0 );
+    return RAJA::TypedListSegment<Index>( &temp[0], static_cast<size_t>(temp.size()) );
```
rhornung67, line 124: I don't understand how this works for non-CPU tests. You should have to pass a camp Resource object into the list segment constructor so it can create a copy of the indices in the proper memory space.

test/functional/forall/atomic/tests/test-forall-atomic-basic.hpp
```diff
@@ -25,10 +27,11 @@ class ForallAtomicBasicFunctionalTest : public ::testing::Test
 template <typename ExecPolicy,
           typename AtomicPolicy,
           typename WORKINGRES,
+          typename SegmentType,
```
rhornung67, line 30: I would have taken a simpler implementation route (that I described on the confluence page) to reduce the depth of googletest template nesting.If this doesn't take too long to compile and link, it works for me.

test/functional/forall/test-forall-atomic-utils.hpp
MrBurmark: Did the default behavior change to host instead of managed when built with cuda?
If we only run on our P9 systems with ATS then host memory is device accessible.
I agree that this should use a camp resource so it will work on any gpu system.

rchen20: This has not been tested on a machine like RZMANTA, but I will make this change to all of the atomic forall tests anyway for good measure.

test/functional/forall/atomic/tests/test-forall-atomic-basic.hpp
rchen20: Yes, I was thinking about the compilation time as well, but it does not seem to add that much compilation time.

test/functional/forall/test-forall-atomic-utils.hpp
rhornung67: Take a look at the forall segment tests to see how Mike and I used the camp resource objects for that. 

@MrBurmark the existing behavior for how list segments manage their index data (compile-time choice based on whether cuda or hip is enabled) is still there until we rip that out after the next release. Then, we will require users to supply camp resources so they can do what they want for any memory resource at runtime.

rchen20: @MrBurmark @rhornung67 Strange, all of the CUDA tests are failing on RZMANTA with the following. Have you seen this before?
[----------] 2 tests from Cuda/ForallIndexSetTest/0, where TypeParam = camp::list<long, camp::resources::v1::Cuda, RAJA::policy::indexset::ExecPolicy<RAJA::policy::sequential::seq_exec, RAJA::policy::cuda::cuda_exec<128ul, false> > >
[ RUN      ] Cuda/ForallIndexSetTest/0.IndexSetForall
CUDAassert: no kernel image is available for execution on the device /g/g20/chen59/work/RAJAatomictests/rajaatomicrangesMANTA/include/RAJA/policy/cuda/MemUtils_CUDA.hpp 210
unknown file: Failure
C++ exception with description "CUDAassert" thrown in the test body.
[  FAILED  ] Cuda/ForallIndexSetTest/0.IndexSetForall, where TypeParam = camp::list<long, camp::resources::v1::Cuda, RAJA::policy::indexset::ExecPolicy<RAJA::policy::sequential::seq_exec, RAJA::policy::cuda::cuda_exec<128ul, false> > > (153 ms)

rhornung67: But, it works on rzansel? What 'sm' level are you compiling with?

rchen20: Originally used sm_70, tried sm_60 but got the same error. I'm trying different CUDA versions now.

rchen20: All of the tests pass with the latest CUDA version 10.2.89 (even the tests where I did not use the CAMP resource). I'll add the camp::resources and update the build scripts to use the latest CUDA.

rhornung67: That's weird. DId they update the system software on rzmanta? Last time I tried cuda 10.2.*, there was a version mismatch with the cuda runtime.

rchen20: I guess they must have updated 10.2.89, but I haven't seen any official announcement from LC or John Gyllenhaal. The RAJA tests pass on RZANSEL with 10.2.89, so I will update the build scripts accordingly.

rhornung67: You mean add build scripts, correct?  We still need to keep the 10.1 scripts around because all our users are using 10.1.x.

Do the tests compile and pass with CUDA 10.1 on both machines?

rchen20: CUDA 10.1 builds and runs correctly on RZANSEL, and builds correctly on RZMANTA but fails the CUDA test runs (seq and omp tests work). I could add 10.2.89 scripts and keep the 10.1's around, or do you think I should try keeping the current set of scripts and altering the CUDA CMAKE options depending on the platform detected?

rhornung67: Keep the existing scripts and rename to include "10.1" instead of just "10".  Then add some for "10.2" using that in the name.

Have you contacted LC about what's going on with rzmanta?  It should work for both CUDA versions.

test/functional/forall/atomic/tests/test-forall-atomic-basic.hpp
```diff
@@ -25,10 +27,11 @@ class ForallAtomicBasicFunctionalTest : public ::testing::Test
...
           typename T>
 void testAtomicFunctionBasic( RAJA::Index_type seglimit )
 {
-  RAJA::TypedRangeSegment<RAJA::Index_type> seg(0, seglimit);
+  SegmentType seg = RSMultiplexer<RAJA::Index_type, SegmentType>().makeseg(seglimit);
```
mdavis36: Should this and seglimit be using type T instead of Index_type?

rchen20: This "seg" is a range segment, and seglimit should be either int or RAJA::Index_type to express the limit of the range. "T" can be a float or double and we probably don't want to pass a float/double into a rangesegment constructor.

rchen20:
@rhornung67 @mdavis36 @artv3 @MrBurmark This is ready to go, would you mind taking another look?

test/functional/forall/atomic-view/tests/test-forall-atomic-view.hpp
```diff
@@ -33,9 +34,11 @@ void testAtomicViewBasic( RAJA::Index_type N )
 
   camp::resources::Resource src_res{WORKINGRES()};
```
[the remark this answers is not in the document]
rchen20: It is moving data from a source array to a destination array, both on the device. And then destination is copied back and checked. We _could_ do it with one giant device array and views on different sections of that array, but that is more confusing.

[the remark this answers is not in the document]
rhornung67: I understand. But you could do everything as you are now with only one working resource. You don't need a resource for each array.  Like this:

https://github.com/LLNL/RAJA/blob/task/region-tests/test/functional/kernel/region/tests/test-kernel-region-utils.hpp#L29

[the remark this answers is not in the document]
rchen20: You're right, I've fixed it now. Thanks!

PR #808, opened 2020-05-07
Fix HIP linking issue on Corona

rchen20:
# Summary

- This PR is a bugfix
- It does the following:
  - HIP was not linking properly on Corona, most likely due to updates to the machine. This adds an rpath to fix the problem. I'm not sure if the rpath is in the correct Cmake variable, other Cmake experts please chime in.

Some caveats for using the updated Corona:
- Must compile on an allocated node (not login node). Try "salloc -N1 -pmi60 --exclusive".
- Almost all of the tests seems to pass (good work team!), but a few are so long-running that they timeout and cause the other tests to fail, and subsequently cause the device itself to fail. If we don't compile and run test-forall-atomic-ref-math, test-forall-atomic-ref-other, and test-kernel, then all the other tests pass.


rchen20:
@davidbeckingsale @rhornung67 I'll put in an LC ticket. It's using a special HIP linker, and it must not be picking up the new libraries they've installed.

PR #814, opened 2020-05-21
Sycl build and forall support

homerdin:
# Summary
This PR allows includes build changes and forall policy support for running on SYCL.  The policy `sycl_exec<BLOCK_SIZE>` is implemented to be used by forall loops.  The tests require a SYCL camp resource, an initial implementation is at (https://github.com/homerdin/camp/tree/sycl).  A build script and configuration file are included for Intel's SYCL compiler.  Support for the "unnamed kernel lambda" and "USM" extensions are required.

# Design review 
1. SYCL does not have global control for device execution, instead objects are used.  A single `cl::sycl::queue` within `src/MemUtils_SYCL.cpp` is used when needed.  
2. The `LOOP_BODY` and `Iterator` are erroring as non-trivially copyable and need to be copied over before execution.
3. The namespace naming follows the pattern for Hip and Cuda, but SYCL also uses the sycl namespace which may complicate maintenance.  
4. The policies are set to synchronous execution by default.  Without sharing a queue with the application there is no way to wait on an asynchronous execution within the application.  Resources will offer this support.


include/RAJA/policy/sycl/forall.hpp
```diff
@@ -0,0 +1,189 @@
...
+    sycl_dim_t gridSize = impl::getGridDim(static_cast<size_t>(len), BlockSize);
+
+    cl::sycl::queue q = ::RAJA::sycl::detail::getQueue();
+
+    // Non-trivially copyable
+    Iterator* idx = (Iterator*) cl::sycl::malloc_device(sizeof(Iterator), q);
+    auto e = q.memcpy(idx, &begin, sizeof(Iterator));
+    e.wait();
+
+    LOOP_BODY* lbody = (LOOP_BODY*) cl::sycl::malloc_device(sizeof(LOOP_BODY), q);
```
artv3, line 114: Are these mallocs pretty fast? I would be interested in knowing if they are negligible compared to the kernel run-time.

homerdin: I've started to implement the RAJA Perf Suite using the Sycl policies.  Getting the loop_body and iterator to the device are showing overheads.  I'm updating this to use unique_ptr's which can be passed to buffer allowing the for the runtime to take ownership and not block to free.  This is still showing overhead and I'm looking into it.

The Kernel policies I'm working on are showing very little overhead.

PR #819, opened 2020-06-08
add ppc architecture check for application of -mno-float128 to CMAKE_CUDA_FLAGS

rrsettgast:
# Summary
This PR adds an architecture check for the the use of `-mno-float128` to `CMAKE_CUDA_FLAGS`, which is a PowerPC flag. On a `x86` machine the current application of `-mno-float128` leads to an error.


MrBurmark:
This breaks me on ubuntu x86 as well.

PR #822, opened 2020-06-10
WorkGroups (kernel fusion)

MrBurmark:
# Add support for a WorkGroup construct that supports kernel fusion

Note that I'm starting this PR now mostly for testing purposes, but I'm open to feedback especially regarding organization.

- This PR is a feature
  - Adds WorkGroup/KernelFusion support at the request of Peter and myself

# Design review

On 6/2 at the AAPS meeting, I mentioned this PR.  We discussed the design ideas:

1. Add WorkGroups to support Kernel Fusion
2. Support different orderings for performance and debugging
3. Support scans (prefix sums) in WorkGroups
4. Support Reductions in WorkGroups

This PR will implement 1, 2.
A later PR will implement 3, 4 as they are extensions.

## Organization

The basic policy definitions are in policy/WorkGroup.hpp.
The generic parts of the implementation are in pattern/WorkGroup.hpp and pattern/WorkGroup/*.
The backend policy definitions are in policy/backend/policy.hpp.
The backend specializations are in policy/backend/Workgroup.hpp and policy/backend/Workgroup/*.

## Classes

WorkPool, WorkGroup, and WorkSite have generic implementations in pattern/WorkGroup.hpp.
WorkPool, WorkGroup, and WorkSite aare generic because they use specializations of the WorkStorage and WorkRunner classes in their implementations.
WorkStorage does not depend on a backend directly but does need a user supplied allocator that works with the backend in use, e.g. host memory for sequential and openmp backends, pinned memory for the cuda backend.
WorkRunner is declared in pattern/WorkGroup/WorkRunner.hpp. A generic ordered and reverse ordered implemention using forall is also in pattern/WorkGroup/WorkRunner.hpp. This generic implementation is used by per backend specializations of WorkRunner in policy/backend/Workgroup/WorkRunner.hpp. Any non-ordered implementations are backend specific and are also  in policy/backend/Workgroup/WorkRunner.hpp if they exist.

## Backend support

1. Ordered is supported by all backends by just using forall to run the loops.
  - Ordered and reverse ordered are supported
2. Unordered only has one implementation in the cuda backend.
  - I can think of many potential implementations for different backends but I'm reluctant to implement them if I don't know that they will be used and don't have a use case in order to test their performance.
  - I plan to add a hip unordered implementation when hip supports device function pointers.


include/RAJA/pattern/detail/WorkGroup.hpp
```diff
@@ -0,0 +1,1138 @@
...
+};
+
+/*!
+ * Generic struct used to layout memory for structs of unknown size.
+ * Assumptions for any size (checked in WorkStruct_construct):
+ *   offsetof(GenericWorkStruct<>, obj) == offsetof(WorkStruct<size>, obj)
+ *   sizeof(GenericWorkStruct) <= sizeof(WorkStruct<size>)
+ */
+template < typename ... CallArgs >
+using GenericWorkStruct = WorkStruct<alignof(std::max_align_t), CallArgs...>;
```
MrBurmark, line 242: @trws @willkill07 I'm curious what you think about how I'm handling the layout of the loop data.
I'm using a struct where the last member (obj) represents the offset where the loop data starts. I know the size when I get space for the WorkStruct and can use the true type, but after that I use it in a generic type erased version because and I do not know the size.

include/RAJA/pattern/detail/WorkGroup.hpp
```diff
@@ -0,0 +1,1138 @@
...
+template < typename T, typename ... CallArgs >
+void Vtable_move_construct(void* dest, void* src)
+{
+  T* dest_as_T = static_cast<T*>(dest);
+  T* src_as_T = static_cast<T*>(src);
+  new(dest_as_T) T(std::move(*src_as_T));
+}
+
+template < typename T, typename ... CallArgs >
+void Vtable_call(const void* obj, CallArgs... args)
```
willkill07, line 46: Shouldn't it be CallArgs&&... for perfect forwarding?

include/RAJA/pattern/detail/WorkGroup.hpp
willkill07: I think this is okay

include/RAJA/pattern/detail/WorkGroup.hpp
```diff
@@ -0,0 +1,1138 @@
...
+ *   offsetof(GenericWorkStruct<>, obj) == offsetof(WorkStruct<size>, obj)
+ *   sizeof(GenericWorkStruct) <= sizeof(WorkStruct<size>)
+ */
+template < typename ... CallArgs >
+using GenericWorkStruct = WorkStruct<alignof(std::max_align_t), CallArgs...>;
+
+
+template < typename loop_in, typename ... CallArgs >
+RAJA_INLINE
+void WorkStruct_construct(void* ptr,
```
willkill07, line 247: Any reason for free functions over public static members? As a user of autocompletion frameworks, seeing WorkStruct_{construct, move_destroy, destroy, call} seems a little jarring when it's an implementation detail.

include/RAJA/pattern/detail/WorkGroup.hpp
MrBurmark: The CallArgs are part of the signature of the call function so they have to stay the same as what the user supplied.

include/RAJA/pattern/detail/WorkGroup.hpp
MrBurmark: Moved them into the class

include/RAJA/pattern/detail/WorkGroup.hpp
MrBurmark: In the updated version you can see that these are part of the class signature not the method. Do you think I should template the method as well for perfect forwarding?

MrBurmark:
I think this PR has gotten far enough along and is stable enough that people can start taking a look at it.

MrBurmark:
One thing that I'm unsure about is how to handle stuff like plugins, due to the fact that the loop body is copied and stored separately from when the kernel is launched.
This is not a problem for the ordered policies which use forall.
However I'm not sure what to do for the unordered versions as they do not copy the loop body before they launch.

include/RAJA/pattern/WorkGroup/SimpleVector.hpp
```diff
@@ -0,0 +1,173 @@
...
+{
+
+namespace detail
+{
+
+/*!
+ * A simple vector like class
+ */
+template < typename T, typename Allocator >
+struct SimpleVector
```
artv3, line 36: Can this be merged with ``RAJAVec.hpp`` . I would also be in favor of a more descriptive name.

MrBurmark:
Yes this is using a newer camp, I committed the update try submodule update.

include/RAJA/pattern/WorkGroup/SimpleVector.hpp
MrBurmark: I missed RAJAVec, that's a good idea.

MrBurmark: Combined.

test/functional/workgroup/tests/test-workgroup-Unordered.hpp
```diff
@@ -0,0 +1,339 @@
...
+      pool(Allocator{});
+
+  pool.enqueue(RAJA::TypedRangeSegment<IndexType>{ begin, end },
+      [=] RAJA_HOST_DEVICE (IndexType i) {
+    working_array[i] += i + IndexType(5);
+  });
+
+  auto group = pool.instantiate();
+
+  auto site = group.run();
```
rchen20, line 86: I guess there's no value in checking the returned "site" because it is a placeholder for each run?

MrBurmark: ya, the WorkSite objects don't have any accessor methods for their state and none of them really have any state currently anyway.

MrBurmark:
I'm not sure either. For example the hip and cuda backends could share most of the unordered implementation that is in the cuda policy directory and the vtable code as well.

test/functional/workgroup/tests/test-workgroup-Unordered.hpp
rchen20: Can there be a situation where pool.instantiate() returns an invalid group?

MrBurmark: There shouldn't be, you can get an empty group however.

MrBurmark:
This PR is mostly feature complete and functional except for prooper plugin support which is waiting on #841 .
This is passing the travis tests, though still taking a while to compile with Intel.
It also has pages in the user guide and tutorial.
I think its ready for another look.

MrBurmark:
This PR is not feature complete with the addition of plugin support to workgroup.
Compile times for some of the test builds may still have to be addressed.

docs/sphinx/user_guide/tutorial/halo-exchange.rst
```diff
@@ -0,0 +1,230 @@
...
+  | 1 | 1 | 2 | 3 | 3 |
+  +---+---+---+---+---+
+  | 1 | 1 | 2 | 3 | 3 |
+  +---+---+---+---+---+
+  | 4 | 4 | 5 | 6 | 6 |
+  +---+---+---+---+---+
+  | 7 | 7 | 8 | 9 | 9 |
+  +---+---+---+---+---+
+  | 7 | 7 | 8 | 9 | 9 |
+  +---+---+---+---+---+
```
artv3, line 88: Nice! I'm a fan of this type of picture.

include/RAJA/pattern/WorkGroup/WorkRunner.hpp
```diff
@@ -0,0 +1,295 @@
...
+  using base = WorkRunnerForallOrdered_base<
+      FORALL_EXEC_POLICY,
+      EXEC_POLICY_T,
+      ORDER_POLICY_T,
+      ALLOCATOR_T,
+      INDEX_T,
+      Args...>;
+  using base::base;
+
+  // run the loops using forall in the reverse order to the order they were enqueued
```
pearce8, line 274: Why reverse order?

include/RAJA/pattern/WorkGroup/WorkRunner.hpp
```diff
@@ -0,0 +1,295 @@
...
+  template < typename WorkContainer >
+  typename base::per_run_storage run(WorkContainer const& storage, Args... args) const
+  {
+    using value_type = typename WorkContainer::value_type;
+
+    typename base::per_run_storage run_storage{};
+
+    auto begin = storage.begin();
+    for (auto iter = storage.end(); iter != begin; --iter) {
+      value_type::call(&*(iter-1), args...);
```
pearce8, line 284: Is "call" asynchronous?  I am interested in how we could add a "this item has been completed" notification (could be just writing a flag to a RAJAVec[iter] of the same size as storage).

include/RAJA/pattern/WorkGroup/WorkRunner.hpp
MrBurmark: This is the generic reverse ordered runner implementation, there is a generic inorder ordered runner above.

include/RAJA/pattern/WorkGroup/WorkRunner.hpp
MrBurmark: This run implements the reverse ordered policy and here call is a wrapper around forall and launches a single cuda kernel per item. Take a look in policy/cuda/workgroup for the unordered cuda WorkRunner.
I don't think doing the notification will be an issue as we can do something similar to what is done in the cuda reductions where the last block does the finalization, but I'm not sure how to express that in the interface.

MrBurmark:
I added a sentence to the docs that the workgroup API may change.

PR #830, opened 2020-06-17
bugfix/taller1/coronarajafixup

dtaller:
# Summary

This PR is a bugfix
- It does the following:
  - Allows RAJA to build on cz corona with ENABLE_HIP On, using clang compiler
      i. Changed a dependency from hip to hip_runtime, which is consistent with chai and umpire
  - Did not check tests and examples

### DIFF
--- a/include/RAJA/policy/hip/policy.hpp
+++ b/include/RAJA/policy/hip/policy.hpp
@@ -155,11 +155,13 @@ struct hip_thread_masked_loop {};
 
 //
 // Operations in the included files are parametrized using the following
-// values for HIP warp size and max block size.
+// values for HIP warp size and max block size. Note that we cannot use __HIPCC__
+// or __CUDACC__ here, because these values need to be defined when the compiler makes
+// passes targeting both the cpu and gpu
 //
-#if defined(__HIPCC__)
+#if defined(RAJA_ENABLE_HIP)
 constexpr const RAJA::Index_type WARP_SIZE = 64;
-#elif defined(__CUDACC__)
+#elif defined(RAJA_ENABLE_CUDA)
 constexpr const RAJA::Index_type WARP_SIZE = 32;
 #endif
 
--- a/include/RAJA/policy/hip/policy.hpp
+++ b/include/RAJA/policy/hip/policy.hpp
@@ -159,9 +159,9 @@ struct hip_thread_masked_loop {};
 // or __CUDACC__ here, because these values need to be defined when the compiler makes
 // passes targeting both the cpu and gpu
 //
-#if defined(RAJA_ENABLE_HIP)
+#if defined(__HIP_PLATFORM_HCC__)
 constexpr const RAJA::Index_type WARP_SIZE = 64;
-#elif defined(RAJA_ENABLE_CUDA)
+#elif defined(__HIP_PLATFORM_NVCC__)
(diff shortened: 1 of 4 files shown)


include/RAJA/policy/hip/policy.hpp
```diff
@@ -155,11 +155,13 @@ struct hip_thread_masked_loop {};
...
+// or __CUDACC__ here, because these values need to be defined when the compiler makes
+// passes targeting both the cpu and gpu
 //
-#if defined(__HIPCC__)
+#if defined(RAJA_ENABLE_HIP)
 constexpr const RAJA::Index_type WARP_SIZE = 64;
-#elif defined(__CUDACC__)
+#elif defined(RAJA_ENABLE_CUDA)
 constexpr const RAJA::Index_type WARP_SIZE = 32;
 #endif
```
MrBurmark, line 166: Does this change account for compiling with hip but targeting nvidia devices?

dtaller: Hi Jason. I took a look again and read some of the documentation, and made the change. I changed this to HIP_PLATFORM_HCC and HIP_PLATFORM_NVCC . These should give the behavior we want and be defined on both cpu and gpu compilations. I made the change, thanks.

I still need to figure out the continuous integration failure thing though

MrBurmark:
Does hip_runtime mean that the code is still being compiled as hip code, or only that the runtime API is included/linked?

dtaller:
> Does hip_runtime mean that the code is still being compiled as hip code, or only that the runtime API is included/linked?

I think that it means that the runtime API is included/linked. This is what chai and umpire did, and it seemed to resolve my issue with the extra quotes around the include that caused them to be ignored ( -O2 "-I/usr/workspace/taller1/CARE_TESTING_CORONA/RAJA_AGAIN_AGAI/include - ... ). However, when tests are enabled, the same issue pops back up (see the file blt/tests/smoke/CMakeLists.txt where the blt_hip_smoke test DEPENDS_ON hip rather than hip_runtime) and I am not sure how to resolve it. I am not great at CMake

davidbeckingsale:
@MrBurmark - hip_runtime dependency doesn't work the same as cuda_runtime 😢 it still pipes everything through hipcc. I looked before but couldn't quite unpick everything to make it only include/link the library rather than use compiler. The primary challenge is that the runtime libraries are different depending on whether you are doing hip->cuda or hip->amd.

MrBurmark:
Then this doesn't seem like a long term solution given that we really do want to build RAJA with hip not just the hip_runtime.
I put up a blt PR with the FindHIP from hip 3.5 (https://github.com/LLNL/blt/pull/368). This is what I've been using to build with hip.

dtaller:
> Then this doesn't seem like a long term solution given that we really do want to build RAJA with hip not just the hip_runtime.
> I put up a blt PR with the FindHIP from hip 3.5 ([LLNL/blt#368](https://github.com/LLNL/blt/pull/368)). This is what I've been using to build with hip.

Thanks! Yes, my solution was good enough for me to continue and compile CARE (raja is a dependency for me), but you are correct in that it is not the best long term.

davidbeckingsale:
@dtaller - is this PR still required?

dtaller:
> @dtaller - is this PR still required?

No, we can kill it

PR #832, opened 2020-06-17
Fix cmake target conflict

adayton1:
RAJA adds a custom cmake target called "docs" which conflicts with other cmake targets in a containing project. This pull request changes the target name to "raja-docs" to avoid the conflict.

- This PR is a bugfix
- It does the following:
  - Renames the "docs" cmake target to be "raja-docs"


adayton1:
My git history is messed up. Closing this pull request and opening a new one.

Issue #829, opened 2020-06-17
Allow using a previously imported camp cmake target


adayton1:
Is there a way for RAJA to pick up on if camp has already been imported as a cmake target before building RAJA? Maybe by adding `if (NOT TARGET camp)`?

adayton1:
Solved by pull request #834.

PR #833, opened 2020-06-17
Allow using imported camp cmake target

adayton1:
If camp is already imported as a cmake target, use that rather than an external camp or the camp submodule.

# Summary

- This PR is a bugfix
- It does the following:
  - Checks if camp is already a cmake target before trying to use an external camp or the camp submodule
  - Fixes issue #829

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,21 +233,23 @@ if (ENABLE_TBB)
     tbb)
 endif ()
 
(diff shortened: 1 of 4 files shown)


davidbeckingsale:
Can you remove the commit with the doc target changes from this PR please?

adayton1:
My history is messed up. Closing this pull request and opening a new one.

PR #836, opened 2020-06-23
Feature/runtimeplugin

JohnHynes:
# Plugin Documentation
Incomplete, but should still be helpful!
https://raja.readthedocs.io/en/feature-runtimeplugin/feature/plugins.html


JohnHynes:
Tests seem to be failing due to the inclusion of "dlfcn.h", on gcc it was working with the compiler flag -ldl as well as the usual -lRAJA and -fopenmp flags. To avoid this needing this flag we could implement dlopen and dlsym as functions in RuntimePluginLoader.hpp along with the bells and whistles that make them work.

davidbeckingsale:
Hey @JohnHynes, you should be able to add '-ldl' using CMake. Let me know if you want a hand with that.

include/RAJA/util/PluginInit.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include <vector>
+
+namespace RAJA {
+namespace plugin {
+
+std::vector<std::string> paths;
+
+void init(const std::string& path)
+{   
+    paths.push_back(path);
```
davidbeckingsale, line 14: I think it would be better if this looped over the plugins and called an "init" on each one. (avoiding the global "paths" variable).

include/RAJA/util/RuntimePluginLoader.hpp
```diff
@@ -0,0 +1,104 @@
...
+    {
+      return;
+    }
+    loadDirectory(env);
+  }
+
+  void preLaunch(RAJA::util::PluginContext p)
+  {
+    // Checking for new plugin directories to add.
+    if (!RAJA::plugin::paths.empty())
```
davidbeckingsale, line 30: I think this code should be moved to the init, doing it every time preLaunch is called will add too much overhead.

examples/plugin/init-plugin.cpp
```diff
@@ -0,0 +1,31 @@
+#include "RAJA/util/PluginStrategy.hpp"
```
davidbeckingsale, line 1: This file isn't needed anymore, we can delete it.

include/RAJA/util/PluginStrategy.hpp
```diff
@@ -25,6 +25,8 @@ class PluginStrategy
     virtual void preLaunch(PluginContext p) = 0;
 
     virtual void postLaunch(PluginContext p) = 0;
+
+    virtual void init(PluginOptions p) = 0;
```
davidbeckingsale, line 29: I think this function should not be a pure virtual. If you provide an empty implementation here, plugins that don't need to do any extra init don't need to implement it.

include/RAJA/util/plugins.hpp
```diff
@@ -38,7 +40,21 @@ callPostLaunchPlugins(PluginContext p) noexcept
...
+      ++plugin)
+  {
+    (*plugin).get()->init(p);
+  }
+}
+
 } // closing brace for util namespace
 } // closing brace for RAJA namespace
 
+#include "RAJA/util/PluginInit.hpp"
```
davidbeckingsale, line 58: Rather that include the file down here, I think you should just move the `init` inside this file, and rename it `init_plugins`.

examples/plugin/test-plugin-dynamic.cpp
```diff
@@ -0,0 +1,24 @@
...
+// SPDX-License-Identifier: (BSD-3-Clause)
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#include "RAJA/RAJA.hpp"
+#include <cstdlib>
+
+int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
+{
+  // Loading .so files as plugins in current directory.
+  RAJA::util::PluginRegistry::Add<RAJA::util::RuntimePluginLoader> P("RuntimePluginLoader", "RuntimePluginLoader");
```
rhornung67, line 14: @JohnHynes please change 'Add' to 'add' (lowercase) so this follows the convention of other RAJA methods. Thanks.

examples/plugin/timer-plugin.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include <iostream>
+#include <chrono>
+
+class TimerPlugin : public RAJA::util::PluginStrategy
+{
+public:
+  void preLaunch(RAJA::util::PluginContext RAJA_UNUSED_ARG(p))
+  {
+    start_time = std::chrono::steady_clock::now();
```
rhornung67, line 11: @davidbeckingsale do we want to use the RAJA Timer utility here instead?

examples/plugin/test-plugin-dynamic.cpp
davidbeckingsale: @rhornung67 - this change is orthogonal. The PluginRegistry was added with the uppercase method a while back.

examples/plugin/timer-plugin.cpp
davidbeckingsale: We could do - this is just an example though.

artv3:
@JohnHynes, looking good! If I want to try this out locally, how should I build the code?

src/RuntimePluginLoader.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "RAJA/util/RuntimePluginLoader.hpp"
+
+#include <dlfcn.h>
+#include <dirent.h>
+
+inline
+bool
+isSharedObject(const std::string filename)
+{
+  return (filename.size() > 3 && !filename.compare(filename.size() - 3, 3, ".so"));
```
MrBurmark, line 17: What level of cross platform (windows) support are we targeting with this capability?

docs/sphinx/user_guide/feature/plugins.rst
```diff
@@ -0,0 +1,97 @@
...
+Required Functions
+^^^^^^^^^^^
+The preLaunch and postLaunch functions are automatically called by RAJA before and after loop execution. This applies to RAJA's kernel and forall implementations.
+
+* ``void preLaunch(PluginContext& p) {}`` - Will occur before kernel/forall execution.
+
+* ``void postLaunch(PluginContext& p) {}`` - Will occur after kernel/forall execution.
+
+^^^^^^^^^^^
+Unrequired Functions
```
MrBurmark, line 32: Optional?

JohnHynes:
@artv3 Thank you! So a quick overview of an easy way to set everything up would be:
1. Build RAJA normally.

2. Set up your plugins to have the factory function "getPlugin" as seen in at the bottom of https://github.com/LLNL/RAJA/blob/feature/runtimeplugin/examples/plugin/timer-plugin.cpp

3. Compile your plugins to be shared object files with a .so extension (-shared).

4. Set an environment variable RAJA_PLUGINS to be the path to your plugin or directory.

5. Your plugins should work!

Alternatively, you may call RAJA::util::init_plugin("path/to/plugins"); in your project. The plugin loader will try to grab every .so file from the given path and use it as a plugin, so when using multiple plugins it's recommended to throw them all into one plugin folder and just point your path to that.

I've been working on a write-up as well, although incomplete some extra info can be found here:
https://github.com/LLNL/RAJA/blob/feature/runtimeplugin/docs/sphinx/user_guide/feature/plugins.rst

While we've tested only a few kokkos plugins, there is built-in support for them using the KOKKOS_PLUGINS environment variable.
https://github.com/kokkos/kokkos-tools

docs/sphinx/user_guide/feature/plugins.rst
```diff
@@ -0,0 +1,97 @@
...
+
+-----------------
+Dynamic Loading
+-----------------
+::
+
+  extern "C" RAJA::util::PluginStrategy *getPlugin ()
+  {
+    return new PluginName;
+  }
```
MrBurmark, line 58: It took me a couple minutes to realize that this is in the shared library.
Could you add a blurb so people can see on a very basic level what is happening during init_plugin and dynamic linking.

MrBurmark:
@JohnHynes this is looking pretty good.
You comment to @artv3 captures useful build knowledge that would be good to have in the docs.
I should let you know that I have a PR (#841) that modifies the plugins too and our PRs may conflict slightly.
Good job.

docs/sphinx/user_guide/feature/plugins.rst
JohnHynes: The writeup is incomplete, but I'll definitely be including more detailed information!

src/RuntimePluginLoader.cpp
JohnHynes: Some of the current code is more geared towards linux at the moment, but cross platform support can pretty easily be added. I'm not too experienced with windows shared object libraries and their differences from linux, so that function is somewhat of a placeholder for something a little more versatile in terms of cross platform support. I do apologize for the lack of commenting about that.

JohnHynes: While on the topic of cross platform support for windows, @willkill07 mentioned this repository:
https://github.com/dlfcn-win32/dlfcn-win32?fbclid=IwAR3OMQuPd3OW-jPQ7SboTZikd3yeDhnoMwlIt9mRb0-yoFacWE4-DTzlC3w

Which provides wrapper functions for dlopen and dlsym compatible with windows when included in the CMake. I haven't gotten around the implementing it yet, but it's worth noting!

examples/plugin/CMakeLists.txt
```diff
@@ -8,3 +8,10 @@
 raja_add_executable(
   NAME plugin-example
   SOURCES test-plugin.cpp counter-plugin.cpp)
+
+raja_add_executable(
+  NAME plugin-example-dynamic
+  SOURCES test-plugin-dynamic.cpp)
+  
+add_library(timer_plugin SHARED timer-plugin.cpp)
+target_link_libraries(timer_plugin PRIVATE RAJA)
```
willkill07, line 17: @davidbeckingsale this `target_link_libraries` call with `PRIVATE RAJA` was needed due to needing camp/RAJA includes. I realize this is tangential, but do you think there is a way we could create an include-only RAJA target that won't automatically link `libRAJA.a` ?

include/RAJA/util/PluginLinker.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+namespace {
+  struct pluginLinker {
+    inline pluginLinker() {
+      (void)RAJA::util::linkRuntimePluginLoader();
+      (void)RAJA::util::linkKokkosPluginLoader();
+    }
+  } pluginLinker;
+}
+#endif
```
willkill07, line 22: You'll want to add a newline at the end of this file

include/RAJA/util/PluginOptions.hpp
```diff
@@ -0,0 +1,31 @@
...
+// SPDX-License-Identifier: (BSD-3-Clause)
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#ifndef RAJA_Plugin_Options_HPP
+#define RAJA_Plugin_Options_HPP
+
+namespace RAJA {
+namespace util {
+
+#include <string>
```
willkill07, line 14: Move this include prior to the namespace nesting. This can do very ugly things

include/RAJA/util/PluginOptions.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+struct PluginOptions
+{
+    PluginOptions(const std::string newstr) : str(newstr) {};
```
willkill07, line 18: parameter should be passed by const reference

src/KokkosPluginLoader.cpp
```diff
@@ -0,0 +1,136 @@
...
+  return (filename.size() > 3 && !filename.compare(filename.size() - 3, 3, ".so"));
+}
+
+namespace RAJA {
+namespace util {
+  
+  KokkosPluginLoader::KokkosPluginLoader()
+  {
+    char *env = ::getenv("KOKKOS_PLUGINS");
+    if (nullptr == env)
```
willkill07, line 26: Pedantic, but flip the operands `(env == nullptr)`

src/KokkosPluginLoader.cpp
```diff
@@ -0,0 +1,136 @@
...
+    {
+      return;
+    }
+    initDirectory(std::string(env));
+
+    for (auto &func : init_functions)
+    {
+      func(0, 20171029, 0, nullptr);
```
willkill07, line 34: 20171029 seems like a magic number. Could this be made a constant near the top of the file?

include/RAJA/util/PluginOptions.hpp
rhornung67: 👍

include/RAJA/util/PluginOptions.hpp
rhornung67: 👍

src/KokkosPluginLoader.cpp
rhornung67: 👍

src/KokkosPluginLoader.cpp
rhornung67: 👍

docs/sphinx/user_guide/feature/plugins.rst
rhornung67: @JohnHynes yes, 'optional' is much better

docs/sphinx/user_guide/feature/plugins.rst
```diff
@@ -0,0 +1,97 @@
...
+.. ## SPDX-License-Identifier: (BSD-3-Clause)
+.. ##
+
+.. _plugins-label:
+
+========
+Plugins
+========
+
+RAJA supports user-made plugins that may be loaded either at the time of compilation or during runtime. These two methods are not mutually exclusive, as plugins loaded dynamically can be run alongside plugins that are loaded statically.
```
rhornung67, line 15: The second sentence should be moved to below the 'Plugin API' section header so it's clearer what you're talking about.

docs/sphinx/user_guide/feature/plugins.rst
```diff
@@ -0,0 +1,97 @@
...
+The preLaunch and postLaunch functions are automatically called by RAJA before and after loop execution. This applies to RAJA's kernel and forall implementations.
+
+* ``void preLaunch(PluginContext& p) {}`` - Will occur before kernel/forall execution.
+
+* ``void postLaunch(PluginContext& p) {}`` - Will occur after kernel/forall execution.
+
+^^^^^^^^^^^
+Unrequired Functions
+^^^^^^^^^^^
+The init and finalize functions have standard implementations and thus are not required to be included in a user-made plugin. Init and finalize are never run by RAJA by default and are only run when the user makes a call to RAJA::util::init_plugin() or RAJA::util::finalize_plugin() respectively.
```
rhornung67, line 34: 'required to be included' -> 'needed'. Shorter and to the point is better.

docs/sphinx/user_guide/feature/plugins.rst
```diff
@@ -0,0 +1,97 @@
...
+
+* ``void init(PluginOptions p) {}`` - Called by the user
```
rhornung67, line 36: A bit more explanation about what these methods do would be helpful.

docs/sphinx/user_guide/feature/plugins.rst
```diff
@@ -0,0 +1,97 @@
...
+
+* ``void finalize() {}``
+
+
+-----------------
```
rhornung67, line 41: Please add an explanation about how static and dynamic loading work differently and when to prefer one over the other.

docs/sphinx/user_guide/feature/plugins.rst
```diff
@@ -0,0 +1,97 @@
...
+::
+
+  #include "RAJA/util/PluginStrategy.hpp"
+  #include <iostream>
+
+  class CounterPlugin : public RAJA::util::PluginStrategy
+  {
+    public:
+    void preLaunch(RAJA::util::PluginContext& p) {
+      if (p.platform == RAJA::Platform::host)
```
rhornung67, line 74: Please add curly braces around the statements in the if and else clauses.

docs/sphinx/user_guide/feature/plugins.rst
```diff
@@ -0,0 +1,97 @@
...
+    
+    private:
+    int host_counter;
+    int device_counter;
+  };
+
+  // Statically loading plugin.
+  static RAJA::util::PluginRegistry::add<CounterPlugin> P("Counter", "Counts number of kernel launches.");
+  
+  // Dynamically loading plugin.
```
rhornung67, line 91: Since this is an example to explain how to build a user-defined plugin, please explain why the dynamic load part needs the 'extern "C"' business.  If this code is in the examples directory, please include it using here directly from the source file, so the docs are easier to maintain; i.e., if the code changes the docs will change with it automatically.  An example of this can be found in the user_guide/tutorial/sca.rst file.

examples/plugin/timer-plugin.cpp
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  void postLaunch(RAJA::util::PluginContext& p)
+  {
+    end_time = std::chrono::steady_clock::now();
+    double elapsedMs = std::chrono::duration<double, std::milli>(end_time - start_time).count();
+
+    if (p.platform == RAJA::Platform::host)
```
rhornung67, line 26: Please add curly braces around the statements in the if and else clauses.

include/RAJA/util/KokkosPluginLoader.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+  class KokkosPluginLoader : public ::RAJA::util::PluginStrategy
+  {
+    using Parent = ::RAJA::util::PluginStrategy;
+    typedef void (*init_function)(const int, const uint64_t, const uint32_t, void*);
+    typedef void (*pre_function)(const char*, const uint32_t, uint64_t*);
+    typedef void (*post_function)(uint64_t);
+    typedef void (*finalize_function)();
+
+  public:
```
rhornung67, line 28: Please move the public section of the class to the top. It's easier for users to see the interface in the docs that way. Most users shouldn't need to look at the implementation so it shouldn't be the first thing they see.

include/RAJA/util/PluginLinker.hpp
rhornung67: 👍

include/RAJA/util/PluginOptions.hpp
rhornung67: 👍

include/RAJA/util/RuntimePluginLoader.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include "RAJA/util/PluginOptions.hpp"
+#include "RAJA/util/PluginStrategy.hpp"
+
+namespace RAJA {
+namespace util {
+
+  class RuntimePluginLoader : public ::RAJA::util::PluginStrategy
+  {
+
+    using Parent = ::RAJA::util::PluginStrategy;
```
rhornung67, line 23: The leading double-colon shouldn't be needed here.

src/KokkosPluginLoader.cpp
```diff
@@ -0,0 +1,136 @@
...
+      perror("[KokkosPluginLoader]: Could not open plugin directory");
+    }
+  }
+
+  void linkKokkosPluginLoader() {}
+
+} // end namespace util
+} // end namespace RAJA
+
+static RAJA::util::PluginRegistry::add<RAJA::util::KokkosPluginLoader> P("KokkosPluginLoader", "Dynamically load plugins ported from the Kokkos library.");
```
rhornung67, line 136: Please add a newline to end of file.

test/integration/plugin_for_test_dynamic.cpp
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  void postLaunch(RAJA::util::PluginContext& RAJA_UNUSED_ARG(p)) {
+  }
+};
+
+extern "C" RAJA::util::PluginStrategy *getPlugin()
+{
+  return new ExceptionPlugin;
+}
```
rhornung67, line 26: Please add a newline to end of file.

test/integration/plugin_for_test_kokkos.cpp
```diff
@@ -0,0 +1,34 @@
...
+    (void)name;
+    (void)devID;
+    (void)kID;
+}
+
+extern "C" void kokkosp_end_parallel_for(const uint64_t kID) {
+    (void)kID;
+}
+
+extern "C" void kokkosp_finalize_library() {}
```
rhornung67, line 34: Please add a newline after this.

CMakeLists.txt
```diff
@@ -254,7 +256,7 @@ endif (NOT TARGET camp)
 blt_add_library(
   NAME RAJA
   SOURCES ${raja_sources}
-  DEPENDS_ON ${raja_depends} camp)
+  DEPENDS_ON ${raja_depends} camp "-l${CMAKE_DL_LIBS}")
```
davidbeckingsale, line 259: Do you need the -l here?

docs/sphinx/user_guide/feature/plugins.rst
```diff
@@ -0,0 +1,130 @@
...
+------------
+
+RAJA supports user-made plugins that may be loaded either at the time of compilation or during runtime. These two methods are not mutually exclusive, as plugins loaded statically can be run alongside plugins that are loaded dynamically.
+
+------------
+Using RAJA Plugins
+------------
+
+^^^^^^^^^^^^^^^^^
+Static vs Dynamic Linking
```
rhornung67, line 26: The term 'linking' should be changed to 'loading' in the user docs and elsewhere.

include/RAJA/util/Registry.hpp
```diff
@@ -91,14 +91,14 @@ namespace util {
 
     /// A static registration template.
     template <typename V>
-    class Add {
+    class add {
```
rhornung67, line 94: What is the reason for the change to all lower case?  Typically, class names are capitalized.

src/KokkosPluginLoader.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    // Getting and storing init function.
+    init_function init = (init_function) dlsym(plugin, "kokkosp_init_library");
+    if (init)
+      init_functions.push_back(init);
+    else
+      printf("[KokkosPluginLoader]: dlsym failed: %s\n", dlerror());
+
+    // Getting and storing preLaunch function.
+    pre_function pre = (pre_function) dlsym(plugin, "kokkosp_begin_parallel_for");
```
rhornung67, line 81: @davidbeckingsale should this be changed -- 'parallel_for' to 'forall' -- to make output more familiar to RAJA users?

include/RAJA/util/Registry.hpp
JohnHynes: I can find a workaround for this. To change the name of the constructor from "Add" to "add" I also had to change the class name.

rhornung67: OK. That's fine, then.

src/KokkosPluginLoader.cpp
DavidPoliakoff: @rhornung67 , if you want to use it with existing Kokkos tools, those tools use this name explicitly. Sorry about that, if we were planning ahead we might have done it differently.

What you might do is ask people to expose another symbol, `begin_forall`, that just calls into the other, I don't know how hard a sell that would be to Tools people though (I could do it for Caliper if it helped)

rhornung67: I'm fine keeping it as is. No worries. I was just asking.  'begin_forall' isn't really the best either, since the plugins also work with the RAJA kernel interface.

rhornung67:
@davidbeckingsale are we going to have to turn off Appveyor CI for the plugin stuff? There should be a way to make it work with the proper preprocessor guards and function calls, correct?

davidbeckingsale:
> @davidbeckingsale are we going to have to turn off Appveyor CI for the plugin stuff? There should be a way to make it work with the proper preprocessor guards and function calls, correct?

Yup, we can just disable it on Windows in CMake and/or code as necessary.

DavidPoliakoff:
Quick recommendation: only disable the dlsym calls on Windows. What we have Windows users doing now is that all of these callbacks can be set through an API. So (in `develop`) we just guard the dlsym calls, and allow somebody to essentially "load" a plugin at runtime by providing the callbacks. That turns out to be really handy.

In RAJA, the equivalent would be "guard the dlsym calls, not any other part of the plugin infrastructure" (though of course you can do all that later, and for now just guard it all)

examples/plugin/CMakeLists.txt
trws: That's not a bad idea, we could get them out of the target by just requesting the includes I suppose, but we probably want everything but the actual link there anyway.

JohnHynes:
@trws Do you know why we're testing for clang 5? @rhornung67 @davidbeckingsale  and I aren't quite sure.

trws:
Only legacy reasons at this point.



On August 4, 2020 at 1:00:58 PM, John (notifications@github.com<mailto:notifications@github.com>) wrote:

@trws<https://github.com/trws> Do you know why we're testing for clang 5? @rhornung67<https://github.com/rhornung67> @davidbeckingsale<https://github.com/davidbeckingsale> and I aren't quite sure.

docs/sphinx/user_guide/feature/plugins.rst
```diff
@@ -0,0 +1,129 @@
...
+
+  static RAJA::util::PluginRegistry::add<PluginName> P("Name", "Description");
+
+
+^^^^^^^^^^^^^^^^^
+Dynamic Loading
+^^^^^^^^^^^^^^^^^
+If the plugin is to be dynamically loaded to a project during runtime, the RAJA Plugin API requires a few conditions to be met. The following must be true about the plugin, not necessarily of the project using it.
+
+1. **The plugin must have following factory function.** This will return a pointer to an instance of your plugin. Thanks to the ``extern "C"`` a project will be able to search for "getPlugin" within the dynamically loaded plugin correctly.
```
MrBurmark, line 106: missing the

include/RAJA/util/PluginStrategy.hpp
```diff
@@ -9,26 +9,30 @@
...
   public:
     PluginStrategy();
 
     virtual ~PluginStrategy() = default;
 
+    virtual void init(PluginOptions p);
+
     virtual void preCapture(PluginContext p);
 
     virtual void postCapture(PluginContext p);
```
MrBurmark, line 29: Should these be pass by reference?

include/RAJA/util/RuntimePluginLoader.hpp
```diff
@@ -0,0 +1,54 @@
...
+    using Parent = RAJA::util::PluginStrategy;
+
+  public:
+    RuntimePluginLoader();
+
+    void init(RAJA::util::PluginOptions p);
+
+    void preCapture(RAJA::util::PluginContext p);
+
+    void postCapture(RAJA::util::PluginContext p);
```
MrBurmark, line 31: Pass by reference?

src/PluginStrategy.cpp
```diff
@@ -14,13 +14,17 @@ namespace util {
 
 PluginStrategy::PluginStrategy() = default;
 
+void PluginStrategy::init(PluginOptions) { }
+
 void PluginStrategy::preCapture(PluginContext) { }
 
 void PluginStrategy::postCapture(PluginContext) { }
```
MrBurmark, line 21: Pass by reference?

src/RuntimePluginLoader.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+void RuntimePluginLoader::preCapture(RAJA::util::PluginContext p)
+{
+  for (auto &plugin : plugins)
+  {
+    plugin->preCapture(p);
+  }
+}
+
+void RuntimePluginLoader::postCapture(RAJA::util::PluginContext p)
```
MrBurmark, line 52: pass by reference?

docs/sphinx/user_guide/feature/plugins.rst
JohnHynes: I'm not sure I see any missing the's, where exactly?

src/KokkosPluginLoader.cpp
```diff
@@ -0,0 +1,145 @@
...
+  {
+    func(0, kokkos_interface_version, 0, nullptr);
+  }
+}
+
+void KokkosPluginLoader::preLaunch(RAJA::util::PluginContext& p)
+{
+  for (auto &func : pre_functions)
+  {
+    func("", 0, &(p.kID));
```
DavidPoliakoff, line 64: Are we providing a blank name for all kernels, or am I misreading?

include/RAJA/util/PluginStrategy.hpp
JohnHynes: They definitely can be. It might cause some weird issues if some RAJA plugin messes with the kID unsigned long in the PluginContext, but I'm pretty sure each plugin gets its own copy anyways so it should be alright.

src/KokkosPluginLoader.cpp
JohnHynes: It is indeed!

DavidPoliakoff: That might be something to take a swing at later, is allowing for named kernels (though I'm sure that's already on the team's mind)

docs/sphinx/user_guide/feature/plugins.rst
MrBurmark: The plugin must have the following factory function

test/integration/CMakeLists.txt
```diff
@@ -29,4 +29,21 @@ endif()
 
 add_subdirectory(plugin)
 
-unset( PLUGIN_BACKENDS )
+if(NOT WIN32)
+raja_add_test(
+  NAME test-plugin-dynamic
+  SOURCES test_plugin_dynamic.cpp)
+
+add_library(dynamic_plugin SHARED plugin_for_test_dynamic.cpp)
```
davidbeckingsale, line 37: Can you replace this with blt_add_library (https://llnl-blt.readthedocs.io/en/develop/api/target.html#blt-add-library), and make it "depend" on hip. You should be able to remove the target_link_libraries line.

test/integration/CMakeLists.txt
```diff
@@ -29,4 +29,21 @@ endif()
...
+target_link_libraries(dynamic_plugin PRIVATE RAJA gtest)
+
+raja_add_test(
+  NAME test-plugin-kokkos
+  SOURCES test_plugin_kokkos.cpp)
+
+add_library(kokkos_plugin SHARED plugin_for_test_kokkos.cpp)
```
davidbeckingsale, line 44: Same comment as above.

src/KokkosPluginLoader.cpp
```diff
@@ -0,0 +1,145 @@
...
+  {
+    func();
+  }
+}
+
+// Initialize plugin from a shared object file specified by 'path'.
+void KokkosPluginLoader::initPlugin(const std::string &path)
+{
+  #ifndef _WIN32
+  void *plugin = dlopen(path.c_str(), RTLD_NOW | RTLD_GLOBAL);
```
MrBurmark, line 88: Does this get dlclose()ed somehwere?

src/RuntimePluginLoader.cpp
```diff
@@ -0,0 +1,148 @@
...
+    plugin->finalize();
+  }
+}
+
+
+// Initialize plugin from a shared object file specified by 'path'.
+void RuntimePluginLoader::initPlugin(const std::string &path)
+{
+  #ifndef _WIN32
+  void *plugin = dlopen(path.c_str(), RTLD_NOW | RTLD_GLOBAL);
```
MrBurmark, line 89: dlclose?

src/KokkosPluginLoader.cpp
JohnHynes: Whoopsie, don't believe so actually. Adding it in should still I think, since the functions are already grabbed and stored in the vectors.

MrBurmark: I think it should only be called once we no longer need the library.

src/KokkosPluginLoader.cpp
```diff
@@ -99,6 +99,8 @@ void KokkosPluginLoader::initPlugin(const std::string &path)
   getFunction<post_function>(plugin, post_functions, "kokkosp_end_parallel_for");
 
   getFunction<finalize_function>(plugin, finalize_functions, "kokkosp_finalize_library");
+
+  dlclose(plugin);
```
MrBurmark, line 103: I think this should be called after finalize to ensure the library isn't unloaded while its being used.

JohnHynes: Hmm, to do that we may have to add an array of pointers so there's something to dlclose. One thing to note is that finalize isn't called by default in RAJA, the user chooses to run it in their code. Maybe call dlclose in the destructor of KokkosPluginLoader? That might be a little volatile though. Thoughts?

include/RAJA/util/PluginStrategy.hpp
```diff
@@ -9,26 +9,30 @@
...
-
 class PluginStrategy
 {
   public:
     PluginStrategy();
 
     virtual ~PluginStrategy() = default;
 
-    virtual void preCapture(PluginContext p);
+    virtual void init(PluginOptions p);
```
MrBurmark, line 25: Should this be by reference too?

include/RAJA/util/RuntimePluginLoader.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+    void preCapture(RAJA::util::PluginContext& p);
+
+    void postCapture(RAJA::util::PluginContext& p);
+
+    void preLaunch(RAJA::util::PluginContext& p);
+
+    void postLaunch(RAJA::util::PluginContext& p);
+
+    void finalize();
```
MrBurmark, line 37: Might be nice to use override to detect signature mismatches.

include/RAJA/util/PluginStrategy.hpp
JohnHynes: We talked about that for a little bit, PluginOptions at the moment is basically just an std::string, which is certainly expensive to copy, however if they all share the same string one plugin could modify it and then another might receive the wrong information. The default init_plugins() command will put in an empty string by default, so this copy is pretty tiny. It's definitely possible to just pass a reference, but that's a design decision we can make

src/KokkosPluginLoader.cpp
MrBurmark: That makes sense, but do we delete the plugin objects?
Ya maybe its probably better not to close it if there isn't a safe place to do so.
Does it make sense to unload and remove dynamically loaded plugins after finalize?

include/RAJA/util/PluginStrategy.hpp
MrBurmark: const& then?

src/KokkosPluginLoader.cpp
JohnHynes: We do delete the plugin objects later on. The array of plugins in RuntimePluginLoader is implemented as: std::vector<std::unique_ptr<PluginStrategy>> plugins; The default destructors of vector and std::unique_ptr take care of it for us. Also, I do believe that's why we didn't have dlclose now that I remember it. It would indeed make sense to unload the plugins after finalize, might be worth adding in a 'plugins.clear()' at the end of finalize. Wouldn't stop the user from adding more plugins afterwards since the vector would still be there

include/RAJA/util/PluginStrategy.hpp
JohnHynes: @DavidPoliakoff The way we've implemented support for Kokkos Tools is by including the kID in RAJA's PluginContext. If we pass that as const, do you think it would interfere with any of the tools? I don't think it should but I'm not sure.

DavidPoliakoff: I'd recommend const for post but not pre. The pre-hooks often write to data, calling things const might break that flow

DavidPoliakoff: Further, though, I think most tools won't provide interesting data given empty names. The kID mostly just exists to speed up some "end" call for some tools

examples/plugin/timer-plugin.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include "RAJA/util/PluginStrategy.hpp"
+
+#include <iostream>
+#include <chrono>
+
+class TimerPlugin : public RAJA::util::PluginStrategy
+{
+public:
+  void preLaunch(RAJA::util::PluginContext& RAJA_UNUSED_ARG(p))
```
MrBurmark, line 16: override

examples/plugin/timer-plugin.cpp
```diff
@@ -0,0 +1,48 @@
...
+  {
+    start_time = std::chrono::steady_clock::now();
+  }
+
+  void postLaunch(const RAJA::util::PluginContext& p)
```
MrBurmark, line 21: overrride

examples/plugin/counter-plugin.cpp
```diff
@@ -13,18 +14,22 @@ class CounterPlugin :
   public RAJA::util::PluginStrategy
 {
   public:
-  void preCapture(RAJA::util::PluginContext p) {
+  void preCapture(const RAJA::util::PluginContext& p) {
```
MrBurmark, line 17: override

examples/plugin/counter-plugin.cpp
```diff
@@ -13,18 +14,22 @@ class CounterPlugin :
...
     if (p.platform == RAJA::Platform::host)
       std::cout << " [CounterPlugin]: Capturing host kernel for the " << ++host_capture_counter << " time!" << std::endl;
     else
       std::cout << " [CounterPlugin]: Capturing device kernel for the " << ++device_capture_counter << " time!" << std::endl;
   }
 
-  void preLaunch(RAJA::util::PluginContext p) {
+  void preLaunch(RAJA::util::PluginContext& p) {
```
MrBurmark, line 24: override

docs/sphinx/user_guide/feature/plugins.rst
```diff
@@ -0,0 +1,129 @@
...
+
+* ``void preCapture(PluginContext& p) {}`` - Will occur before capture of kernel/forall.
+
+* ``void postCapture(PluginContext& p) {}`` - Will occur after capture of kernel/forall.
+
+* ``void preLaunch(PluginContext& p) {}`` - Will occur before kernel/forall execution.
+
+* ``void postLaunch(PluginContext& p) {}`` - Will occur after kernel/forall execution.
+
+* ``void finalize() {}`` - runs on all plugins when the user makes a call to ``finalize_plugins``
```
MrBurmark, line 88: Function signatures.
Are these just signature or are they examples to the user? Should they have override or not?

test/integration/plugin/plugin_to_test.cpp
```diff
@@ -16,7 +16,7 @@ class CounterPlugin :
   public RAJA::util::PluginStrategy
 {
   public:
-  void preCapture(RAJA::util::PluginContext p) {
+  void preCapture(const RAJA::util::PluginContext& p) {
```
MrBurmark, line 19: override

test/integration/plugin_for_test_dynamic.cpp
```diff
@@ -0,0 +1,23 @@
...
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+#include "RAJA/util/PluginStrategy.hpp"
+
+#include <exception>
+
+class ExceptionPlugin :
+  public RAJA::util::PluginStrategy
+{
+  public:
+  void preLaunch(RAJA::util::PluginContext& RAJA_UNUSED_ARG(p)) {
```
MrBurmark, line 15: override

include/RAJA/util/PluginStrategy.hpp
MrBurmark: This is a little confusing to me, are the Kokkos Tools setting kID in the pre hook?

src/KokkosPluginLoader.cpp
MrBurmark: That makes sense, I missed the unique_ptr.
Does the plugin store the dlopened library and dlcose it when its destroyed?

include/RAJA/util/PluginStrategy.hpp
davidbeckingsale: Yes, they are.

DavidPoliakoff: Right, the typical pattern is that the begin hook does some start event, and then adds some context into a data structure there. They overwrite the contents of kID, and then on the end event they use kID to look up that earlier context. This saves a string lookup, which is of debatable utility in my view, but it keeps tool people from yelling at me, which is the most important thing

davidbeckingsale:
Here's the error from the HIP build:

`/home/raja/include/RAJA/util/KokkosPluginLoader.hpp:35:10: warning: 'finalize' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]`

docs/sphinx/user_guide/feature/plugins.rst
JohnHynes: I'm not entirely sure if they need override or not, the idea there being that the user could just copy paste those in if they needed to. Whatever we end up settling on I'll update those to.

JohnHynes: I would think that they should indeed be overrided

include/RAJA/util/PluginStrategy.hpp
MrBurmark: Ok that makes sense, 
It seems somewhat dangerous to have an unprotected kID in the context. Other plugins could use it to speedup their own post hooks. This also leaves the platform member exposed.
Is there a design that doesn't expose these variables, or only exposes them to the kokkos plugins?

MrBurmark: Would it work to make kID be a private mutable member of the context and give the context a private const friend accessor for KokkosPluginLoader?
That way the signatures of the plugins could all be const ref and only the kokkos plugins could access kID.

JohnHynes: Sounds like a good idea, I'm trying to get it working but it seems to be giving me a little trouble. I'll see what I can do!

include/RAJA/util/plugins.hpp
```diff
@@ -48,7 +50,7 @@ callPostCapturePlugins(PluginContext p) noexcept
 
 RAJA_INLINE
 void
-callPreLaunchPlugins(PluginContext p) noexcept
+callPreLaunchPlugins(PluginContext& p)
```
MrBurmark, line 53: Should this take by const& now?

.travis.yml
```diff
@@ -80,7 +75,7 @@ matrix:
       - COMPILER=g++
       - IMG=hip
       - HCC_AMDGPU_TARGET=gfx900
-      - CMAKE_EXTRA_FLAGS="-DENABLE_HIP=On -DENABLE_OPENMP=Off -DENABLE_CUDA=Off -DENABLE_WARNINGS_AS_ERRORS=Off"
+      - CMAKE_EXTRA_FLAGS="-DENABLE_HIP=On -DENABLE_OPENMP=Off -DENABLE_CUDA=Off -DENABLE_WARNINGS_AS_ERRORS=Off -DHIP_HIPCC_FLAGS="-fPIC""
```
davidbeckingsale, line 78: It doesn't look like the flag is being correctly set because of the nested quotes:
```suggestion
      - CMAKE_EXTRA_FLAGS="-DENABLE_HIP=On -DENABLE_OPENMP=Off -DENABLE_CUDA=Off -DENABLE_WARNINGS_AS_ERRORS=Off -DHIP_HIPCC_FLAGS=-fPIC"
```

JohnHynes:
@rhornung67 We've fixed the hip build!
We added -DHIP_HIPCC_FLAGS=-fPIC into .travis.yml  and that allows for the dynamic libraries to be compiled and pass their tests!
@davidbeckingsale and I think that we're ready to merge if nobody has any more comments or concerns.

MrBurmark:
👍 seconded

PR #838, opened 2020-06-25
Improve handling of external libraries

adayton1:
# Summary

- This PR is a bugfix
  - Modifies CMakeLists.txt to import a CAMP target, to use an existing target for CUB or to create one, to simplify options for external camp and CUB, and to remove unused CHAI options.
  
# Design review
- This may not be the right way to set things up, so I am certainly open to feedback. I think the changes for CAMP should be fine, but I'm not sure about the changes for CUB.


adayton1:
To get a better feel for what I am trying to do, see https://github.com/LLNL/CARE/blob/submodules3/tpl/CMakeLists.txt.

trws:
It looks like you're planning to use RAJA as a subdirectory in another cmake project.  I'd probably be fine with supporting that as an option this way with cub and camp and so-forth detected as pre-existing targets, but is there a reason you've removed the option to use external source directories through cmake options?

adayton1:
I removed the explicit ENABLE_EXTERNAL_CUB since it seemed to be redundant. If you have already specified CUB_DIR, then it's almost guaranteed that you want to use an external CUB. I changed EXTERNAL_CAMP_SOURCE_DIR to camp_DIR so that you can simply call find_package(camp) since the camp install provides campConfig.cmake.

trws:
That is how camp works, but it isn't redundant.  The find_package check will only find installed copies, which you could already do by setting camp_DIR or setting the cmake search path as normal without changing anything, the extra variable is there to select an uninstalled source version of camp for things like development and testing.

adayton1:
OK, now I get how things were working with camp before. I've reverted those changes. I'm looking more closely at CUB now.

Issue #847, opened 2020-07-14
Extra parameter in host-config file


adrienbernede:
Host config files are defining "RAJA_COMPILER".

https://github.com/LLNL/RAJA/blob/e1f3de7f71637383665379c97802481bfce9318f/host-configs/lc-builds/toss3/clang_X.cmake#L8

However, when tracking this variable usage, it seems it is if fact also definde in 

https://github.com/LLNL/RAJA/blob/e1f3de7f71637383665379c97802481bfce9318f/cmake/SetupCompilers.cmake#L29

And in the end is apparently not used since a different logic is used in `include/RAJA/config.hpp.in`:

https://github.com/LLNL/RAJA/blob/e1f3de7f71637383665379c97802481bfce9318f/include/RAJA/config.hpp.in#L127

Am I missing something?

rhornung67:
No longer an issue

PR #848, opened 2020-07-15
Feature/uberenv

adrienbernede:
# Summary

This PR is the beginning of a refactoring of host-config files in RAJA.

In particular:
- It introduces Uberenv to drive Spack, installing the dependencies for the required spec. 
- A _local_ copy of the RAJA Spack package has been modified with an additional phase that generate a host-config file for the configuration demanded.

Using uberenv thus allow to generated a host-config file through Spack.

Documentation of the process was added in the "developer guide" section.

# Context

The targeted process was vetted in Umpire:
- [This PR] Being able to generate the host-config file for any configuration instead of crafted them manually.
- [This PR] Store the most relevant ones, especially for CI purpose
- [Next PR?] Add CI on Gitlab making use of store host-config files.

Note: The only reason we don’t systematically generate the host-config files is because it’s not efficient (requires downloading Spack, and potentially installing some dependencies, if not provided externally: CMake typically).

# Discussion

Based on https://github.com/LLNL/RAJA/issues/847, I did not port the definition of `RAJA_COMPILER` variable.
As stated in https://github.com/LLNL/Umpire/issues/367, not all the configuration is embedded in the host-config file.

# What’s next?

This PR can be reviewed as-is. It has no impact on RAJA development workflow yet. It introduces Uberenv, and 2 host-config files were produces and version-controlled: one for lassen and one for quartz.

Uberenv

[Uberenv.pdf](https://github.com/LLNL/RAJA/files/4928543/Uberenv.pdf)


davidbeckingsale:
@artv3 @rhornung67 - can you give this a quick look? It doesn't remove or break any existing functionality.

scripts/uberenv/project.json
```diff
@@ -0,0 +1,9 @@
+{
+"package_name" : "raja",
+"package_version" : "develop",
+"package_final_phase" : "hostconfig",
+"package_source_dir" : "../..",
+"spack_url": "https://github.com/spack/spack",
+"spack_commit": "bc53bb9b7cbf250c98cfe77d334ed30d6b958c21",
```
adrienbernede, line 7: @rhornung67 
Uberenv is configured use a specific commit in Spack for the sake of reproducibility.
I used the same commit as in other projects.

The spack reference should be updated when:
- We want to keep up with spack develop or new release (bug fix, improvements).
- Uberenv is updated and relies on a newer version (new feature) in Spack.
- RAJA package changed (for example, the local copy of RAJA package should be pushed on Spack).

For that last case in particular, the local copy of the package should only be present if it is newer than the Spack version. It is then useful to test new changes in the spack package before pushing them, just like in this PR.

adrienbernede:
I think this is good to go then!

davidbeckingsale:
@LLNL/raja-core - I'm going to merge this PR once CI is done. It only adds new functionality and doesn't impact any existing code/scripts/tests

adrienbernede:
@davidbeckingsale Can you merge this? I need to push a tiny fix to the RAJA local package. I would rather push it in #856.

PR #858, opened 2020-07-20
Feature/burmark1/plugin2raja (take 2)

MrBurmark:
This is a copy of #841 which I inadvertently merged. See the previous PR for additional comments and discussion.

# Split plugins into those run at capture and those run at launch

This is a breaking change for any plugins (CHAI) that already exist as it renames the plugin registry.

The main changes are in forall and kernel where the plugins are invoked.

- This PR is a feature
- It does the following:
  - Extends the Plugins with two extra methods and makes all methods optional
    - Methods called at Capture (before the loop is executed)
      - before the loop body is copied
      - after the loop body is copied 
    - Methods called at Launch
      - before the loop is executed
      - after the loop is executed
  - Split plugins help support work groups that copy the loop bodies and execute the loops in separate regions of code called at different times
  - Extend plugin tests to use different overloads of forall interface with different policies.

I ran the batched matrix multiply example to see if moving the location of force inline changed the runtime, but it seemed about the same with develop and this branch.


include/RAJA/util/plugins.hpp
```diff
@@ -14,11 +14,42 @@
...
 namespace util {
 
+template <typename T>
+auto trigger_updates_before(T&& item) -> typename std::remove_reference<T>::type
+{
+  return item;
+}
+
+
+inline
```
MrBurmark, line 24: @mdavis36 mentioned using RAJA_INLINE here in #841

MrBurmark:
Ok I've made all the requested changes and the tests now cover forall and kernel with different backends.
I will create a Chai PR with updates to its usage of the RAJA plugin interface once this is merged.
Can I get a review @davidbeckingsale @trws @artv3 @rhornung67 @mdavis36?

PR #894, opened 2020-09-08
[WIP] Fixing omp-target resource issues.

mdavis36:
Fixes compilation issues with omp-target in the latest RAJA release.

This is currently a work in progress.


include/RAJA/util/resource.hpp
```diff
@@ -36,7 +37,7 @@ namespace RAJA
 
   template<typename e>
   struct get_resource{
-    using type = camp::resources::Host;
+    using type = Host;
```
artv3, line 40: Could we keep the full namespace? I ran into namespace issues when building a large application with RAJA.

mdavis36:
@rhornung67 @MrBurmark @rchen20 @rchen20 @trws @ajkunen 

A bit of a status update on this fix to get omp-target builds to a state where they compile with this release.

The initial problem that was referenced in the most recent comment of #372 has been fixed. However currently we are not in a state where we can compile with omp-target. There are 4 primary culprits, 2 examples and 2 tests:

### Examples:
- target-kernel `omp-target-kernel.cpp` 
- omp-target-ltimes `omp-target-ltimes.cpp`

Both of these return :
`raja/include/RAJA/pattern/kernel/Lambda.hpp:153:3: error: no matching function for call to object of type 'const tuple_element_t<0L, camp::tuple<  "lambda's file location here"  > >'`
The source files for these examples haven't really been touched in 2 years and were committed with a "WIP" message so I'm not sure if either of these have ever worked properly with an omp-target build. 

### Tests:
- test-omp-target-kernel.exe `tests/old-tests/unit/omp-target/kernel.cpp`

This test used to build before commit 98a3dc9 "add kernel statement sanity test" which now causes this test to throw the static assert:
`Executable statement with no enclosed statements, this is almost certainly a bug`

- test-plugin-kernel-OpenMPTarget.exe `tests/integration/plugin/tests/test-plugin-kernel.hpp`

This test was added as of b36cb44 and since has not compiled with omp-target. This test also throws a similar error to the examples:
`raja/include/RAJA/pattern/kernel/Lambda.hpp:153:3: error: no matching function for call to object of type 'const tuple_element_t<0L, camp::tuple<PluginTestCallable> >' (aka 'const PluginTestCallable')`

### Moving Forward
I can post the full error log below if anyone has any immediate ideas on how to fix the issues above, my current theory is that the lambda/tuple data types are getting messed up as they are passed through the omp-target kernel code flow, but I'm not entirely sure.

However, in the interest of getting a buildable solution for the 0.12.1 release I suggest we may want to turn off these specific examples and tests for omp-target and continue to chase these issues for a fix in a later release.

I am currently building successfully with these tests/examples disabled using the `blueos_clang-ibm-2019.10.03_omptarget.sh` script. `blueos_xl-202.0.6.26_omptarget.sh` appears to be building fine but it is unbelievably slow and I have yet to get the end of a build process.

include/RAJA/util/resource.hpp
mdavis36: Fixed 👍

rhornung67:
Mike,

Thank you for digging into this and updating everyone on the status. We should plan on meeting soon to discuss the remaining issues and see if we can arrive at a plan to resolve them. At this point, I think our priority is to get the v0.12.1 release out today, if possible, since several apps need several fixes to move forward. So, disabling the tests and examples needed to get a successful openmp-target build seems reasonable for now.

Does everyone agree?  If so, how should we note the current state of this in the release notes?

--Rich

test/functional/forall/reduce-multiple-segment/tests/test-forall-segment-multiple-ReduceMin.hpp
```diff
@@ -50,7 +50,7 @@ void ForallReduceMinMultipleTestImpl(IDX_TYPE first,
 
   DATA_TYPE current_min = default_val;
 
-  RAJA::ReduceMin<REDUCE_POLICY, DATA_TYPE> min0;
+  RAJA::ReduceMin<REDUCE_POLICY, DATA_TYPE> min0(0);
```
MrBurmark, line 53: are these tests to test that default reduction arguments work?

MrBurmark:
Thanks Mike, very thorough analysis.

That sounds good to me, we should add an issue to take another look at these as well.

test/functional/forall/reduce-multiple-segment/tests/test-forall-segment-multiple-ReduceMin.hpp
mdavis36: This is required as the default constructor for omp-target Reduce_Data has been deleted.

rchen20:
After digging into the tests/old-tests/unit/omp-target/kernel.cpp error, I found that the RAJA::kernel implementation of omp_target_parallel_for_exec is not able to find the RangeSegments being passed in:

error: static_assert failed "Not all segments have been used in a loop: Either add more loop statements, or use the Lambda<N, Args..> form"

It might be best to turn this test off and fix it after the v0.12.1 release as has been suggested.

-Robert

test/functional/forall/reduce-multiple-segment/tests/test-forall-segment-multiple-ReduceMin.hpp
rhornung67: To clarify, omp-target reductions have a different interface than the others? If this is the case, please add an issue so that we can resolve this inconsistency.

MrBurmark: Can you also add a comment to the constructors in the test to indicate which should be default.

test/functional/forall/reduce-multiple-segment/tests/test-forall-segment-multiple-ReduceSum.hpp
```diff
@@ -39,6 +39,8 @@ void ForallReduceSumMultipleStaggeredTestImpl(IDX_TYPE first,
   working_res.memcpy(working_array, test_array, sizeof(DATA_TYPE) * last);
 
 
+  // Workaround for broken omp-target reduction interface.
+  // This should be `sumX;` not `sumX(0);`
```
MrBurmark, line 43: Should this be moved down?

MrBurmark (approved):

rchen20 (approved):

rhornung67 (approved):

mcordery:
As an additional comment to this, and this isn't something for you guys but is more on the BLT team, there's an issue with BLTs cmake system overriding OpenMP flags set by the user (at least for the compiler I'm using). Issue is here
https://github.com/LLNL/blt/issues/398

PR #903, opened 2020-10-01
Replacing spack_configs by submodule

adrienbernede:
# Summary

This pull request moves Spack configuration files used by Uberenv in a submodule.
The goal is to share those configuration files between projects, rather than duplicate them.

Spack configuration files will then be stored in https://github.com/LLNL/radiuss-spack-configs.

Linked to https://github.com/LLNL/Umpire/pull/453, https://github.com/LLNL/CHAI/pull/157.


adrienbernede:
LGTM

trws:
Why this rather than uberenv itself, and possibly these along with it, being a submodule?

adrienbernede:
So you mean Uberenv as a submodule and host-config as a submodule of Uberenv? I have to say I never thought about it that way.

Making Uberenv a submodule was kind of blocking, or will require a fork. But will nested submodule, I still won’t be able to control the host-config version... right?

adrienbernede:
If however you meant 2 submodules side by side, that’s the roadmap, yes.

davidbeckingsale:
To add to what Adrien said, uberenv (currently) needs to contain project-specific configuration information so using it as a submodule would require each project to maintain an uberenv fork. Nested submodules could work, with everyone having a forked uberenv pointing at a shared "configuration" module for the compilers/packages.

adrienbernede:
I’m no submodule master, but wouldn’t a uberenv submodule version be tight down to a specific host-config version?
2 drawbacks:
- pushing a change in host-config would require updating uberenv as well.
- It wouldn’t be easy to have a project using a different set of host-config... (I know that’s not the intended case here).

trws:
Ok, that makes sense.  I had assumed uberenv was generic and would take configs from the side.

davidbeckingsale (approved):

davidbeckingsale:
LGTM

davidbeckingsale:
@rhornung67 - this is the "shared" configurations. We now have the same compilers.yaml and packages.yaml for Umpire, RAJA and CHAI. I will merge this PR, then put up one where I add all the compiler configurations we are currently testing in Umpire.

PR #908, opened 2020-10-14
MultiView CHAI Integration

rchen20:
# Summary

- This PR is an update:
  - Modifies MultiView to allow better CHAI integration. The compiler has difficulty deducing the non-const version of the array-of-pointers in MultiView when a CHAI::ManagedArray is the set of pointers. Allowing a custom non-const type to be passed in as a template parameter resolves this, while the default remains the same cv-removed type as before. Following this PR, there will be a corresponding CHAI PR uses this ability.


davidbeckingsale (approved):

rhornung67 (approved):
I think this looks good.

PR #910, opened 2020-10-14
make common iterable and view types trivially copyable

trws:
This PR makes View, TypedRange[Stride]Segment, and [strided_]numeric_iterator trivially copyable.  It may be worth adding a check into a test file somewhere to check the type trait to make sure this is true all the time, but for now this should solve many of the warnings @rhornung67 was getting with target offload.

### DIFF
--- a/include/RAJA/index/RangeSegment.hpp
+++ b/include/RAJA/index/RangeSegment.hpp
@@ -102,28 +102,16 @@ struct TypedRangeSegment {
   RAJA_HOST_DEVICE TypedRangeSegment() = delete;
 
   //! move constructor
-  RAJA_HOST_DEVICE constexpr TypedRangeSegment(TypedRangeSegment&& o)
-      : m_begin(std::move(o.m_begin)), m_end(std::move(o.m_end))
-  {
-  }
+  constexpr TypedRangeSegment(TypedRangeSegment&& o) = default;
 
   //! copy constructor
-  RAJA_HOST_DEVICE constexpr TypedRangeSegment(TypedRangeSegment const& o)
-      : m_begin(o.m_begin), m_end(o.m_end)
-  {
-  }
+  constexpr TypedRangeSegment(TypedRangeSegment const& o) = default;
 
   //! copy assignment
-  RAJA_HOST_DEVICE RAJA_INLINE TypedRangeSegment& operator=(
-      TypedRangeSegment const& o)
-  {
-    m_begin = o.m_begin;
-    m_end = o.m_end;
-    return *this;
-  }
+  RAJA_INLINE TypedRangeSegment& operator=(TypedRangeSegment const& o) = default;
 
   //! destructor
-  RAJA_HOST_DEVICE RAJA_INLINE ~TypedRangeSegment() {}
+  RAJA_INLINE ~TypedRangeSegment() = default;
 
   //! swap one TypedRangeSegment with another
   /*!
@@ -301,34 +289,19 @@ struct TypedRangeStrideSegment {
   }
 
   //! disable compiler generated constructor
-  RAJA_HOST_DEVICE TypedRangeStrideSegment() = delete;
+  TypedRangeStrideSegment() = delete;
 
   //! move constructor
-  RAJA_HOST_DEVICE TypedRangeStrideSegment(TypedRangeStrideSegment&& o)
-      : m_begin(std::move(o.m_begin)),
-        m_end(std::move(o.m_end)),
-        m_size(std::move(o.m_size))
-  {
-  }
+  TypedRangeStrideSegment(TypedRangeStrideSegment&& o) = default;
 
   //! copy constructor
-  RAJA_HOST_DEVICE TypedRangeStrideSegment(TypedRangeStrideSegment const& o)
-      : m_begin(o.m_begin), m_end(o.m_end), m_size(o.m_size)
-  {
-  }
+  TypedRangeStrideSegment(TypedRangeStrideSegment const& o) = default;
 
   //! copy assignment
-  RAJA_HOST_DEVICE TypedRangeStrideSegment& operator=(
-      TypedRangeStrideSegment const& o)
-  {
-    m_begin = o.m_begin;
-    m_end = o.m_end;
-    m_size = o.m_size;
-    return *this;
-  }
+  TypedRangeStrideSegment& operator=(TypedRangeStrideSegment const& o) = default;
 
   //! destructor
-  RAJA_HOST_DEVICE ~TypedRangeStrideSegment() {}
+  ~TypedRangeStrideSegment() = default;
 
   //! swap one TypedRangeStrideSegment with another
   /*!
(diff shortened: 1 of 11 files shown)


rhornung67 (approved):
I tested this with the PerfSuite OpenMP target and CUDA. The only warnings we get now are from lambdas being non-trivially copyable (even in kernels just capturing scalars and pointers -- no views, etc.), which is a good improvement. I also ran the RAJA tests with HIP. Only one sort test failed, but I don't think that is due to these changes.

Thanks @trws 🥇

trws:
Glad to hear it!  Any chance you'd point me to one of the lambdas that's popping a warning? I have a feeling something in the kernel machinery, probably something in LoopData, is the culprit.

rhornung67:
> Glad to hear it! Any chance you'd point me to one of the lambdas that's popping a warning? I have a feeling something in the kernel machinery, probably something in LoopData, is the culprit.

@trws I mis-spoke earlier. Looking at this more closely, there are five Perf Suite kernels that generate clang10 compiler warnings (in src/basic: INIT_VIEW1D_OFFSET, REDUCE3_INT, TRAP_INT, FIRST_MIN -- and in src/stream: DOT). All use RAJA::forall (no warnings from RAJA::kernel). INIT_VIEW_OFFSET uses an offset view and the others use reductions (no views). All kernels that use RAJA::kernel use default views. So it looks like the culprits are reductions and offset layout.

I have a branch that uses your branch here: https://github.com/LLNL/RAJAPerf/tree/opmtarget-debug-v2

There is a build script in the branch that shows the CMake options that you can run by typing
./scripts/lc-builds/blueos_clang10.0.0_omptarget.sh  in the top-level Perf Suite directory.

trws:
That makes sense unfortunately, some of the reductions can’t be 
trivially copyable IIRC, but I’ll take a look.  The offset view I 
probably just missed when I was going through them.

include/RAJA/index/RangeSegment.hpp
```diff
@@ -102,28 +102,16 @@ struct TypedRangeSegment {
   RAJA_HOST_DEVICE TypedRangeSegment() = delete;
 
   //! move constructor
-  RAJA_HOST_DEVICE constexpr TypedRangeSegment(TypedRangeSegment&& o)
-      : m_begin(std::move(o.m_begin)), m_end(std::move(o.m_end))
-  {
-  }
+  constexpr TypedRangeSegment(TypedRangeSegment&& o) = default;
```
rhornung67, line 105: @trws for defaulted methods, should we remove argument names to avoid compiler warnings?

rhornung67 (approved):
@trws minor question about named args in defaulted methods. Otherwise, this looks good to merge.

include/RAJA/index/RangeSegment.hpp
trws: There shouldn't be any warnings from them, since there's no body, but if you prefer it stylistically I have no issue with removing the names, they don't really add anything at the moment.

rhornung67: I'm fine either way. I wasn't sure if there were potential warnings lurking with named args for some compilers. Actually, we have named args in some places and others without. I'll go through and remove the names so everything is consistent.  Are you good with that?

rhornung67:
@trws by removing constexpr are you following the logic of this: https://stackoverflow.com/questions/55182242/declaring-defaulted-assignment-operator-as-constexpr-which-compiler-is-right

trws:
That’s a good reason, but in this case I did it because in C++11 mode 
it’s not valid for it to be necessarily, and causes at least a warning 
that the methods would not be const-qualified by default in 14, which 
was causing the build to fail on EDG-based compilers.

rhornung67:
@trws good detective work!  Hopefully, your changes will get us past this hurdle.

rhornung67:
> @trws good detective work! Hopefully, your changes will get us past this hurdle.

Oh Crap.

trws:
No kidding, I’m digging into it but clearly I tripped something up 
pretty badly.

trws:
Ah, so… As annoying as this is, it’s actually a good thing.  The 
static assert I added is catching cases where the view is not trivially 
copyable, I need to figure out *why* it isn’t trivially copyable on 
these configurations, but I’d rather this than think we had it fixed 
and being wrong.

trws:
I'm seriously tempted to say take just 9b1b336 from this for now, and leave the rest for later.  There would be warnings on some compilers, but I'm having a crazy time trying to figure out what in the layout/idx changes is causing the issues we're seeing.

rhornung67:
> I'm seriously tempted to say take just [9b1b336](https://github.com/LLNL/RAJA/commit/9b1b336299fa43de1a4f7bb3d487f298d2a1733c) from this for now, and leave the rest for later. There would be warnings on some compilers, but I'm having a crazy time trying to figure out what in the layout/idx changes is causing the issues we're seeing.

I'll take that commit out for a spin. If it works. I say let's put this to bed for now.

rhornung67:
@trws I ran a bunch of tests with that commit and everything works. Can we also readd the trivially-copyable stuff for numeric_iterators and non-offset views and go with that. That all seemed to work last time I tried it and squahed a bunch of trivially-copyable warnings for openmp target.

trws:
I'll rebase this and try it with just that, if the intel 19 on LC works with the combo I'll put it up that way.

trws:
It seems to have passed for intel 19 just fine with all but the offset layout changes in there, which makes me wonder why that didn't work when I guarded it another way, but however it happened this was clean on LC.

rhornung67:
Thanks @trws I'll run this through the gauntlet (Perf Suite, CUDA, HIP, XL, etc.) and keep you posted.

rhornung67:
@trws I ran a bunch of tests with different compilers, erc. This looks good. I was about to push a change to turn off those static_asserts in the test, but you beat me to it. I pushed a change to the build scripts that Mike and I were using to test earlier so they don't make it into develop. If checks pass with this, we should merge it. I'll work on finishing up a release.

trws:
Looks like an update to the radiuss-spack-configs submodule somehow borked it, the remote url is missing is what the git command claims in travis.

rhornung67:
> Looks like an update to the radiuss-spack-configs submodule somehow borked it, the remote url is missing is what the git command claims in travis.

That shouldn't be in your branch, should it? I thought that was added to develop very recently.

trws:
Hmm, it looks like the reply I thought I put in was lost in the ether somewhere, so apologies in advance if this ends up being a double-post.

It looks like the submodule snuck into the commit that otherwise just removes the build script options.  Hopefully we can just amend it and remove the submodule from there?

trws:
I rebased to pull the submodule out, and accidentally merged the last two commits into one, sorry about that @rhornung67, but at least this should be clean now.

trws:
Ok, with this all green, I'm for merging at this point.  @rhornung67, work for you?  I'll try to figure out how we can get the rest of the way as a separate step.

rhornung67:
@trws Looks good. I ran a bunch more tests this morning. Merge away!  

@mdavis36 after this is merged (into your branch), please make a PR to merge your branch into develop. Thank you.

PR #928, opened 2020-11-09
Extensions for RAJA Teams

ajkunen:
# Summary

This PR adds the following to RAJA Teams:
- Tiling
-  Generalized executors for nested loops, allows all permutations/combinations of 1,2,3-nested loops and orders for CUDA/HIP
- SIMD support for Teams
- Reduction support with example. 
- Global thread policies for GPUs


artv3:
Awesome! Can we get a unit test? :)

ajkunen:
@artv3  I thought you'd never ask!

rhornung67:
@ajkunen so tests for this would resemble more complex kernel functional tests that we have not yet started converting over to the newer test organization. Before you start putting tests together, let's talk about how they should be organized so we don't have to rework them later.

examples/raja-teams.cpp
```diff
@@ -157,23 +168,26 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
        RAJA::expt::Resources(RAJA::expt::Teams(N_tri), RAJA::expt::Threads(N_tri)),
        [=] RAJA_HOST_DEVICE(RAJA::expt::LaunchContext ctx) {
 
-         RAJA::expt::loop<teams_x>(ctx, RAJA::RangeSegment(0, N_tri), [&](int r) {
+         RAJA::expt::tile<teams_x>(ctx, 4, RAJA::RangeSegment(0, N_tri), [&](RAJA::RangeSegment const &r_tile) {
```
artv3, line 171: What does this 4 do?

include/RAJA/pattern/teams/teams_cuda.hpp
```diff
@@ -416,121 +322,120 @@ struct LoopExecute<cuda_thread_xyz_nested_direct, SEGMENT> {
...
+      TILE_T tile_size,
+      SEGMENT const &segment,
       BODY const &body)
   {
-    const int len1 = segment1.end() - segment1.begin();
-    const int len0 = segment0.end() - segment0.begin();
+
+    const int len = segment.end() - segment.begin();
+
+    for (int tx = internal::get_cuda_dim<DIM>(threadIdx) * tile_size;
```
artv3, line 367: Nice! Something we have been wanting to do add is the ability to block-out unused threads in the thread loops. For example when we have 3D thread block but we only want to run with threads in the x dimension. Is that something we can add in this PR? Or should we add in a following PR? The thought would just to add in in the nested loop policies. 

The idea here would just add additional if(threadIdx.y < 1) , if(threadIdx.z < 1) for the threaded loops.

artv3:
@rhornung67 @ajkunen @MrBurmark , could I trouble you guys with a review? Please see the ``raja-teams-tile.cpp`` example for usage of tiling with RAJA teams. 

An area of future work could be an interface for perfectly nested loop, I was writing out support for up to 3 levels, but maintaining will be hard to do.

include/RAJA/pattern/teams/teams_core.hpp
```diff
@@ -222,6 +224,25 @@ RAJA_HOST_DEVICE RAJA_INLINE void loop(CONTEXT const &ctx,
...
+                                                                    segment,
+                                                                    body);
+#else
+  LoopIdxExecute<typename POLICY_LIST::host_policy_t, SEGMENT>::exec(ctx,
+                                                                  segment,
+                                                                  body);
+#endif
+}
+
 template <typename POLICY_LIST,
```
MrBurmark, line 246: loop_idx?

artv3: that function returns the index within the tile, should we change the name?

trws: It does? It looks like it runs the loop?

include/RAJA/pattern/teams/teams_core.hpp
```diff
@@ -248,22 +269,135 @@ template <typename POLICY_LIST,
...
 #if defined(RAJA_DEVICE_CODE)
-  LoopExecute<typename POLICY_LIST::device_policy_t, SEGMENT>::exec(
+  LoopIdxExecute<typename POLICY_LIST::device_policy_t, SEGMENT>::exec(
       ctx, segment0, segment1, segment2, body);
 #else
-  LoopExecute<typename POLICY_LIST::host_policy_t, SEGMENT>::exec(
+  LoopIdxExecute<typename POLICY_LIST::host_policy_t, SEGMENT>::exec(
       ctx, segment0, segment1, segment2, body);
 #endif
 }
```
trws, line 286: This is a style thing, but perhaps make this more like:

```
#If defined(RAJA_DEVICE_CODE)
using pol = typename POLICY_LIST::device_policy_t;
#else 
using pol = typename POLICY_LIST::host_policy_t;
#endif
```

This helps make it clear that they're doing the exact same thing, but with different policies, rather than replicating everything.

include/RAJA/pattern/teams/teams_core.hpp
artv3: Ah yes, sorry I should be more clear, it does run the loop, the usage I had in mind was like this:

```
        RAJA::expt::loop_idx<loop_t>(ctx, r_tile, [&](int global_id, int loc_id) {

             }); // loop_idx
```
Where global_id is the the global index, and loc_id was the offset within the tile.

include/RAJA/pattern/teams/teams_core.hpp
trws: Actually, looking through more, maybe even do this once higher up in the file and re-use it, or have a helper metafunction for it? There seem to be a lot of places with approximately this pattern, so it could cut down on the preprocessor checks by a good bit.

include/RAJA/pattern/teams/teams_core.hpp
trws: Is that global_id of the tile, or global index in the dimension as a whole?  I ask partly because I'm not sure how useful the loc_id is if you don't have the tile_id as well, but the global index could be useful either way so I'm not even 100% sure which two I would want from this.

artv3: It is the global index of the of iteration space as a whole. Currently, doing tiling within teams uses two loops. A tile loop (which loops over tiles), and the inner loop which loops over the values within the tile, usage looks like this: 

```
RAJA::expt::tile_idx<teams_x>(ctx, TILE_SZ, RAJA::RangeSegment(0, N_tri), [&](RAJA::RangeSegment const &r_tile, int tile_id) {           
  
  RAJA::expt::loop_idx<loop_t>(ctx, r_tile, [&](int global_id, int loc_id) {  

             }); // loop_idx                                                                                                                     
         });  // tile_idx         
```
In the tile_idx function, we loop over the tiles, r_tile is the tile that gets generated by partitioning the iteration space (and fed into loop_idx), and tile_id is the tile count.  Loop_idx then loops over the entries in r_tile, global_id is the global index with respect to the global iteration space, and loc_id is the local offset.

artv3: It definitely depends on what is being coded up that determines what would be needed, these are just options to provide what may be needed. We also have variants that do not return the index (::tile<teams> , ::loop<loop_t>). I'm open to ideas, happy to get the discussion going.

trws: Ok, that makes more sense, thanks @artv3.  Is this also how the user expresses team/block and thread level parallelism or are these separate?

artv3: Yes this is definitely one way, here users could map teams to tiles and threads to iterations within the tile. Shared memory can be allocated between the tile and loops so threads can share data.

trws: What's the other way? I suppose I was asking mainly to see if this is something where we could avoid some api duplication.

artv3: The other way is to map teams/threads directly to loops, in finite element applications we have loops over elements, then inner loops over degree's of freedom inside an element ( No tiling necessary). 

In CUDA this would look like 

```
int e = blockIdx.x; 

     __shared__ double s_A[]; 

      int c = threadIdx.x 
      s_A[c] = .... 
```

Or using RAJA teams: 

```
RAJA::expt::loop<teams_x>(ctx, RAJA::RangeSegment(0, Elements), [&](int e) {

RAJA_SHARED double s_A[]

  RAJA::expt::loop<threads_x>(ctx, RAJA::RangeSegment(0, Ndofs), [&](int c) {
              s_A[c] = ... 
  }); 

});
```

The inner loops could be 1,2,3D and they would be doing a series of small matrix-matrix multiples.

include/RAJA/pattern/teams/teams_core.hpp
artv3: I believe the last commit addresses this now. Thanks!

examples/raja-teams-tile.cpp
```diff
@@ -0,0 +1,183 @@
...
+    RAJA::View<int, RAJA::Layout<2>> D(Ddat, N_tri, N_tri);
+
+    RAJA::expt::launch<launch_policy>(select_cpu_or_gpu,
+       RAJA::expt::Resources(RAJA::expt::Teams(N_tri), RAJA::expt::Threads(N_tri)),
+       [=] RAJA_HOST_DEVICE(RAJA::expt::LaunchContext ctx) {
+
+        const int TILE_SZ = 4;
+        RAJA::expt::tile_idx<teams_x>(ctx, TILE_SZ, RAJA::RangeSegment(0, N_tri), [&](RAJA::RangeSegment const &r_tile, int tile_id) {
+
+             RAJA::expt::loop_idx<loop_t>(ctx, r_tile, [&](int global_id, int loc_id) {
```
ajkunen, line 162: @artv3 do we want to call this loop_icount to be consistent with forall_icount?

rhornung67: Good suggestion 👍

artv3: Yes, that is a good idea.

include/RAJA/pattern/teams/teams_core.hpp
trws: Looks great!

examples/teams_matrix-multiply.cpp
```diff
@@ -0,0 +1,928 @@
...
+//----------------------------------------------------------------------------//
+#if defined(RAJA_ENABLE_CUDA)
+
+  std::cout << "\n Running CUDA tiled mat-mult with shared memory ...\n";
+
+  std::memset(C, 0, N*N * sizeof(double));
+
+  using seq_loop =  RAJA::expt::LoopPolicy<RAJA::loop_exec, RAJA::loop_exec>;
+
+  RAJA::expt::launch<launch_policy>(RAJA::expt::DEVICE,
```
artv3, line 699: Hi folks, I've pushed this teams_matrix-multiply.cpp example that presents a number of variants of matrix-multiply. This includes Tiling with shared memory. I'm still working on the documentation, but I think looks pretty cool! I also have a native CUDA version on top.

artv3:
@rhornung67 @ajkunen @trws @MrBurmark  @rchen20 , could I trouble you all for a review? This PR adds tiling capabilities as well as the `` teams_matrix-multiply.cpp `` example which presents variants of mat-mat multiply using teams, teams+tiling, teams+tiling+shared memory, I also added the shared memory mat-mat example in ``tut_matrix-multiply.cpp `` to compare with the teams variant. 

An area of future work will be an interface for perfectly nested loops.

include/RAJA/policy/cuda/kernel/For.hpp
```diff
@@ -162,12 +162,11 @@ struct CudaStatementExecutor<
 template <typename Data,
           camp::idx_t ArgumentId,
           int ThreadDim,
-          int MinThreads,
           typename... EnclosedStmts,
           typename Types>
 struct CudaStatementExecutor<
     Data,
-    statement::For<ArgumentId, RAJA::cuda_thread_xyz_loop<ThreadDim, MinThreads>, EnclosedStmts...>,
+    statement::For<ArgumentId, RAJA::cuda_thread_xyz_loop<ThreadDim>, EnclosedStmts...>,
```
MrBurmark, line 169: Why did MinThreads go away?

artv3: It looks like because we had a change of definition of the following struct: 

```
template<int dim, int min_threads>
struct cuda_thread_xyz_loop{};
```
to 
```
template<int ... dim>                                                                                                                                                                                         
struct cuda_block_xyz_loop{}; 
```

What do you think @ajkunen , should we have separate policies for team loops? This is a good discussion point.

artv3: Resolved, this change is okay.

examples/teams_matrix-multiply.cpp
```diff
@@ -0,0 +1,866 @@
...
+//printResult<double>(Cview, N);
+
+//----------------------------------------------------------------------------//
+
+  std::cout << "\n Running CUDA tiled mat-mult ...\n";
+
+  std::memset(C, 0, N*N * sizeof(double));
+
+  //
+  // This example takes the extend of the col and row loops and breaks
```
rchen20, line 484:
```suggestion
  // This example takes the extents of the col and row loops and breaks
```

examples/teams_matrix-multiply.cpp
```diff
@@ -0,0 +1,866 @@
...
+
+//----------------------------------------------------------------------------//
+
+  std::cout << "\n Running HIP tiled mat-mult ...\n";
+
+  std::memset(C, 0, N*N * sizeof(double));
+  hipErrchk(hipMemcpy( d_C, C, N * N * sizeof(double), hipMemcpyHostToDevice ));
+
+  //
+  // This example takes the extend of the col and row loops and breaks
```
rchen20, line 584:
```suggestion
  // This example takes the extents of the col and row loops and breaks
```

MrBurmark (approved):

rchen20 (approved):

include/RAJA/pattern/teams/teams_openmp.hpp
```diff
@@ -104,6 +104,76 @@ struct LoopExecute<omp_parallel_for_exec, SEGMENT> {
...
+      LaunchContext const RAJA_UNUSED_ARG(&ctx),
+      SEGMENT const &segment0,
+      SEGMENT const &segment1,
+      BODY const &body)
+  {
+
+    const int len1 = segment1.end() - segment1.begin();
+    const int len0 = segment0.end() - segment0.begin();
+
+#pragma omp parallel for RAJA_COLLAPSE(2)
```
rhornung67, line 139: Do we always want collapse here? Collapsing can improve or degrade performance depending on the loop lengths.

include/RAJA/pattern/teams/teams_openmp.hpp
```diff
@@ -156,6 +226,103 @@ struct LoopExecute<omp_parallel_nested_for_exec, SEGMENT> {
...
```
rhornung67, line 244: Same comment about collapse.

include/RAJA/pattern/teams/teams_openmp.hpp
```diff
@@ -104,6 +104,76 @@ struct LoopExecute<omp_parallel_for_exec, SEGMENT> {
...
+      SEGMENT const &segment1,
+      SEGMENT const &segment2,
+      BODY const &body)
+  {
+
+    const int len2 = segment2.end() - segment2.begin();
+    const int len1 = segment1.end() - segment1.begin();
+    const int len0 = segment0.end() - segment0.begin();
+
+#pragma omp parallel for RAJA_COLLAPSE(3)
```
rhornung67, line 163: Same comment about collapse.

include/RAJA/pattern/teams/teams_openmp.hpp
```diff
@@ -104,6 +104,76 @@ struct LoopExecute<omp_parallel_for_exec, SEGMENT> {
...
+  template <typename BODY>
+  static RAJA_INLINE RAJA_HOST_DEVICE void exec(
+      LaunchContext const RAJA_UNUSED_ARG(&ctx),
+      SEGMENT const &segment,
+      BODY const &body)
+  {
+
+    int len = segment.end() - segment.begin();
+#pragma omp parallel for
+    for (int i = 0; i < len; i++) {
```
rhornung67, line 122: Would it be better to run the loop from begin() to end() and avoid the offset calculation in the loop, then use a local counter variable for the second argument to body(); i.e., int ii = 0 outside of loop, then do ii++ in the loop? I am wondering whether the compiler would optimize better....

Same question in other places in the code.

include/RAJA/pattern/teams/teams_openmp.hpp
```diff
@@ -104,6 +104,76 @@ struct LoopExecute<omp_parallel_for_exec, SEGMENT> {
...
+
+    const int len2 = segment2.end() - segment2.begin();
+    const int len1 = segment1.end() - segment1.begin();
+    const int len0 = segment0.end() - segment0.begin();
+
+#pragma omp parallel for RAJA_COLLAPSE(3)
+    for (int k = 0; k < len2; k++) {
+      for (int j = 0; j < len1; j++) {
+        for (int i = 0; i < len0; i++) {
+          body(*(segment0.begin() + i),
```
rhornung67, line 167: Same question about the loop structure as above and below in other places....

include/RAJA/pattern/teams/teams_openmp.hpp
artv3: @rhornung67 , I think that would end up in a race condition for ii with omp. One option could be to make change it to 

```
    RAJA_EXTRACT_BED_IT(segment);                                                                                                                              
                                                                                                                                                               
#pragma omp parallel for                                                                                                                                       
    for (decltype(distance_it) i = 0; i < distance_it; ++i) {                                                                                                  
                                                                                                                                                               
      body(*(begin_it + i), i);                                                                                                                                
    } 
```
    
Which is more consistent with the rest of RAJA, would this work?

include/RAJA/pattern/teams/teams_openmp.hpp
artv3: That is a fair point, one option could be to add collapse policies, but in general I'm not sure what the API for perfectly nested loops in teams should look like. At the moment users can express perfectly nested loops by nesting RAJA::loops like so: 

```
RAJA::loop<pol1>(Range1){
 RAJA::loop<pol2>(Range2){
  RAJA::loop<pol3>(Range3){
 }
 }
}
```

Or we can support an interface that takes in a policy (or policies) ? With multiple segments? We have some support for this at the moment, but super experimental at the moment.  

```
RAJA::loop<pol3>(Range1, Range2, Range3) {} 
```

rhornung67: @ajkunen or @trws do you guys have thoughts on how best to handle loop collapse as a user-specified choice? I don't think it is wise to always collapse the loops.

include/RAJA/pattern/teams/teams_openmp.hpp
rhornung67: That's a good point about a race condition since the declaration and initialization would occur inside an openmp parallel region. We should try a few things with different compilers to see how they impact performance. I was thinking in terms of trying to make the code as simple as possible for compilers to optimize.

trws: When I ran some of these numbers before, some of the compilers strongly preferred having an integer rather than iterator as their loop variable.  Others it didn't matter, but the version there worked well in general.  It wouldn't hurt to see what happens with another round of tests though, if we could get rid of that extra boiler-plate and directly use the iterators I'd be pretty happy about it, especially if the compilers did well with the range-based for loop (which wasn't supported in OpenMP yet when we did that refactor so it wasn't portable):

```c++
for(auto & i : segment) {
  body(i);
}
```

rhornung67: Yes, I wasn't clear in my original comment. I meant to do something like 

int ibeg = segment.begin();
int iend = segment.end();
for (int i = ibeg; i < iend; ++i) ...

trws: That's what the BED_IT one does though, right? To get from an iterator to what we need to give the body requires either an offset and dereference or `begin_it[I]`, which should usually be (but is not always) the same.

artv3:
Folks, I've been doing some performance testing, the tiling implementation here is kinda slow. Something seems off.... 

Doing the following
```
      RAJA::expt::tile<teams_y>                                                                                                                                                                               
        (ctx, CUDA_BLOCK_SIZE, row_range, [&] (RAJA::RangeSegment const &row_tile) {                                                                                                                          
          RAJA::expt::tile<teams_x>                                                                                                                                                                           
            (ctx, CUDA_BLOCK_SIZE, col_range, [&] (RAJA::RangeSegment const &col_tile) {                                                                                                                      
                                                                                                                                                                                                              
              RAJA::expt::loop<threads_y>(ctx, row_tile, [&] (int row) {                                                                                                                                      
                  RAJA::expt::loop<threads_x>(ctx, col_tile, [&] (int col) {                                                                                                                                  
                                                                                                                                                                                                             
                    double dot = 0.0;                                                                                                                                                                         
                    for (int k = 0; k < N; ++k) {                                                                                                                                                             
                      dot += Aview(row, k) * Bview(k, col);                                                                                                                                                   
                    }                                                                                                                                                                                         

                    Cview(row, col) = dot;                                                                                                                                                                    
                                                                                                                                                                                                              
                  });                                                                                                                                                                                         
                });                                                                                                                                                                                           
            });                                                                                                                                                                                               
        }); 
```

is much slower than doing this: 
```
   RAJA::expt::loop<global_thread_y>(ctx, row_range, [&] (int row) {                                                                                                                                          
       RAJA::expt::loop<global_thread_x>(ctx, col_range, [&] (int col) {                                                                                                                                      
                                                                                                                                                                                                              
          double dot = 0.0;                                                                                                                                                                                   
          for (int k = 0; k < N; ++k) {                                                                                                                                                                       
            dot += Aview(row, k) * Bview(k, col);                                                                                                                                                             
          }                                                                                                                                                                                                   
          Cview(row, col) = dot;                                                                                                                                                                              
      });                                                                                                                                                                                                     
    });         
```

Could there be overhead in generating the row and col tiles?

artv3:
I'll push up a benchmark tomorrow, any thoughts or suggestions would be greatly appreciated.

ajkunen:
Arturo,

Is there an code example in the repo that demonstrates this? I would be happy to take a look.

What are you seeing for register pressure?  I’m guessing the tiling loops may be greatly increasing register pressure… perhaps even spilling?

-Adam

include/RAJA/pattern/teams/teams_cuda.hpp
ajkunen: @artv3  just follow up here:  Do you mean "loop over x, but not over y or z"... like an exclusive x loop?
Like cuda_exclusive_thread_x_loop?
That would be relatively easy to implement.

artv3:
> Arturo, Is there an code example in the repo that demonstrates this? I would be happy to take a look. What are you seeing for register pressure? I’m guessing the tiling loops may be greatly increasing register pressure… perhaps even spilling?


@ajkunen thanks, I just pushed up mat-mat-perf.cpp that compares performance. I took a look via nvvp and there is increased register usage with tiling. The native version uses 32 registers per thread and achieves 100% of the theoretical occupancy while the tiling version (teams kernel #1) uses 48 registers per thread and achieves 62.5%  of the theoretical occupancy.

ajkunen:
Ok,

I noticed you were using RangeSegment, which defaults to a 64-bit signed int.
Just making one change:  Switched RangeSegment to TypedRangeSegment<int> so that indices fit into single 32-bit registers.

This clearly doesn’t solve everything, but it’s mostly an improvement. I’ll keep looking.

RangeSegment:

RAJA matrix multiplication example...
CUDA kernel 0
Delta t0-tf: 2504 milliseconds
CUDA kernel 1
Delta t0-tf: 1298 milliseconds
CUDA kernel 2
Delta t0-tf: 982 milliseconds

RAJA TEAM kernel 1
Delta t0-tf: 2721 milliseconds
RAJA TEAM kernel 2
Delta t0-tf: 3027 milliseconds
RAJA TEAM kernel 3
Delta t0-tf: 1646 milliseconds

RAJA Kernel 2
Delta t0-tf: 1413 milliseconds

TypeRangeSegment<int>:


RAJA matrix multiplication example...
CUDA kernel 0
Delta t0-tf: 2520 milliseconds
CUDA kernel 1
Delta t0-tf: 1299 milliseconds
CUDA kernel 2
Delta t0-tf: 982 milliseconds

RAJA TEAM kernel 1
Delta t0-tf: 2922 milliseconds
RAJA TEAM kernel 2
Delta t0-tf: 2202 milliseconds
RAJA TEAM kernel 3
Delta t0-tf: 1397 milliseconds

RAJA Kernel 2
Delta t0-tf: 1239 milliseconds

include/RAJA/pattern/teams/teams_openmp.hpp
trws: We're currently doing it with multiple segments and a policy in kernel, so if RAJA users are used to anything it would probably be that.  If we want to keep the single segment setup though, which does have non-trivial advantages, we could go more the TBB or SYCL style route and make a couple of multi-dimensional segments.  Something like:

```c++
RAJA::loop<pol>(RAJA::range2d(len1, len2), [](RAJA::range2idx idx){
// idx.x, idx.y
});
// or
RAJA::loop<pol>(RAJA::range2d(len1, len2), RAJA::unpack([](int I, int j){});
// or 
RAJA::loop<pol>(RAJA::range2d(len1, len2), RAJA::unpack([](int I, int j){}); // slightly harder
```

artv3:
@ajkunen, @MrBurmark  playing around with with  __launch_bounds__(maxthreads, minblocksperMP) method we can get the get the performance to be closer. The following compares performance with tiling only (no shared memory) 

CUDA kernel 1 with __launch_bounds__(512, 2)
Delta t0-tf: 1205 milliseconds

RAJA TEAM kernel 1   __launch_bounds__(512, 1)
Delta t0-tf: 1242 milliseconds

RAJA Kernel 1  __launch_bounds__(512, 1)
Delta t0-tf: 1237 milliseconds

The fastest impl uses global thread id with RAJA teams and the following parameters: 
RAJA TEAM kernel 1  __launch_bounds__(512, 2) 
Delta t0-tf: 1157 milliseconds

artv3:
> @ajkunen, @MrBurmark, by playing around with with **launch_bounds**(maxthreads, minblocksperMP) method we can get the get the performance to be closer. The following compares performance with tiling only (no shared memory)
> 
> CUDA kernel 1 with **launch_bounds**(512, 2)
> Delta t0-tf: 1205 milliseconds
> 
> RAJA TEAM kernel 1 **launch_bounds**(512, 1)
> Delta t0-tf: 1242 milliseconds
> 
> RAJA Kernel 1 **launch_bounds**(512, 1)
> Delta t0-tf: 1237 milliseconds
> 
> The fastest impl uses global thread id with RAJA teams and the following parameters:
> RAJA TEAM kernel 1 **launch_bounds**(512, 2)
> Delta t0-tf: 1157 milliseconds

Unfortunately this trick doesn't help the variant with shared memory, tinkering with it but the performance of tiling + shared memory looks like this, note its better if we don't modify launch bounds. 

RAJA matrix multiplication example...
CUDA kernel 2 
Delta t0-tf: 980 milliseconds

RAJA TEAM kernel 2  
Delta t0-tf: 1646 milliseconds

RAJA Kernel 2  
Delta t0-tf: 43313 milliseconds

artv3:
So good news we can get within 5% of the CUDA shared memory version with the following RAJA teams code: 

```
   RAJA::expt::loop<teams_y>(ctx, RAJA::RangeSegment(0,Ny), [&] (int by) {                                                                                                                               
     RAJA::expt::loop<teams_x>(ctx, RAJA::RangeSegment(0,Nx), [&] (int bx) {                                                                                                                             
                                                                                                                                                                                                         
         RAJA_TEAM_SHARED double As[CUDA_BLOCK_SIZE][CUDA_BLOCK_SIZE];                                                                                                                                   
         RAJA_TEAM_SHARED double Bs[CUDA_BLOCK_SIZE][CUDA_BLOCK_SIZE];                                                                                                                                   
         RAJA_TEAM_SHARED double Cs[CUDA_BLOCK_SIZE][CUDA_BLOCK_SIZE];                                                                                                                                   
                                                                                                                                                                                                         
         RAJA::expt::loop<threads_y>(ctx, RAJA::RangeSegment(0,CUDA_BLOCK_SIZE), [&] (int ty) {                                                                                                          
             RAJA::expt::loop<threads_x>(ctx, RAJA::RangeSegment(0,CUDA_BLOCK_SIZE), [&] (int tx) {                                                                                                      
                                                                                                                                                                                                         
                 Cs[ty][tx] = 0.0;                                                                                                                                                                       
               });                                                                                                                                                                                       
           });                                                                                                                                                                                           
                                                                                                                                                                                                         
                                                                                                                                                                                                         
         for (int k = 0; k < (CUDA_BLOCK_SIZE + N - 1)/CUDA_BLOCK_SIZE; k++) {                                                                                                                           
                                                                                                                                                                                                         
           RAJA::expt::loop<threads_y>(ctx, RAJA::RangeSegment(0,CUDA_BLOCK_SIZE), [&] (int ty) {                                                                                                        
               RAJA::expt::loop<threads_x>(ctx, RAJA::RangeSegment(0,CUDA_BLOCK_SIZE), [&] (int tx) {                                                                                                    
                                                                                                                                                                                                         
                   const int Row = by*CUDA_BLOCK_SIZE + ty;                                                                                                                                              
                   const int Col = bx*CUDA_BLOCK_SIZE + tx;                                                                                                                                              
                                                                                                                                                                                                         
                   if (k*CUDA_BLOCK_SIZE + tx < N && Row < N)                                                                                                                                            
                     As[ty][tx] = A[Row*N + k*CUDA_BLOCK_SIZE + tx];                                                                                                                                     
                   else                                                                                                                                                                                  
                     As[ty][tx] = 0.0;                                                                                                                                                                   
                                                                                                                                                                                                         
                   if (k*CUDA_BLOCK_SIZE + ty < N && Col < N)                                                                                                                                            
                     Bs[ty][tx] = B[(k*CUDA_BLOCK_SIZE + ty)*N + Col];                                                                                                                                   
                   else                                                                                                                                                                                  
                     Bs[ty][tx] = 0.0;                                                                                                                                                                   
                                                                                                                                                                                                         
                 });                                                                                                                                                                                     
             });                                                                                                                                                                                         
                                                                                                                                                                                                         
           ctx.teamSync(); 

           RAJA::expt::loop<threads_y>(ctx, RAJA::RangeSegment(0,CUDA_BLOCK_SIZE), [&] (int ty) {                                                                                                        
               RAJA::expt::loop<threads_x>(ctx, RAJA::RangeSegment(0,CUDA_BLOCK_SIZE), [&] (int tx) {                                                                                                    
                                                                                                                                                                                                         
                   for (int n = 0; n < CUDA_BLOCK_SIZE; ++n)                                                                                                                                             
                     Cs[ty][tx] += As[ty][n] * Bs[n][tx];                                                                                                                                                
                                                                                                                                                                                                         
                 });                                                                                                                                                                                     
             });                                                                                                                                                                                         
           ctx.teamSync();                                                                                                                                                                               
         }                                                                                                                                                                                               
                                                                                                                                                                                                         
                                                                                                                                                                                                         
         RAJA::expt::loop<threads_y>(ctx, RAJA::RangeSegment(0,CUDA_BLOCK_SIZE), [&] (int ty) {                                                                                                          
             RAJA::expt::loop<threads_x>(ctx, RAJA::RangeSegment(0,CUDA_BLOCK_SIZE), [&] (int tx) {                                                                                                      
                                                                                                                                                                                                         
                 int Row = by*CUDA_BLOCK_SIZE + ty;                                                                                                                                                      
                 int Col = bx*CUDA_BLOCK_SIZE + tx;                                                                                                                                                      
                                                                                                                                                                                                         
                 if (Row < N && Col < N)                                                                                                                                                                 
                   Cview(Row, Col) = Cs[ty][tx];                                                                                                                                                         
               });                                                                                                                                                                                       
           });    
           
      });                                                                                                                                                                                                 
   });   
```

Performance numbers: 
CUDA kernel 2 
Delta t0-tf: 980 milliseconds

RAJA TEAM kernel 2  
Delta t0-tf: 1027 milliseconds

Here I used the direct policies for both block and thread. It seems that there is a win by doing a more direct port.

artv3:
Okay I just pushed the driver that runs all the kernels. Here I removed the tiling methods from RAJA teams, they are still on for RAJA::kernel though. The RAJA teams version can be viewed as a direct port of the CUDA kernels, for teams I did not adjust launch bounds. 

```
RAJA matrix multiplication example...
CUDA kernel 0 
Delta t0-tf: 3855 milliseconds
CUDA kernel 1 
Delta t0-tf: 1393 milliseconds
CUDA kernel 2 
Delta t0-tf: 1147 milliseconds

RAJA Teams kernel 0  
Delta t0-tf: 3149 milliseconds
RAJA Teams kernel 1  
Delta t0-tf: 1407 milliseconds
RAJA Teams kernel 2  
Delta t0-tf: 1193 milliseconds

RAJA Kernel 1  
Delta t0-tf: 1525 milliseconds
RAJA Kernel 2  
Delta t0-tf: 49415 milliseconds
```

ajkunen:
Arturo, this is really good work.

A number of us have seen this before, where the _direct policies are much faster than _loop.  (Actually, this was one reason for having them in the first place).

Two questions:

  *   How much did it help to change the k-loop from a RAJA loop to a bare loop?
  *   I think you should also try changing the RangeSegment to TypedRangeSegment<int>.  I’ve seen this drastically reduce the register pressure, and can sometimes really impact the performance.  I wonder if we could get parity with CUDA then?!?  Worth a shot.


Perhaps we should have a “performance tuning guide” that details some of this stuff for each platform???
There’s a bunch of these kinds of things for CUDA… I’m sure they’ll be different for the AMD hardware (once we really start pounding on that).  There’s also a bunch of things for the CPU architectures (where 64-bit indices seem to be faster).


-Adam

rhornung67:
I think the idea of starting to put together a performance tuning guide is a good one. I am working on a PR that reorganizes our sphinx documentation so that we can more easily assemble and link in other docs, such as a dev guide (very basic start to this exists already), etc.

--Rich

examples/mat-mat-perf.cpp
```diff
@@ -160,14 +158,14 @@ __global__ void matMultKernel2(int N, double* C, double* A, double* B)
     __syncthreads();
 
     for (int n = 0; n < CUDA_BLOCK_SIZE; ++n)
-      Cs[threadIdx.y][threadIdx.x] += As[threadIdx.y][n] * Bs[n][threadIdx.x];
+      dot += As[threadIdx.y][n] * Bs[n][threadIdx.x];
```
artv3, line 161: I think this should be a different variant, we need to figure how to make this portable using RAJA.

artv3:
@ajkunen, there has been some additional pushes here where we have learned a bit more. @MrBurmark let's formalize the driver and results a bit more and we can touch base with the team on these results.

include/RAJA/pattern/teams/teams_openmp.hpp
artv3: I like the idea of multi-dimensional segments, it would fit well with the application I work on.

trws: That would be relatively easy to implement too, at least for a few bounded dimension counts.  For others we could make an N-dimensional version that hands out a tuple or similar, but it would be syntactically nastier.  Will have to give a bit of thought to some of the implementation details to get all the behavior you probably want, but may make life easier for a few patterns.

artv3:
Folks, I think some of the RAJA internals will need some editing for newer versions of ROCM. I've been trying to run our mat-mat example on AMD GPUS, but with rocm 10 I hit the following error: 

```
/RAJA/include/RAJA/policy/hip/policy.hpp:320:12: error: no viable conversion from returned value of type 'const typename __HIP_Coordinates<__HIP_ThreadIdx>::Y' to function return type 'decltype(d.x)' (aka 'const __HIP_Coordinates<__HIP_ThreadIdx>::X')
    return d.y;

```

artv3:
@ajkunen I've pushed improved docs with the ability to block out kernels with if/def for our tiling example. Unfortunately, I ran into problems with AMD GPUs, see note above.

artv3:
Folks, I pushed up fixes for AMD for RAJA teams. RAJA kernel..... will need some additional refactoring I've had to if/def out the runs in the mat-mat-perf file so I could get some data. 

But some here are some of the faster variations, the Device kernels are the native ones. 

``# 0``
Device kernel 0 
Maps CUDA blocks to rows 
Uses block stride loop for columns 
Delta t0-tf: 10670 milliseconds

	 result -- PASS

``# 2``
Device kernel 1 
Generates global 2D threads 
Each thread performs dot product 
Delta t0-tf: 4731 milliseconds

	 result -- PASS
	 
``# 6``
Device kernel 2 shared + register 
Performs multiplication by loading submatrices 
into shared memory tiles 
Solution is accumulated in register space (not portable) 
Delta t0-tf: 1861 milliseconds

	 result -- PASS
	 
	 
``# 8``
RAJA Teams kernel 0  
Maps CUDA blocks to rows 
Uses block stride loop for columns 
Delta t0-tf: 10781 milliseconds

	 result -- PASS

``# 10``
RAJA Teams kernel 1 w RAJA tiling  
Uses RAJA tiling to map blocks to tiles 
Maps threads to iterates within a tile 
Delta t0-tf: 5481 milliseconds

	 result -- PASS
	 
``# 12``
RAJA Teams kernel 1 global  
Generates global 2D threads, each thread performs a dot product 
Delta t0-tf: 5526 milliseconds

	 result -- PASS


``# 14``
RAJA Teams kernel 2 RAJA tiled shared 
Uses RAJA methods to break up the iteration space into tiles
Performs multiplication by loading submatrices into tiles of shmem 
Solution is accumulated in shared memory 
Delta t0-tf: 1971 milliseconds

	 result -- PASS

``# 18``
RAJA Teams kernel 2 direct tiling
Performs multiplication by loading submatrices into tiles 
Solution is accumulated in shared memory 
Delta t0-tf: 2025 milliseconds

	 result -- PASS

artv3:
I found a work around for getting kernel to run, loop_exec is missing in with the hip backend. But seq_exec on the gpu is defined the same way as we would define loop_exec so we may use that.  Looks like the non-portable implementation is doing the best run number 26. 

``# 22``
RAJA Kernel 1  
Uses RAJA methods to perform tiling
Delta t0-tf: 5544 milliseconds

	 result -- PASS

``# 24``
RAJA Kernel 2 shared 
Uses RAJA methods to break up the iteration space into tiles
Performs multiplication by loading submatrices into tiles 
Solution is accumulated in shared memory 
Delta t0-tf: 2707 milliseconds

	 result -- PASS

``# 26`` 
RAJA Kernel 2 + reg  
Uses RAJA methods to break up the iteration space into tiles
Performs multiplication within tiles 
Solution is accumulated in register 
Delta t0-tf: 1479 milliseconds

	 result -- PASS

ajkunen:
This is really good work @artv3 

I tried out changing TypedRangeSegment to see what would happen.
I tried changing the members to be values.
I also tried modifying the constructor to not have a trinary operator in it.

Neither of these changes made any difference in performance or register usage for any of the variants.

rhornung67:
> This is really good work @artv3
> 
> I tried out changing TypedRangeSegment to see what would happen.
> I tried changing the members to be values.
> I also tried modifying the constructor to not have a trinary operator in it.
> 
> Neither of these changes made any difference in performance or register usage for any of the variants.

Hmmmm. That's frustrating. What else could it be?

ajkunen:
I think it's probably time to reorganize the files to match the rest of RAJA
RAJA/pattern/teams should have non-policy stuff.
RAJA/policy/  should have the policy implementations (like RAJA/policy/cuda/teams, RAJA/policy/sequential/teams, etc.)

artv3:
> I think it's probably time to reorganize the files to match the rest of RAJA
> RAJA/pattern/teams should have non-policy stuff.
> RAJA/policy/ should have the policy implementations (like RAJA/policy/cuda/teams, RAJA/policy/sequential/teams, etc.)

Sounds good, I'll do some organization of the files.

artv3:
@rhornung67 , @ajkunen , @MrBurmark, @rchen20, @trws  I reorganized the files and thanks to @MrBurmark we now have support for RAJA reductions in RAJA Teams. There are a few topics to discuss further but I was wondering if we  could get a review of this PR. 

I also moved the matrix-matrix multiply performance example to a different branch so we may move it to the RAJA perfsuite.

artv3:
LGTM

artv3:
LGTM

artv3:
Folks, just an update there seems to be a bug with CUDA working on tracking it down this week.

artv3:
LGTM

artv3:
We think we figured it out! Thanks  @mdavis36 @MrBurmark !

artv3:
LGTM

artv3:
LGTM

artv3:
LGTM

examples/teams_matrix-multiply.cpp
```diff
@@ -0,0 +1,866 @@
...
+//    or device.
+//
+// 2) RAJA Team loops follows the thread/block programming models of CUDA/HIP
+//    and considers programming using a group of threads in which we group into
+//    teams. Number of threads and teams are defined inside the Resources struct.
+//
+// 3) Launch context is used synchronize threads within a team, an example of this
+//    is presented further below.
+//
+// 4) Parallism is expressed through RAJA loops. Hierachial parallism can be
```
rchen20, line 305:
```suggestion
// 4) Parallelism is expressed through RAJA loops. Hierarchical parallelism can be
```

artv3:
LGTM

rchen20 (approved):

MrBurmark (approved):

artv3:
LGTM

artv3:
ping @rhornung67 @davidbeckingsale .

rhornung67 (approved):

artv3:
LGTM

artv3:
All green! Good to merge?

rhornung67:
> All green! Good to merge?

Sorry @artv3 I merged another one before I saw this. I merged in develop and restarted Gitlab CI. This will be next. When it comes back green, go ahead and merge.

rhornung67:
LGTM

artv3:
LGTM

artv3:
LGTM

artv3:
LGTM

artv3:
Alright alright! Took some tries but we almost have all green! Just waiting on appveyor and I'll merge !

PR #931, opened 2020-11-19
docs: fix alignment in cuda/hip policy table

balos1:
The cuda/hip policy table does not render in the current version of the documentation. This PR fixes a misaligned entry in the cuda/hip policy table so that it renders.


artv3 (approved):

rhornung67 (approved):
Thanks!

PR #939, opened 2021-01-05
Allow RAJA to build as a shared library on Windows

liu15:
- This PR is a bugfix that allows RAJA to build as a shared library on Windows


include/RAJA/util/Registry.hpp
```diff
@@ -86,7 +86,7 @@ namespace util {
 
     // begin is not defined here in order to avoid usage of an undefined static
     // data member, instead it's instantiated by RAJA_INSTANTIATE_REGISTRY.
-    static iterator begin();
+    static RAJASHAREDDLL_API iterator begin();
```
rhornung67, line 89: Are these the only places that we will have to add this macro? Or, will new locations pop up based on usage?

liu15: I'm not entirely sure.  It should not be needed for anything that is inlined.

trws:
This is... kinda dangerous. Why is it desirable to build it as a DLL rather than a static library on windows?  On linux and other unices it doesn't really matter, but on Windows this could mean ending up with multiple independent copies of raja in the same process, and might violate some of our synchronization guarantees.  That's a danger on Windows no matter what we do I suppose, since it could get embedded in other DLLs anyway, but what's the end goal of adding this functionality?

liu15:
The ALE3D Windows build uses shared libraries heavily (it is a bit idiosyncratic) in order to reduce the executable size (the ALE3D executable is huge).  Some of the libraries that depend on RAJA are therefore shared, requiring RAJA to be shared.

I'm not clear how linking this as a shared library results in multiple independent copies on the same process (or is this if you use dlopen to open the library?).

trws:
Every DLL is effectively (for language definition purposes) an independent application, as long as all libraries load the same raja DLL, you get one.  If any of them load a different one, different compile or version or what have you, they will both be loaded separately.  That's just how the windows linking model works, and part of how they provide ABI stability the way they do.  Usually it's convenient, but in some cases it's dangerous for anything that relies on global state.

It would actually link just fine statically into multiple DLLs, but it would make the result slightly larger and might cause the problem I was just mentioning.  If you need this, I'm fine with it, but I'd strongly prefer it be built into the cmake setup as described in an old article [here](https://blog.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/) until and unless we actually go through and do symbol visibility markings for all platforms, or otherwise generate an export table or similar.  Trying to do it piecemeal is likely to cause a long process of whack-a-mole adding symbols as cases pop up downstream.

liu15:
We do need this right now and I would prefer not having to patch the official RAJA releases any more than we need to.  Unfortunately, our Windows build does not currently automatically use cmake :( 

We hope to get to that eventually, but we have a long ways to go.

davidbeckingsale:
This was discussed at the RAJA developer meeting - I think we should proceed with merging this PR, and then transition to using CMake's builtin support. This is already how we support shared Windows builds in Umpire, and I believe Axom has taken the same path.

trws:
Ok, if we want to take this, I’d call it not officially supported 
(yet) but take it if the macro is made to work (or at least named) 
portably.  Make it something like `RAJA_EXPORT` and give it the 
visibility attribute on gcc-like platforms.  That’ll be the start of 
what we need to do to do it right longer-term anyway, so may as well.

rhornung67:
It sounds like we've reached some level of detente with this. I just merged the latest changes to develop into the branch so it's consistent. @trws and @davidbeckingsale are you OK with merging this after CI checks pass?

trws:
I'd prefer a rename on the macro, but if that has to wait or helps get ALE3d moving it can wait.

liu15:
We are currently patching an older version of RAJA.  I am planning on updating RAJA in the near future and I can adjust to whatever you guys decide.

davidbeckingsale (approved):

PR #943, opened 2021-01-12
Default to using cub from cuda toolkit

### LINKED ISSUE #932: Compilation issue with cuda 11 and thrust and cub sumbodule
This issue shows up in code that attempts to use thrust and RAJA together with cuda 11. It happens because Thrust requires the cub provided in the cuda installation and hence RAJA's cub submodule is incompatible. Given that cub is not shipping in the cuda install, would it make sense to use that cub by default instead of the submodule when available to avoid this compatibility issue?

MrBurmark:
# Summary
Change the default cub to the external cub that comes installed in the cuda toolkit.

- This PR is something else
  - Defaults to use external cub from the cuda toolkit
  - Fixes #932


MrBurmark:
I added the cub paths I found in the cuda toolkit installs on LC, but I'm not sure if that necessarily covers every eventuality.

davidbeckingsale:
I think this is a good move. If we always want to get the cub from the toolkit should we use CUB_DIR still, or just use the paths in the toolkit dir?

MrBurmark:
I hadn't thought of going that far, I left CUB_DIR as an option but don't really expect it to be used.

cmake/thirdparty/FindCUB.cmake
```diff
@@ -11,7 +11,9 @@ find_path(CUB_INCLUDE_DIRS
   NAMES cub/cub.cuh
   HINTS
     ${CUB_DIR}/
-    ${CUB_DIR}/include)
+    ${CUB_DIR}/include
+    ${CUDA_TOOLKIT_ROOT_DIR}/include
+    ${CUDA_TOOLKIT_ROOT_DIR}/include/thrust/system/cuda/detail)
```
trws, line 16: Is this still an actively used path? I know it was there when they only distributed enough for thrust to work, but would have thought the versions with a full copy of cub wouldn't use it anymore.

MrBurmark: The ${CUDA_TOOLKIT_ROOT_DIR}/include path is where I found cub in cuda 11.
The ${CUDA_TOOLKIT_ROOT_DIR}/include/thrust/system/cuda/detail is where I found cub in cuda 9, 10. This install did work for the examples and tests but it could be only a partial implementation, I didn't look at it that closely.

trws:
Ok.  As long as it’s working for our tests it should be fine either 
way.  If we end up adding a feature that uses something it doesn’t 
have we can deal with it then.  Thanks!

trws (approved):
Hit update to merge in develop.  Assuming that comes back green, I'm happy.

rhornung67 (approved):
@MrBurmark have you tested this with multiple versions of CUDA?

MrBurmark:
I tried this with one version of cuda 9, 10, and 11 but not exhaustively.
The thing I don't like about this is that RAJA will just build but people who aren't aware of the change will not be able to build until they add the cub to their includes.

trws:
Do the include paths not get added to the interface of RAJA's exported cmake dependency?

MrBurmark:
Will that help people with autoconf based build systems?

trws:
No, but having it appear in the includes in the generated pkg-config file would.  Which looks like it's not currently happening, so that's actually probably a good thing to fix if possible:

https://github.com/LLNL/RAJA/blob/a9406e9b061fd0f55c42e6c5a5a739977dd20ab8/share/raja/pkg-config/RAJA.pc.in#L8

MrBurmark:
Ok, Is there a way to easily add the cub includes conditionally for just cuda builds with external cub?

trws:
Probably the easiest is to add another expansion in there, `@RAJA_CUB_INCLUDES@` maybe, that defaults to blank but when built that way contains the appropriate includes.

CMakeLists.txt
```diff
@@ -271,17 +253,31 @@ target_include_directories(RAJA
...
+  if(NOT ENABLE_EXTERNAL_CUB)
+    install(DIRECTORY tpl/cub/ DESTINATION include FILES_MATCHING PATTERN *.cuh)
+  endif()
 endif()
-if(NOT ENABLE_EXTERNAL_ROCPRIM)
-  install(DIRECTORY tpl/rocPRIM/rocprim/include/ DESTINATION include FILES_MATCHING PATTERN *.hpp)
+if (ENABLE_HIP)
+  if(NOT ENABLE_EXTERNAL_ROCPRIM)
+    install(DIRECTORY tpl/rocPRIM/rocprim/include/ DESTINATION include FILES_MATCHING PATTERN *.hpp)
+  endif()
```
MrBurmark, line 280: I changed this so that non-external cub and rocprim are only installed with RAJA when we are using cuda or hip respectively. This makes sense to when we are using hip to target amd gpus, but if we are using hip to target nvidia gpus we need to install cub instead of rocprim. Do we have a configure option for this so we can easily make this distinction?

CMakeLists.txt
```diff
@@ -183,48 +183,30 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel" AND CMAKE_CXX_COMPILER_VERSION V
...
-      set(raja_depends
-        ${raja_depends}
-        cub)
-    else()
-      message(WARNING "External CUB not found.")
-      set(ENABLE_EXTERNAL_CUB Off)
-    endif()
+    set(raja_depends
+      ${raja_depends}
+      cub)
```
MrBurmark, line 195: Note that I moved this into SetupPackages.cmake so the INCLUDE_DIRS are available in SetupRajaConfig.cmake.

CMakeLists.txt
```diff
@@ -183,48 +183,30 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel" AND CMAKE_CXX_COMPILER_VERSION V
...
-      blt_register_library(
-        NAME cub
-        INCLUDES $<BUILD_INTERFACE:${CUB_INCLUDE_DIRS}>)
-      set(raja_depends
-        ${raja_depends}
-        cub)
-    else()
-      message(WARNING "External CUB not found.")
-      set(ENABLE_EXTERNAL_CUB Off)
-    endif()
```
MrBurmark, line 204: Note that I moved this into SetupPackages.cmake so the INCLUDE_DIRS are available in SetupRajaConfig.cmake.

CMakeLists.txt
rhornung67: I don't think so. @davidbeckingsale would it be best to add support for this to BLT?

rhornung67 (approved):

CMakeLists.txt
MrBurmark: Currently the configuration with hip targeting cuda is broken. Shall I always look for cub or install cub when hip is enabled (as was done previously) to workaround the issue until we have enough information to do this right?

rhornung67: Whatever it takes to get it working for now, I guess. Do we have to worry about version issues between rocm and cub?

MrBurmark: Not that I know of. In this case you are really using nvcc so the main version issues will be the same one this PR is trying to address.
I'm thinking of adding ENABLE_CUB that defaults to On when ENABLE_CUDA or ENABLE_HIP is on and Off otherwise. Do you think this would create confusion because it is always required for cuda builds and hip(cuda) builds?

rhornung67: Not sure. @ajkunen @trws @davidbeckingsale what do you guys think?

trws:
As of now, we don't support building the HIP version with nvcc anyway do
we? I wouldn't worry too much about it for now, but it's something we
might want to document.

MrBurmark:
I know we use cub instead of rocprim with hip(nvcc) so we at least have thought about it.

MrBurmark:
Updated documentation regarding cub use and HIP targeting nvidia GPUs.

CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ set(RAJA_CXX_STANDARD_FLAG "default" CACHE STRING "Specific c++ standard flag to
 option(ENABLE_TBB "Build TBB support" Off)
 option(ENABLE_TARGET_OPENMP "Build OpenMP on target device support" Off)
 option(ENABLE_CLANG_CUDA "Use Clang's native CUDA support" Off)
-option(ENABLE_EXTERNAL_CUB "Use an external cub for scans" Off)
+option(ENABLE_EXTERNAL_CUB "Use an external cub for scans" On)
```
rhornung67, line 56: This should only be on if CUDA is enabled or HIP building for nvidia hardware, correct?

MrBurmark: This is always used in places guarded by ENABLE_CUDA.
This PR does not include support for HIP targeting CUDA.

rhornung67: OK. It seems a little froggy to me though. Does it seem confusing to you?

MrBurmark: I know what you mean, I can put this option in a conditional if that makes you feel better.

MrBurmark: As in, if ENABLE_CUDA option On else option Off

rhornung67: @davidbeckingsale @trws @ajkunen do you have any opinions on this?

trws: Is this still doing anything? We had it to override the vendored version, but are no longer using a vendored version.

MrBurmark: Currently if its turned off we install cub from the submodule. We remove the cub submodule entirely and only support external cub.

rhornung67: It seems we want to support three alternatives: 1) use cub submodule, 2) use external cub that user provides, and 3) use cub associated with the given nvcc toolchain. Correct, or do we just want 2 and 3?

To me, it would be clearer if option names and CMake logic follows the options we support. Here, "EXTERNAL_CUB" only captures one option and using it for other choices seems confusing.

trws:
I think option 1 is gone with this PR, which is why I was asking.  We 
could make it that if the user provides a cub, we use it, if not we fall 
back to the mvcc provided one, and this option goes away.  Otherwise we 
can keep it for compatibility reasons so we can issue an error if it’s 
on and the user doesn’t give us a cub path.

MrBurmark:
To make things a little more clear here is the behaviours based on the 1, 2, 3 from @rhornung67's comment.
Currently the default is 1 (cub submodule) with 2 and 3 possible by setting EXTERNEL_CUB to On and setting CUB_DIR.
With this PR the default is 3 (cub from cuda toolkit) with 2 possible by setting CUB_DIR and 1 possible by setting EXTERNEL_CUB to Off.

trws:
So the submodule is staying?

MrBurmark:
I don't mind getting rid of it if everyone agrees.

rhornung67:
> I don't mind getting rid of it if everyone agrees.

As long as it works for versions of CUDA from 9.2 (which I think we still need to support for a while), I'm OK with getting rid of the submodule. Then, it seems like ENABLE_EXTERNAL_CUB should be off by default if using cub associated with the specified nvcc compiler, which I assume would be the default use case. And ENABLE_EXTERNAL_CUB is turned on by a user if they want to point to something else. 

Please update the parts of the user guide to describe the new usage pattern. I believe cub dependence is mentioned in the getting started section and where the cmake options are describe in the configuration options section.

Thank you.

MrBurmark:
I have removed the CUB submodule and changed the behavior of ENABLE_EXTERNAL_CUB.
Now ENABLE_EXTERNAL_CUB=Off, which is the default, means that the CUB installed with the cuda toolkit will be used. ENABLE_EXTERNAL_CUB=On uses an external CUB at CUB_DIR. In actuality the implementation does the same thing in both cases, but CUB_DIR takes precedence.
Also in order to allow hip targetting nvcc to work setting ENABLE_EXTERNAL_CUB=On causes cub to be used with any build.

cmake/SetupPackages.cmake
```diff
@@ -27,4 +27,35 @@ if (ENABLE_TBB)
...
+endif ()
+
+if (ENABLE_CUDA OR ENABLE_EXTERNAL_CUB)
+  find_package(CUB)
+  if (CUB_FOUND)
+    blt_register_library(
+      NAME cub
+      INCLUDES ${CUB_INCLUDE_DIRS})
+  else()
+    if (ENABLE_EXTERNAL_CUB)
```
rhornung67, line 39: Is there a good reason why the logic is different for cub and rocprim here and below?

docs/sphinx/user_guide/config_options.rst
```diff
@@ -137,16 +137,23 @@ and their default settings:
       Variable                 Default
       ======================   ==========================================
       ENABLE_CLANG_CUDA        Off (when on, ENABLE_CUDA must also be on)
-      ENABLE_CUB               On (when CUDA enabled)
+      ENABLE_EXTERNAL_CUB      Off (when CUDA enabled)
```
rhornung67, line 140: Is ENABLE_EXTERNAL_ROCPRIM also dumented in this file?

cmake/SetupPackages.cmake
MrBurmark: We still have a rocprim submodule so we don't use find_package for that. I split up the error messages to be more useful and informative depending on the setting.

docs/sphinx/user_guide/config_options.rst
MrBurmark: I'm not sure, but its behavior is not changed by this PR.

rhornung67: Apparently it is not documented. Would you please add documentation for that option in the table? Thank you.

cmake/SetupPackages.cmake
rhornung67: OK. Why do we want to keep the rocprim submodule?

MrBurmark: I'll check if rocprim is included in the rocm install, but I don't remember seeing it in there.

rhornung67: Not a big deal. We should document the difference between how we handle cub and rocprim in our user guide.

MrBurmark: It appears that rocprim is part of the rocm install after all. I'll change the code to use that instead of the submodule.

docs/sphinx/user_guide/config_options.rst
MrBurmark: added

cmake/SetupPackages.cmake
MrBurmark: done

MrBurmark:
It looks like cub and rocPRIM are not always installed as part of the cuda toolkit and rocm respectively,
The cuda and rocm install on travis and my ubuntu rocm 3.8 install do not have the libraries.

trws:
That's certainly true for rocm, but what cuda install is missing it?
There may be a package split, say in the debian/ubuntu split of cuda,
that breaks it out.

MrBurmark:
On ubuntu thrust, and hence cub, is in /usr/include. If left unset CUDA_TOOLKIT_ROOT_DIR is set to /usr by cmake or blt. This happens to work because findCUB then looks for thrust in ${CUDA_TOOLKIT_ROOT_DIR }/include/thrust. This doesn't seem like the ideal way to find thrust/cub to me but it works.

MrBurmark:
On the AMD side of things you can install rocprim as well as rocm as packages on ubuntu. Then rocprim appears in the rocm install as expected and that works. So we could advise people to install rocprim as well as rocm or go back to the submodule.

MrBurmark:
Or should we keep the submodules and strongly recommend that people use the cub and rocprim that came with their cuda toolkit or rocm install for compatibility and up-to-dateness, @rhornung67 @davidbeckingsale @ajkunen @trws?

rhornung67:
@MrBurmark this PR has merge conflicts that can't be resolved by pulling in the latest develop changes.

MrBurmark:
Merged, can I get a review @rhornung67 @davidbeckingsale @trws @ajkunen?

rhornung67:
@MrBurmark the gitlab ci resource allocation script for quartz failed --don't know why. I just restarted and test pipeline is running. Fingers crossed.

rhornung67 (approved):
Looks good @MrBurmark I don't understand what's going on with the travis jobs (intel18, gcc6, for example).

MrBurmark:
@rhornung67 What about travis?

rhornung67:
> @rhornung67 What about travis?

Look at the log output for gcc6 and intel18: https://github.com/LLNL/RAJA/pull/943/checks?check_run_id=2176554179

the logs end before all tests are run, but the checks still pass....

MrBurmark:
I assumed my browser was just choking trying to display the whole log. Maybe I can get the raw log to work?

rhornung67:
The raw log is the same as the "window" log.

MrBurmark:
Ya I was just wondering if the graphical window was adding enough overhead that the browser gave up after a while, but the raw log looks the same.

rhornung67:
I don't get it. Feel free to merge.

PR #1019, opened 2021-04-16
Sort and Scan Resource Arguments

### LINKED ISSUE #1013: RAJA scan and sort do not support resources
RAJA scan and sort do not support resources yet which makes it much harder to transition to using resources because they always run in stream 0 on cuda and hip.

MrBurmark:
# Add resource arguments to sort and scan

This PR adds support for resources in the sort and scan API methods.
This PR also removes deprecated sort and scan API methods taking iterators.
Existing calls can be easily refactored using span.
For example `begin, end -> RAJA::make_span(begin, N)`.

Note to self to update the documentation if it still uses iterators.

- This PR is a refactoring
- It does the following:
  - Modifies/refactors RAJA scan and sort methods to take resources
  - Fixes #1013


include/RAJA/pattern/scan.hpp
```diff
@@ -31,202 +31,66 @@
...
-                       Container &c,
+RAJA_INLINE
+concepts::enable_if_t<resources::EventProxy<Res>,
+                      type_traits::is_execution_policy<ExecPolicy>,
+                      type_traits::is_resource<Res>,
+                      type_traits::is_range<Container>>
+inclusive_scan_inplace(ExecPolicy&& p,
+                       Res& r,
+                       Container&& c,
                        Function binop = Function{})
```
MrBurmark, line 80: It seems that earlier nvccs are having trouble distinguishing these overloads.

mdavis36: Just with inclusive_scan_implace or are you seeing this with other methods as well?

MrBurmark: It does affect the other sor and scan policy by value interface functions. Luckily @trws gave me a workaround that looks promising.

mdavis36 (approved):
Assuming tests passed this looks good :+1:

rhornung67 (approved):

artv3 (approved):

PR #1031, opened 2021-04-26
Fixing omp-target cmake test bug

mdavis36:
# Summary

- This PR is a bugfix
  - It fixes compiler errors from a missing cmake variable needed in some generated OpenMP Target cpp test files.


rhornung67 (approved):
Thanks @mdavis36

PR #1046, opened 2021-05-11
[WIP] reducer stream safety, allocator support, and openmp agnostic headers

MrBurmark:
# Summary
Make reducers stream safe, add support for user specified allocators, and allow openmp enabled library's headers to be used without openmp and still link with the library.

- This PR is a (refactoring, bugfix, feature, something else)
  - Refactors gpu reducer memory management to be stream safe
  - Refactors gpu memory (and possibly cpu memory) allocators to be changed by the user
  - Refactors use of openmp in gpu reduction headers to allow the headers to be used without or without openmp compiler support. This moves the actual use of openmp into the library portion so the headers do not depend on it. Then code compiled with or without an openmp compiler can link to the same raja library.

1. First idea or goal
2. Second idea
3. Third idea

- (impractical)
- (too big)
- (not a good idea anyway)


include/RAJA/util/Allocator.hpp
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * \brief  Allocator Provides a generic interface for allocation in RAJA
+ *
+ * Allocator& device_pool = RAJA::cuda::get_device_allocator();
+ *
+ * RAJA::cuda::set_device_allocator<SomeOtherDevicePoolAllocator>(args...);
+ *
+ ******************************************************************************
+ */
+struct Allocator
```
MrBurmark, line 41: Here is the Allocator class that is used to make memory allocation more flexible.

include/RAJA/policy/cuda/MemUtils_CUDA.hpp
```diff
@@ -42,84 +42,66 @@ namespace RAJA
...
-  bool free(void* ptr)
-  {
-    cudaErrchk(cudaFreeHost(ptr));
-    return true;
-  }
-};
-
-//! Allocator for device memory for use in basic_mempool
-struct DeviceAllocator {
+extern void set_device_allocator(RAJA::Allocator* allocator);
```
MrBurmark, line 48: Backends expose functions to get, set, and reset the allocators. This allows apps to provide their own allocators at startup and remove them when shutting down. Currently this is only exposed for cuda and hip, but a host allocator could be exposed as well.

include/RAJA/util/AllocatorUmpire.hpp
```diff
@@ -0,0 +1,113 @@
...
+ * Example:
+ *
+ * umpire::Allocator umpire_allocator;
+ *
+ * RAJA::AllocatorUmpire<umpire::Allocator> raja_alocator(umpire_allocator);
+ *
+ ******************************************************************************
+ */
+template <typename allocator_t>
+struct AllocatorUmpire : Allocator
```
MrBurmark, line 43: This class is designed to wrap an umpire allocator to make it easy for codes to replace RAJA's allocators with umpire pools.

PR #1051, opened 2021-05-17
Bugfix/burmark1/use blt cxx std

### LINKED ISSUE #1049: Automatic c++ std detection can result in multiple standards in cxx flags
I noticed this with the new xl build script, but manually setting BLT_CXX_STD causes more than one -std=c++?? to end up in CMAKE_CXX_FLAGS.

MrBurmark:
# Summary
Set BLT_CXX_STD where we find the cxx std version. Remove RAJA_CXX_STANDARD_FLAG and use BLT_CXX_STD  instead.

- It does the following:
  - Modifies/refactors the way we set the cxx std version in cmake
  - Fixes #1049

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ matrix:
     env:
     - COMPILER=/opt/intel/bin/icpc
     - IMG=icc18
-    - CMAKE_EXTRA_FLAGS="-DENABLE_FORCEINLINE_RECURSIVE=Off -DENABLE_TBB=On"
+    - CMAKE_EXTRA_FLAGS="-DENABLE_FORCEINLINE_RECURSIVE=Off -DENABLE_TBB=On -DBLT_CXX_STD=c++14"
   - compiler: nvcc10.2
     env:
     - COMPILER=g++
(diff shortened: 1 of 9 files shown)


davidbeckingsale (approved):
Nice, this a cool solution.

rhornung67 (approved):
Looks great. Thanks @MrBurmark

trws (commented):
I like it.  The only thing I'd request, and it's an addition admittedly, is a check that the user-set flag is at least c++11.

davidbeckingsale:
I think it's a CMake version issue in Travis, and on GitLab there seems to be some incompatibility between Intel and the GCC headers.

trws:
I haven't checked the gitlab logs, but unless you change the gcc that intel uses for its STL from 4.9.3 to one of the devtoolset ones or one of the ones in TCE it will fail on 17.  Checking for c++98 seems reasonable.

rhornung67 (approved):
Thanks @MrBurmark

rhornung67 (approved):
Thanks @MrBurmark !

PR #1067, opened 2021-05-28
Docs for naming RAJA Teams based kernels

artv3:
# Summary

Adds a short documentation on using the kernel_name parameter in RAJA Team methods. 
Currently we only support NVTX tools. I noted in the docs that there may be some API changes as we integrate with other tools.


rhornung67 (approved):
Thanks @artv3

PR #1069, opened 2021-06-01
Fix extra cmake arguments in scripts

MrBurmark:
# Summary
Fix how we were handling extra cmake arguments tin our build scripts. Previously the compiler versions were also accidently being passed along. In some cases extra cmake arguemnts could not be passed in because the script expected an exact number of arguments.

- This PR is a bugfix
  - Fixes an iddue in our build scripts


artv3 (approved):

rchen20 (approved):

rhornung67 (approved):
Looks good @MrBurmark I tried a couple of things to do essentially what you are doing here, but couldn't get it to work. Nice job!

PR #1073, opened 2021-06-03
feature/desul-atomics

JohnHynes:
PR to import atomic support from https://github.com/desul/desul as a new back-end for RAJA atomics.

Useful CMake flags for testing:

>  -DRAJA_ENABLE_DESUL_ATOMICS=On


include/RAJA/policy/desul/atomic_desul_openmp.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+#else  // not defined RAJA_COMPILER_MSVC
+
+struct omp_atomic {};
+
+RAJA_SUPPRESS_HD_WARN
+template <typename T>
+RAJA_HOST_DEVICE
+RAJA_INLINE T
+atomicAdd(omp_atomic, T volatile *acc, T value) {
```
JohnHynes, line 41: So the atomic_desul_openmp file acts as the middle-man between the RAJA and desul APIs. At the moment RAJA uses this ```T volatile *acc``` parameter, but the desul API doesn't like ```volatile```.

trws: Right, volatile is orthogonal to the way that atomic works these days, and we decided not to use it.  Ideally it would be best if we could remove `volatile` from the raja interface, but if we have to leave it there, the `acc` variable will have to go through either a `const_cast<std::remove_volatile_t<decltype(*acc)>*>(acc)` or similar to cast off the `volatile`.

JohnHynes: Perhaps a 'dumb' question, but is there an advantage or a requirement to use ```std::remove_volatile_t``` as opposed to a simple ```const_cast<T*>(acc)``` to remove ```volatile``` in this case? So far I've only made the latter work for compilation, but I suspect that there may be something else going on behind the scenes that results in failed tests.

trws: Not really.  The `const_cast` should be fine here.  I do think that trying a switch to not taking volatile in the first place would be a good idea though.  If we can get away with it, the result will be more likely to behave correctly if someone passes in a pointer to a variable that was declared volatile than if we cast it away from the pointer unconditionally.

JohnHynes: Makes sense, the undefined behavior of casting volatile away might be causing some issues in testing from what I can gather. Moving the current atomic build away from volatile seems like the way to go long-term, so I'll tinker around and see if I can get that to work.

include/RAJA/policy/desul/atomic_desul_openmp.hpp
```diff
@@ -0,0 +1,190 @@
...
+RAJA_SUPPRESS_HD_WARN
+template <typename T>
+RAJA_HOST_DEVICE
+RAJA_INLINE T atomicExchange(omp_atomic, T volatile *acc, T value)
+{
+  (void)acc;
+  (void)value;
+  //desul::atomic_(...);  Not sure which desul function to call here.
+  return T{0};
+}
```
JohnHynes, line 171: @trws Not sure which desul function would be called here.

JohnHynes:
Currently only 3 tests are failing :

> 	 89 - test-forall-atomic-basic-OpenMP.exe (Failed)
> 	184 - test-reducer-reset-seq.exe (Failed)
> 	193 - test-atomic-ref-exchanges.exe (Failed)

Not sure why ```test-reducer-reset-seq``` fails, I didn't see anything about atomics within the files relating to it.

trws:
It would be `atomic_exchange`, I’m not sure why we don’t have that 
in the generic set, but it looks like it exists in all the specific 
backends.  I’ll look into that.

JohnHynes:
Building using one of the pre-made scripts solved the issue of test-reducer-reset-seq failing, leaving only two tests failing now regarding ```atomicExchange``` and ```atomicCAS```.

> 	 89 - test-forall-atomic-basic-OpenMP.exe (Failed)
	193 - test-atomic-ref-exchanges.exe (Failed)

rhornung67:
@JohnHynes did you figure out what was different with the build script; e.g., compiler flags?

JohnHynes:
@rhornung67 I'm working on finding the specific flag(s), but it's nice that it seems to be a flag issue and not something far worse.

rhornung67:
When you figure it out, please let us know.  We've run into this before.

include/RAJA/policy/desul/atomic_desul_openmp.hpp
```diff
@@ -0,0 +1,194 @@
...
+template <typename T>
+RAJA_HOST_DEVICE
+RAJA_INLINE T atomicCAS(omp_atomic, T volatile *acc, T compare, T value)
+{
+  return desul::atomic_compare_exchange_strong(const_cast<T*>(acc)
+                                , compare
+                                , value
+                                , raja_default_desul_order{} // success order
+                                , raja_default_desul_order{} // failure order
+                                , raja_default_desul_scope{});
```
JohnHynes, line 186: @trws Just wanted to double check that the success and failure memory orders here make sense.

JohnHynes:
Appears that the missing flag that was causing test-reducer-reset-seq to fail was ```-finline-limit=20000```

trws:
Oh my, that's not good.  That's either a compiler bug, or a raja bug where we only get correct behavior when something gets inlined out, probably due to a reference collapsing to the original after inlining or some similar.

include/RAJA/policy/desul/atomic_desul_openmp.hpp
trws: Yeah, that's perfectly fine.  If we had a stronger memory order by default, it might help on a small subset of platforms (powerpc for example) to have a weaker one on failure, but it's more consistent to keep it the same until we have a way for the user to decide for themselves.

rhornung67:
@trws we've seen this issue before. It seems to only occur with the gnu compiler.

include/RAJA/policy/desul/atomic_desul_openmp.hpp
```diff
@@ -0,0 +1,189 @@
...
+RAJA_SUPPRESS_HD_WARN
+template <typename T>
+RAJA_HOST_DEVICE
+RAJA_INLINE T atomicInc(omp_atomic, T volatile *acc, T val)
+{
+  return desul::atomic_fetch_add(const_cast<T*>(acc)
+                                , val
+                                , raja_default_desul_order{}
+                                , raja_default_desul_scope{});
+}
```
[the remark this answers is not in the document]
JohnHynes: I didn't see these tested anywhere in unit tests either unless I'm mistaken, which is why they might have only appeared in the integration test.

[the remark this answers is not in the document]
trws: Ah, that's a tricky one unless you're used to cuda behavior, have a look [here](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomicinc).  The atomicInc function only ever increments by 1, but it wraps at `val`.  Makes it easier to write things like atomic circular buffers, or avoid rollover in signed types.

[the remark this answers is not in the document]
trws: I don't think we have an equivalent operation in desul right now, just the non-value version.  Would be really good to add one though.

Would you be interested in taking a crack at it?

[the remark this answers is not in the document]
JohnHynes: Oh wow that's interesting. I'll take a crack at it and see if I can make an implementation work!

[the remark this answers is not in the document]
rhornung67: Thanks @JohnHynes 👍

crtrott:
FYI there are some more changes coming to the DESUL atomics internal stuff I have uncovered while doing full broadband testing in Kokkos. Probably next week or so. (no interface changes ...)

trws:
That's probably a good thing.  We haven't seen any interface changes we need other than the fix on the generic inc/dec thing.  Which reminds me actually, do you think it would be a good idea to disambiguate the "normal" inc/dec from the wrapping versions? We're currently using the same name for both and it's clearly causing confusion.  Maybe something like `atomic_wrapping_{inc/dec}`?

dalg24:
Other suggestion: `atomic_modulo_{inc,dec}`

rhornung67:
I like @dalg24 's suggestion for the name. 'Modulo' seems to have more meaning than 'wrapping'.

trws:
That seems reasonable. It's not literally implemented that way but if that's clearer I'm good with it.

rhornung67:
Whatever will be most clear to users  I'm good with.

JohnHynes:
I still need to write the raja unit tests for some of the atomics, most notably the wrapping inc/dec. While everything gets tested during integration testing, the unit tests don't exist.

trws:
Looks like we have an issue on msvc with `RAJA::policy::omp::omp_atomic` being defined more than once.  Maybe both the desul and non-desul versions are included at the same time?

JohnHynes:
@trws I haven't run into that one yet, it might be pulling in both, but it probably stems from how we've moved around the atomic policy declarations.

JohnHynes:
@trws Just pushed a change in the macro guards that should<sup>tm</sup> fix the issue with msvc.

include/RAJA/policy/openmp/policy.hpp
```diff
@@ -288,6 +305,11 @@ struct omp_synchronize : make_policy_pattern_launch_t<Policy::openmp,
 ///////////////////////////////////////////////////////////////////////
 ///
 
+///
+/// Type alias for atomics
+///
+using policy::omp::omp_atomic;
```
davidbeckingsale, line 311: I think you might need to put the ifdef for MSVC here, instead of up in the namespace.

include/RAJA/policy/desul/atomic_desul_openmp.hpp
```diff
@@ -0,0 +1,184 @@
...
+using raja_default_desul_order = desul::MemoryOrderRelaxed;
+using raja_default_desul_scope = desul::MemoryScopeDevice;
+
+namespace RAJA
+{
+
+#if defined(RAJA_COMPILER_MSVC)
+
+// For MS Visual C, just default to builtin_atomic for everything
+using omp_atomic = builtin_atomic;
```
JohnHynes, line 31: This line here seems to cause errors as far as I can tell, seems to result in `builtin_atomic` being undefined.

JohnHynes:
@trws @davidbeckingsale Seemed to have fixed the issue, just had to make sure everything in openmp/atomic.hpp was still guarded properly after moving the msvc guard to policy/openmp/policy.hpp

trws:
Ok, looks like we need a rebase or merge to resolve the script conflict.  After that hopefully this would pass?

JohnHynes:
@trws The change to the script is small and just allows for more args to be passed in. That change can be omitted if needed.

JohnHynes:
Also, atomic_desul_openmp.hpp  and atomic_desul_cuda.hpp don't need to

trws:
It needs to either be omitted or merged so we don't have a conflict, mainly so appveyor will build it or at least get cleared.

CMakeLists.txt
```diff
@@ -75,6 +75,7 @@ option(RAJA_ENABLE_BOUNDS_CHECK "Enable bounds checking in RAJA::Views/Layouts"
 option(RAJA_TEST_EXHAUSTIVE "Build RAJA exhaustive tests" Off)
 option(RAJA_TEST_OPENMP_TARGET_SUBSET "Build subset of RAJA OpenMP target tests when it is enabled" On)
 option(RAJA_ENABLE_RUNTIME_PLUGINS "Enable support for loading plugins at runtime" Off)
+option(RAJA_ENABLE_DESUL_ATOMICS "Enable support of desul atomics" Off)
```
JohnHynes, line 78: @davidbeckingsale It appears that the desul tests, which rely on kokkos, are being included by default when ``-DDESUL_ENABLE_TESTS=Off`` is not specified. Would we be able to default this to "Off" similar to this?

davidbeckingsale: Yup, you can add `set(DESUL_ENABLE_TESTS Off CACHE BOOL "")` to turn them off here.

JohnHynes:
@rhornung67 @davidbeckingsale Which files should I be looking at to make CI tests that enable desul atomics? And which compilers do you think we should be targeting for them?

rhornung67:
@JohnHynes I think we agreed that it is sufficient to test only on Gitlab CI at this point. I suggest adding one desul atomic build and test run for corona (HIP backend), lassen (CUDA backend), and ruby (OpenMP backend).  I believe the scripts that would need to be modified are in here: https://github.com/LLNL/RAJA/tree/develop/.gitlab But, I don't see an example showing where you would add the CMake option to enable desul atomics. @davidbeckingsale can you chime in?

davidbeckingsale:
> @JohnHynes I think we agreed that it is sufficient to test only on Gitlab CI at this point. I suggest adding one desul atomic build and test run for corona (HIP backend), lassen (CUDA backend), and ruby (OpenMP backend). I believe the scripts that would need to be modified are in here: https://github.com/LLNL/RAJA/tree/develop/.gitlab But, I don't see an example showing where you would add the CMake option to enable desul atomics. @davidbeckingsale can you chime in?

Yup that's right. I emailed John and Kristi about it

include/RAJA/policy/cuda/policy.hpp
```diff
@@ -122,6 +123,19 @@ struct cuda_reduce_base
                                                 RAJA::Platform::cuda> {
 };
 
+//
+// Cuda atomic policy for using cuda atomics on the device and
+// the provided Policy on the host
+//
+template<typename host_policy>
+struct cuda_atomic_explicit{};
```
rchen20, line 131: Just curious, why have we moved this here? Was there a reason it couldn't be kept in policy/cuda/atomic.hpp?

include/RAJA/policy/cuda/reduce.hpp
```diff
@@ -60,23 +66,23 @@ template <typename T>
 struct atomic<sum<T>> {
   RAJA_DEVICE RAJA_INLINE void operator()(T& val, const T v)
   {
-    RAJA::atomicAdd<T>(RAJA::cuda_atomic{}, &val, v);
+    RAJA::atomicAdd(RAJA::cuda_atomic{}, &val, v);
```
rchen20, line 69: Should we include the template parameter? e.g. RAJA::atomicAdd<T>

include/RAJA/policy/loop/atomic.hpp
```diff
@@ -27,7 +27,7 @@
 namespace RAJA
 {
 
-using loop_atomic = seq_atomic;
+// Delete?
```
rchen20, line 30: I think we'll need a backward compatibility solution for this. Maybe something like`using loop_atomic = policy::loop::loop_atomic;`.

test/include/RAJA_test-atomic-types.hpp
```diff
@@ -32,4 +32,11 @@ using AtomicDataTypeList =
 #endif
               double >;
 
+using AtomicDataUnsignedTypeList =
```
rchen20, line 35: I think we'll need to guard this to run only when Desul is activated. CUDA and HIP cannot do unsigned atomics.

include/RAJA/policy/cuda/policy.hpp
JohnHynes: If I'm remembering correctly, @davidbeckingsale and I had to move omp_atomic or some other atomic policy out of its atomic.hpp file and into its policy.hpp file otherwise we would need more macro guards due to including duplicate function definitions. Once we did it with one, it made sense to do it with the others. I know for some dependencies it sort of "cascaded" once we moved one or two as well.

include/RAJA/policy/cuda/reduce.hpp
JohnHynes: Hmm, I think it might interfere with the templated types in desul/atomic.hpp since it would be ``RAJA::atomicAdd<policy, T>(...)`` and there are no defaulted types as of now. I might need to double-check that.

include/RAJA/policy/loop/atomic.hpp
JohnHynes: Makes sense to me, I'll go ahead and change that.

test/include/RAJA_test-atomic-types.hpp
JohnHynes: Hmm, I was under the impression that CUDA did support unsigned atomics from my brief work helping with desul. If that isn't the case then a macro guard definitely makes sense.

rchen20: Ah, you're right, it should work with `unsigned int` and `unsigned long long int`. I think I had put the unsigned cases into the exhaustive tests before.

include/RAJA/policy/cuda/reduce.hpp
JohnHynes: From testing I haven't found any issues one way or the other, so it might just be worth keeping the ``<T>`` in since removing it doesn't seem to do much. I could've sworn there was some issue with it at some point though, but that might have just been an artifact of a different bugfix.

test/include/RAJA_test-atomic-types.hpp
JohnHynes: Ahh cool, is that just for CUDA or does HIP also support unsigned atomics? I'm not too familiar with HIP.

include/RAJA/policy/cuda/reduce.hpp
trws: If it isn't necessary, I'd leave it out and let the compiler infer it from the arguments.  One less thing to worry someone will accidentally break later by mucking with T or the way the argument declarations are written.

test/include/RAJA_test-atomic-types.hpp
trws: They should both support them.  Hip is trying to be cuda compatible, and in both cases it should just work for everything (the wrapping atomic increment is special, since that supports very few targets, as are the floating point ones but for integral types it should just go).

trws:
As of right now, I'm getting pretty happy with this.  @rhornung67, how are you feeling about enabling this as a non-default option for the upcoming release?

I don't think we want to make it default right away, partly because of the dependency and partly because of the RDC question, but getting it out there for testing might be good.  Speaking of the RDC question, (@rhornung67, @davidbeckingsale) is there a decision on whether we want to just flat require RDC for everything? If we can, then use of DESUL atomics for non-hardware supported types becomes much simpler for users, and it would fix some issues we have otherwise had to work around in raja as well.  Now that LTO is possible in NVCC and (I think) the clang-based compilers, the perf hit is basically gone, and it makes dealing with externs/globals on the device a lot more sane.

rhornung67:
Since it's been a while since the last release and there is a lot of new stuff in it and this release has taken a while, I think it is best to do that in another release in 2-3 months. We can announce it as an option to try in the release notes and encourage users to try the new atomics out. I also think we should do some performance testing to see how well this performs relative to the atomic support we have had for a while. 

w.r.t. to RDC, I think we should hold off on requiring that until it gets easier for apps trying out the ROCm stack. I've heard from several users that they have had problems with it. Some users rely on nvidia's device LTO (added in CUDA 11) for Sierra to recover performance that they lost from separate compilation. They are working through compiler errors that appear at link time (!) using -fgpu-rdc.

trws:
Ok, it isn't necessary to have the RDC thing, just users need to be aware of the semantics they get for atomics on things like complex if they don't have it.  We tried to make sure desul will do something sane-ish if they don't set it up right, but it still wouldn't be a fun bug to track down.  Probably something we'll want to put in the docs.

JohnHynes:
@kab163 @rhornung67 Seems like the new CI tests are still failing despite working locally for me. I'm not exactly sure why, the errors don't tell me much:

> WARNING: junit.xml: no matching files              
ERROR: No files to upload                          
ERROR: Job failed: exit status 1

davidbeckingsale:
> @kab163 @rhornung67 Seems like the new CI tests are still failing despite working locally for me. I'm not exactly sure why, the errors don't tell me much:
> 
> > WARNING: junit.xml: no matching files
> > ERROR: No files to upload
> > ERROR: Job failed: exit status 1

I think it's because of the cmake error about the install stuff. CMake exits with a failing code, so even though it will "work", the compilation never happens. If you get the desul stuff fixed and point at your branch with the fix from the submodule it should start to work.

JohnHynes:
Ahh that makes sense, I'll go ahead and work on the desul install then. I picked away at it earlier but didn't get a chance to really dive in.

include/RAJA/policy/desul/atomic.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+
+namespace RAJA
+{
+
+RAJA_SUPPRESS_HD_WARN
+template <typename AtomicPolicy, typename T>
+RAJA_HOST_DEVICE
+RAJA_INLINE T
+atomicAdd(AtomicPolicy, T volatile *acc, T value) {
```
JohnHynes, line 33: @davidbeckingsale I just realized that we could maybe have defaults for  ``AtomicPolicy`` for each function here since it just gets ignored anyways. It would allow for users to say something like ``RAJA::atomicAdd(pointer, value);`` without needing to specify a dummy-policy. This should maintain the backwards compatibility granted by ``AtomicPolicy`` as well.

JohnHynes: Actually, I'm not sure that would work out in preserving backward compatibility since ``AtomicPolicy`` comes first in the parameter order.

test/unit/atomic/test-atomic-incdec.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+#include "RAJA/RAJA.hpp"
+
+#include "RAJA_gtest.hpp"
+
+#if defined(RAJA_ENABLE_CUDA)
+#include "RAJA_unit-test-forone.hpp"
+#endif
+
+using unsigned_types =
```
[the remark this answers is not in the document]
JohnHynes: Should I go ahead and try to make this a cartesian product? It seems like the method used in other tests (e.g. ones found in test/integration/forall) can't be used in the atomic tests without adding in some extra support for that. Other tests in the same test/unit/atomic directory use this same method. e.g. ``test-atomic-ref-bitwise.cpp`` and others use the ones found in ``test-atomic-ref.hpp``.

test/unit/atomic/test-atomic-incdec.cpp
```diff
@@ -0,0 +1,192 @@
...
+                    >;
+
+
+template <typename T>
+class AtomicCUDAIncDecUnitTest : public ::testing::Test
+{};
+
+TYPED_TEST_SUITE_P( AtomicCUDAIncDecUnitTest );
+
+GPU_TYPED_TEST_P( AtomicCUDAIncDecUnitTest, CUDAIncDecs )
```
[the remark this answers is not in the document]
JohnHynes: Since these tests are CUDA-specific, does it make sense to switch from a gpu testing pattern to a normal one?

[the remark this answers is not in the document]
MrBurmark: The GPU test macros just make a function with your code in it and call it in the test instead of having your code directly in the test. The testing pattern found in test/unit/resource are what I'm referring to as the normal pattern. The normal pattern uses a .cpp.in file and uses cmake to generate a version per backend to avoid code duplication. In the normal testing pattern cartesian product is used in the cpp file to instantiate test templates with combinations of different types.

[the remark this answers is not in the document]
JohnHynes: Ahh, that makes sense. I'll go ahead and try to make that change.

JohnHynes:
@rhornung67 @davidbeckingsale I pushed a first pass for the documentation. Let me know what you think, I tried to keep it concise and descriptive.

rhornung67:
> @rhornung67 @davidbeckingsale I pushed a first pass for the documentation. Let me know what you think, I tried to keep it concise and descriptive.

I enabled docs for this branch on RTD so we can see how it appears when it is generated. Here's the link: https://raja.readthedocs.io/en/feature-desul-atomics/

docs/sphinx/user_guide/feature/atomic.rst
```diff
@@ -153,3 +155,28 @@ CUDA architecture level is chosen:
...
+to DESUL typically yields positive or neutral performance results, some atomic
+operations may perform worse when using DESUL.
+
+To enable DESUL Atomics:
+
+#. Ensure that RAJA and its dependencies are configured to use C++14.
+#. Set ``RAJA_ENABLE_DESUL_ATOMICS=On``.
+
+Enabling DESUL Atomics alters RAJA atomic functions to be wrapper-functions for their
+DESUL counterparts. This removes any code changes necessary for switching between
```
rhornung67, line 174:
```suggestion
DESUL counterparts. This removes the need for user code changes to switch between
```

docs/sphinx/user_guide/feature/atomic.rst
```diff
@@ -153,3 +155,28 @@ CUDA architecture level is chosen:
...
+DESUL and RAJA implementations, except for cases where RAJA atomic helper functions
```
rhornung67, line 175:
```suggestion
DESUL and RAJA implementations. The exception to this is when RAJA atomic helper functions
```

docs/sphinx/user_guide/feature/atomic.rst
```diff
@@ -153,3 +155,28 @@ CUDA architecture level is chosen:
...
+are used instead of the backwards-compatible API functions specified by :ref:`atomic-ops`.
```
rhornung67, line 176:
```suggestion
are used instead of the backwards-compatible API functions specified by :ref:`atomic-ops`. By *helper functions*, we mean the RAJA atomic methods which take a reduction policy object as the first argument, instead of specifying the reduction policy type as a template parameter. 
```

docs/sphinx/user_guide/feature/atomic.rst
```diff
@@ -153,3 +155,28 @@ CUDA architecture level is chosen:
...
+
+DESUL compiles atomic functions in accordance to the scope in which they are
```
rhornung67, line 178:
```suggestion
DESUL atomic functions are compiled with the proper back-end implementation based on the scope in which they are
```

docs/sphinx/user_guide/feature/atomic.rst
```diff
@@ -153,3 +155,28 @@ CUDA architecture level is chosen:
...
+called, this removes the requirement of using atomic policies to specify
```
rhornung67, line 179:
```suggestion
called, which removes the need to specify atomic policies for
```

docs/sphinx/user_guide/feature/atomic.rst
```diff
@@ -153,3 +155,28 @@ CUDA architecture level is chosen:
...
+target backends. As a result, atomic policies such as ``cuda_atomic`` or ``omp_atomic``
```
rhornung67, line 180:
```suggestion
target back-ends. As a result, atomic policies such as ``cuda_atomic`` or ``omp_atomic``
```

docs/sphinx/user_guide/feature/atomic.rst
```diff
@@ -153,3 +155,28 @@ CUDA architecture level is chosen:
...
+will be ignored when DESUL is enabled, but are still necessary to pass in as parameters
```
rhornung67, line 181:
```suggestion
are ignored when DESUL is enabled, but are still necessary to pass in as parameters
```

docs/sphinx/user_guide/feature/atomic.rst
```diff
@@ -153,3 +155,28 @@ CUDA architecture level is chosen:
...
+to the RAJA API.
```
rhornung67, line 182:
```suggestion
to the RAJA API. This will likely change in the future and RAJA atomic policies will be removed.
```

docs/sphinx/user_guide/feature/atomic.rst
```diff
@@ -153,3 +155,28 @@ CUDA architecture level is chosen:
 
     * CUDA native 64-bit double `atomicAdd` is used.
 
+---------------------
+DESUL Atomics Support
+---------------------
+
+RAJA supports the use of `DESUL Atomics <https://github.com/desul/desul>`_ as
```
rhornung67, line 162:
```suggestion
RAJA provides support for the use of `DESUL Atomics <https://github.com/desul/desul>`_ as
```

docs/sphinx/user_guide/feature/atomic.rst
```diff
@@ -153,3 +155,28 @@ CUDA architecture level is chosen:
...
+an alternative backend to the default implementation of RAJA atomics. Enabling
```
rhornung67, line 163:
```suggestion
an alternative backend to the default implementation of RAJA atomics. DESUL atomics are considered an **experimental** feature in RAJA at this point. DESUL atomics
```

docs/sphinx/user_guide/feature/atomic.rst
```diff
@@ -153,3 +155,28 @@ CUDA architecture level is chosen:
...
+this feature may impact the performance of some atomic functions. While switching
```
rhornung67, line 164:
```suggestion
may impact the performance of some atomic functions. While switching
```

docs/sphinx/user_guide/feature/atomic.rst
```diff
@@ -153,3 +155,28 @@ CUDA architecture level is chosen:
...
+to DESUL typically yields positive or neutral performance results, some atomic
```
rhornung67, line 165:
```suggestion
to DESUL atomics typically yields positive or neutral performance results, some atomic
```

include/RAJA/policy/desul/atomic.hpp
trws: Sadly not, defaults can only be at the end.  It would have to be an overload, and the overload would probably be ambiguous with the templated one sadly.

trws:
I'm happy with this as it stands, there may be some work as we work to improve some of the performance items @JohnHynes found, but I'd say this is good to go.

trws (approved):

rhornung67 (dismissed):
Nice work @JohnHynes !!

scripts/uberenv/packages/raja/package.py
```diff
@@ -287,8 +288,12 @@ def hostconfig(self, spec, prefix, py_site_pkgs_dir=None):
                                         hip_root))
             cfg.write(cmake_cache_entry("HIP_CLANG_PATH",
                                         rocm_root + '/llvm/bin'))
-            cfg.write(cmake_cache_entry("HIP_HIPCC_FLAGS",
-                                        '--amdgpu-target=gfx906'))
+
+            hipcc_flags = '--amdgpu-target=gfx906'
+            if "+desul" in spec:
+                hipcc_flags += ';-std=c++14'
```
kab163, line 294: This confuses me. The ";" isn't a typo, is it?

kab163: Oh, that enables the flags to be added together, I assume? Hmm lol that seems dangerous :p

trws:
Yeah, that’s cmake’s syntax for a list.  It’s semicolon separated. 
  At least it isn’t meant to be used in environment variables much 
(unlike LUA_PATH, which is, and which is also semicolon separated ><).

kab163:
> Yeah, that’s cmake’s syntax for a list. It’s semicolon separated. At least it isn’t meant to be used in environment variables much (unlike LUA_PATH, which is, and which is also semicolon separated ><).
> […](#)
> On 14 Sep 2021, at 13:20, Kristi wrote: @kab163 commented on this pull request. > @@ -287,8 +288,12 @@ def hostconfig(self, spec, prefix, > py_site_pkgs_dir=None): hip_root)) cfg.write(cmake_cache_entry("HIP_CLANG_PATH", rocm_root + '/llvm/bin')) - cfg.write(cmake_cache_entry("HIP_HIPCC_FLAGS", - '--amdgpu-target=gfx906')) + + hipcc_flags = '--amdgpu-target=gfx906' + if "+desul" in spec: + hipcc_flags += ';-std=c++14' Oh, that enables the flags to be added together, I assume? Hmm lol that seems dangerous :p -- You are receiving this because you were mentioned. Reply to this email directly or view it on GitHub: [https://urldefense.us/v3/__https://github.com/LLNL/RAJA/pull/1073*discussion_r708603095__;Iw!!G2kpM7uM-TzIFchu!krdVL0CXONFZ0o311_W7KEAvgPZIosFc1eGwp6B25CLHnlcVpnCKV-Tb91FoO9o2_w$](https://urldefense.us/v3/__https://github.com/LLNL/RAJA/pull/1073*discussion_r708603095__;Iw!!G2kpM7uM-TzIFchu!krdVL0CXONFZ0o311_W7KEAvgPZIosFc1eGwp6B25CLHnlcVpnCKV-Tb91FoO9o2_w%24)

Yea, I see your point. Though for what it's worth, I think it is nicer (though probably not any better) to add all the flags to a list and then set the list to the "HIP_HIPCC_FLAGS" cmake option. Umpire does something like this here (https://github.com/LLNL/Umpire/blob/48435212edc65c4a42dacb7a034aca70c6932e29/scripts/spack_packages/umpire/package.py#L173-L184). But again, either way works!

scripts/uberenv/packages/raja/package.py
```diff
@@ -287,8 +288,13 @@ def hostconfig(self, spec, prefix, py_site_pkgs_dir=None):
...
             cfg.write(cmake_cache_entry("HIP_CLANG_PATH",
                                         rocm_root + '/llvm/bin'))
-            cfg.write(cmake_cache_entry("HIP_HIPCC_FLAGS",
-                                        '--amdgpu-target=gfx906'))
+
+            hipcc_flags = ['--amdgpu-target=gfx906']
+            if "+desul" in spec:
+                hipcc_flags.append('-std=c++14')
+            
+            cfg.write(cmake_cache_entry("HIP_HIPCC_FLAGS", ';'.join(hipcc_flags)))
```
JohnHynes, line 296: @kab163 Is this change what you had in mind?

kab163: Sure! It probably doesn't make too much difference now when there's only 2 flags that could be added, but if we needed to add a bunch of flags in the future, this approach may look better.

davidbeckingsale (approved):

trws:
I'm happy here, and would happily squash and merge this right now.

trws:
Lots of fixes landing in desul from the Sandia end btw, shaping up pretty nicely.

MrBurmark (approved):

PR #1075, opened 2021-06-08
Run gitlab ci jobs on ruby instead of quartz

kab163:
# Summary

- This PR is improving the gitlab ci
- It moves the gitlab ci tests from quartz to ruby instead (less traffic)


rhornung67 (approved):
Thanks @kab163 I think this change is in one of the CI issues we have, probably a check box in one that has several tasks. If so, please check it off.  Thank you.

Issue #1070, opened 2021-06-02
Add "Flatten policy" in RAJA Teams


artv3:
Users have requested a "flatten policy" which reshape {3,2}D thread teams to 1D.

PR #1078, opened 2021-06-10
Add a flatten policy to RAJA Teams 

artv3:
# Summary

-Enables to reshaping multi-dimensional threads in a block to 1D. 
Addresses https://github.com/LLNL/RAJA/issues/1070

### DIFF
--- a/include/RAJA/policy/cuda/teams.hpp
+++ b/include/RAJA/policy/cuda/teams.hpp
@@ -278,6 +278,50 @@ struct LoopExecute<cuda_global_thread<DIM0, DIM1, DIM2>, SEGMENT> {
 };
 
 /*
+Reshape threads in a block into a 1D iteration space
+*/
+struct flatten_block_threads_direct;
+struct flatten_block_threads_loop;
+
+template<typename SEGMENT>
+struct LoopExecute<flatten_block_threads_direct, SEGMENT>
+{
+  template<typename BODY>
+  static RAJA_INLINE RAJA_DEVICE void exec(
+      LaunchContext const RAJA_UNUSED_ARG(&ctx),
+      SEGMENT const &segment,
+      BODY const &body)
+  {
+    const int len = segment.end() - segment.begin();
+    {
+      const int tid = threadIdx.x + blockDim.x*(threadidx.y + blockDim.y*threadIdx.z);
+
+      if (tid < len) body(*(segment.begin() + tid));
+    }
+  }
+};
+
+template<typename SEGMENT>
+struct LoopExecute<flatten_block_threads_loop, SEGMENT>
+{
+  template<typename BODY>
+  static RAJA_INLINE RAJA_DEVICE void exec(
+      LaunchContext const RAJA_UNUSED_ARG(&ctx),
+      SEGMENT const &segment,
+      BODY const &body)
+  {
+    const int len = segment.end() - segment.begin();
+    {
+      for(int tid = threadIdx.x + blockDim.x*(threadidx.y + blockDim.y*threadIdx.z);
+          tid < len;
+          tid += blockDim.x * blockDim.y * blockDim.z){
+          body(*(segment.begin() + ti));
+      }
+    }
+  }
+};
+
+/*
   CUDA thread loops with block strides
 */
 template <typename SEGMENT, int DIM>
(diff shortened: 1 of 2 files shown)


include/RAJA/policy/hip/teams.hpp
```diff
@@ -268,6 +268,48 @@ struct LoopExecute<hip_global_thread<DIM0, DIM1, DIM2>, SEGMENT> {
...
+  {
+    const int len = segment.end() - segment.begin();    
+    for(int tid = threadIdx.x + blockDim.x*(threadIdx.y + blockDim.y*threadIdx.z);
+        tid < len;
+        tid += blockDim.x * blockDim.y * blockDim.z){
+      body(*(segment.begin() + tid));
+    }
+  }
+};
+
```
tomstitt, line 312: Don't you need to use the hip vars here? Like `hipThreadIdx_x` instead of `threadIdx.x`? Here's a list: https://rocmdocs.amd.com/en/latest/Programming_Guides/HIP-GUIDE.html#built-in-variables

include/RAJA/policy/cuda/teams.hpp
```diff
@@ -270,6 +270,48 @@ struct LoopExecute<cuda_global_thread<DIM0, DIM1, DIM2>, SEGMENT> {
...
+      BODY const &body)
+  {
+    const int len = segment.end() - segment.begin();
+    {
+      const int tid = threadIdx.x + blockDim.x*(threadIdx.y + blockDim.y*threadIdx.z);
+
+      if (tid < len) body(*(segment.begin() + tid));
+    }
+  }
+};
```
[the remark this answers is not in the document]
artv3: Yes. In some cases its actually been observed that expressing the code in terms of a for loop adds some overhead. Thus if we know our range is smaller than the number of threads per block we can just use this policy and avoid the overhead.

include/RAJA/policy/hip/teams.hpp
artv3: We have been using threadIdx.x in RAJA; they are both compatible.

tomstitt: In that case the routines are exactly the same so can we just have one in a common file (or include one file from the other?)

tomstitt: Also is the `threadIdx` struct something hip guarantees keeping in their api? I think the answer is probably yes if it's being used in other parts of RAJA

artv3: Yes!, that having a common file will be a part of a different PR.

tomstitt (dismissed):
Thanks @artv3 !

artv3:
Hi @rhornung67, can we merge this PR in?

ajkunen:
sorry, should have commented a while ago.

Could we make it more general?  what if you only want to flatten x and z?   or just y?

I think naming wise, it would be better rename

>  flatten_block_threads_direct

 to 

>  flatten_block_threads_xyz_direct

That way, we could extend it later to cover other subsets.

ajkunen (changes requested):
rename flatten_block_threads_direct to flatten_block_threads_xyz_direct

artv3:
Hi @ajkunen , I just pushed a more flexible framework. Could you let me know if that what you were looking for?

include/RAJA/policy/cuda/teams.hpp
```diff
@@ -381,6 +381,137 @@ struct LoopExecute<cuda_global_thread<DIM0, DIM1, DIM2>, SEGMENT> {
...
+using flatten_block_threads_yz_loop = flatten_block_threads_loop<1,2>;
+using flatten_block_threads_zx_loop = flatten_block_threads_loop<2,0>;
+using flatten_block_threads_zy_loop = flatten_block_threads_loop<2,1>;
+
+using flatten_block_threads_xyz_loop = flatten_block_threads_loop<0,1,2>;
+using flatten_block_threads_xzy_loop = flatten_block_threads_loop<0,2,1>;
+using flatten_block_threads_yxz_loop = flatten_block_threads_loop<1,0,2>;
+using flatten_block_threads_yzx_loop = flatten_block_threads_loop<1,2,0>;
+using flatten_block_threads_zxy_loop = flatten_block_threads_loop<2,0,1>;
+using flatten_block_threads_zyx_loop = flatten_block_threads_loop<2,1,1>;
```
tomstitt, line 419: it sounds like you're going to combine these files in the future but this set of definitions is the same in both as far as I can tell and it could be better to only have it in one place

tomstitt (dismissed):

artv3:
ping @ajkunen

PR #1080, opened 2021-06-14
Fix workgroup zero length synchronize bug, and bring resource usage into reduction runtime

MrBurmark:
# Summary
Fix a bug that happens if only zero length loops were enqueued in a cuda or hip workgroup that would cause an error to be emitted from synchronize stating that it did not recognize the stream in use. This was causing some workgroup tests to fail intermittently with cuda and hip.
This also streamlines some parts of the operations around synchronize internally, and also pushes the use of resources deeper into the RAJA runtime.

This is built on #1024 .

- This PR is a bugfix,
- It does the following:
  - Fixes intermittent error in cuda and hip workgroup tests


artv3:
If its a small change, could this be bundled with https://github.com/LLNL/RAJA/pull/1024?

MrBurmark:
@artv3 It could be and that's how I originally had it, but I made enough changes before I figured out the root cause of the problem that I thought it was better as a separate PR.

rhornung67 (approved):

artv3 (approved):
Looks good, but does the title of the PR need to be updated? Looks like some good refactoring was also done. For example async is now done within the ::launch calls.

MrBurmark:
@artv3 That is true, this is mostly refactoring now so I updated the name to match more of what was changed.

PR #1085, opened 2021-06-21
Improving Gitlab CI performance.

rchen20:
# Summary

- This PR is an improvement
- It does the following:
  - Reduces CPU resources requested during each build of the Gitlab CI. Hopefully build_and_test times will improve.
  - Request max number of CPU resources per allocation, rather than specifying limited amounts.
  - Discussed with @adrienbernede and @kab163. The strategy is to first throw more resources at the jobs, while restricting the build phases to fewer cores to allow more tasks to run in parallel. We would like the inherent scheduler to distribute the work for us, rather than maintaining a complicated set of scripts and dependencies. Failing that, we may need to sub-divide the current stages into more stages and jobs.

### DIFF
--- a/.gitlab/ruby-templates.yml
+++ b/.gitlab/ruby-templates.yml
@@ -23,13 +23,14 @@
 
 ####
 # In pre-build phase, allocate a node for builds
+# NOTE: Not specifying 'salloc -c 56' should allocate the max number of CPU cores
 allocate_resources (on ruby):
   variables:
     GIT_STRATEGY: none
   extends: .on_ruby
   stage: r_allocate_resources
   script:
-    - salloc -N 1 -c 36 -p pdebug -t 45 --no-shell --job-name=${ALLOC_NAME}
+    - salloc -N 1 -p pdebug -t 45 --no-shell --job-name=${ALLOC_NAME}
 
 ####
 # In post-build phase, deallocate resources
--- a/.gitlab/ruby-templates.yml
+++ b/.gitlab/ruby-templates.yml
@@ -31,6 +31,7 @@ allocate_resources (on ruby):
   stage: r_allocate_resources
   script:
     - salloc -N 1 -p pdebug -t 45 --no-shell --job-name=${ALLOC_NAME}
+    
 
 ####
 # In post-build phase, deallocate resources
@@ -52,4 +53,3 @@ release_resources (on ruby):
 
 .build_and_test_on_ruby_advanced:
   extends: [.build_and_test_on_ruby, .advanced_pipeline]
-
--- a/.gitlab/ruby-templates.yml
+++ b/.gitlab/ruby-templates.yml
@@ -31,7 +31,6 @@ allocate_resources (on ruby):
   stage: r_allocate_resources
   script:
     - salloc -N 1 -p pdebug -t 45 --no-shell --job-name=${ALLOC_NAME}
-    
 
 ####
 # In post-build phase, deallocate resources
--- a/scripts/gitlab/build_and_test.sh
+++ b/scripts/gitlab/build_and_test.sh
@@ -12,6 +12,7 @@ set -o nounset
 
 option=${1:-""}
 hostname="$(hostname)"
+truehostname=${hostname//[0-9]/}
 project_dir="$(pwd)"
 
 build_root=${BUILD_ROOT:-""}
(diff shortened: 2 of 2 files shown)


scripts/gitlab/build_and_test.sh
```diff
@@ -112,14 +112,15 @@ then
...
+    # NOTE: 'cmake --build . -j 12' is an attempt to reduce individual build resources
     rm -rf ${build_dir} 2>/dev/null
     mkdir -p ${build_dir} && cd ${build_dir}
 
     date
     cmake \
       -C ${hostconfig_path} \
       ${project_dir}
-    cmake --build . -j 32
+    cmake --build . -j 12
```
adrienbernede, line 123: I now remember why 32 was chosen:
`build_and_test.sh` runs both on quartz and lassen, and lassen has 64 cores per nodes.
I think we should consider having a value that depends on the hostname. (using a mapping table?).
It seems that 24 would be reasonable on `ruby`. 32 was the usual choice on `lassen`, and 12 worked well on `quartz`.

kab163: I think that's a great idea!

scripts/gitlab/build_and_test.sh
```diff
@@ -111,15 +112,21 @@ then
     echo "~~~~~ Building RAJA"
     echo "~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~"
 
+    # Map CPU core allocations
+    declare -A core_counts=(["lassen"]=32 ["ruby"]=24 ["corona"]=32)
```
rchen20, line 116:
```suggestion
    declare -A core_counts=(["lassen"]=40 ["ruby"]=24 ["corona"]=32)
```

rchen20:
For some reason the XL gcc and XL CUDA jobs were exceeding their allocation times. I've upped the lassen build core count from 32 -> 40.

.gitlab/ruby-templates.yml
```diff
@@ -23,13 +23,14 @@
...
 # In pre-build phase, allocate a node for builds
+# NOTE: Not specifying 'salloc -c 56' should allocate the max number of CPU cores
 allocate_resources (on ruby):
   variables:
     GIT_STRATEGY: none
   extends: .on_ruby
   stage: r_allocate_resources
   script:
-    - salloc -N 1 -c 36 -p pdebug -t 45 --no-shell --job-name=${ALLOC_NAME}
+    - salloc -N 1 -p pdebug -t 45 --no-shell --job-name=${ALLOC_NAME}
```
rchen20, line 33:
```suggestion
    - salloc -N 1 -p pdebug -t 45 --no-shell --job-name=${ALLOC_NAME}
    
```

kab163 (approved):
Assuming Adrien agree with the core_counts numbers (looks like they work!), this looks great to me! Thank you!

rchen20:
> Assuming Adrien agrees with the core_counts numbers (looks like they work!), this looks great to me! Thank you!

I ran it twice and R_build_and_test timed out on a couple gcc jobs. Should we increase the core count on ruby?

adrienbernede:
> Assuming Adrien agrees with the core_counts numbers (looks like they work!), this looks great to me! Thank you!

Nooo! Don't ask me... I'm gonna be obsessed by the fact 40 does not fit very well in 128... lol.
As long as the overall time is the same or shorter, I'm fine with it!

kab163:
> > Assuming Adrien agrees with the core_counts numbers (looks like they work!), this looks great to me! Thank you!
> 
> I ran it twice and R_build_and_test timed out on a couple gcc jobs. Should we increase the core count on ruby?

Oh I see. Yea sure, see what happens lol (Maybe that needs 32)

adrienbernede:
> > Assuming Adrien agrees with the core_counts numbers (looks like they work!), this looks great to me! Thank you!
> 
> I ran it twice and R_build_and_test timed out on a couple gcc jobs. Should we increase the core count on ruby?

28 is a good number on Ruby ;)

scripts/gitlab/build_and_test.sh
```diff
@@ -111,15 +112,21 @@ then
     echo "~~~~~ Building RAJA"
     echo "~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~"
 
+    # Map CPU core allocations
+    declare -A core_counts=(["lassen"]=40 ["ruby"]=24 ["corona"]=32)
```
rchen20, line 116:
```suggestion
    declare -A core_counts=(["lassen"]=40 ["ruby"]=28 ["corona"]=32)
```

.gitlab/ruby-templates.yml
```diff
@@ -23,13 +23,15 @@
...
+
```
rchen20, line 34:
```suggestion
```

rchen20:
@kab163 @adrienbernede Looks like 28 was the magic core count for ruby. This should be ready to go.

kab163 (approved):

kab163:
> @kab163 @adrienbernede Looks like 28 was the magic core count for ruby. This should be ready to go.

Sounds great!

adrienbernede:
For the record it is possible to re-trigger the pipeline in Gitlab UI, no need for micro-commits ;)

adrienbernede (approved):

rchen20:
> For the record it is possible to re-trigger the pipeline in Gitlab UI, no need for micro-commits ;)

Ah thanks, I did not see the "Retry" button until now.

adrienbernede:
Retry may only re-run the failing jobs. But you can go to "CI/CD - pipelines", click "run pipeline", choose your branch and click "run". This will trigger a new pipeline, but it’s status will report to Github just fine.

PR #1089, opened 2021-06-24
Fix synchronization in examples

artv3:
Folks, I noticed that with the transition of running team kernels on the CAMP default stream we introduced some failing examples. This occurs since our examples draw from the CUDA default stream (different than CAMP default stream). 

This PR adds a fix by adding synchronization after kernels.

### DIFF
--- a/examples/raja-teams.cpp
+++ b/examples/raja-teams.cpp
@@ -42,11 +42,11 @@ using launch_policy = RAJA::expt::LaunchPolicy<
 #endif
 #if defined(RAJA_ENABLE_CUDA)
     ,
-    RAJA::expt::cuda_launch_t<true>
+    RAJA::expt::cuda_launch_t<false>
 #endif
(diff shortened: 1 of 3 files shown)


trws:
I assume the false tells it to sync? For the tut matmult one, adding a device sync to the checkResults function should take care of it for all of them if async is desirable there.

trws:
Actually, that might be nice for the teams one too, or for that one actually use a resource so the async behavior works correctly.

artv3:
> I assume the false tells it to sync? For the tut matmult one, adding a device sync to the checkResults function should take care of it for all of them if async is desirable there.

Correct, I think leaving the policies as synchronous is fine under tut variant. We can cover async execution in other examples; like "-Resource"

mdavis36 (approved):

MrBurmark (approved):
It looks like these lacked synchronization before. It looks like you can use the policy or synchronize in the check function.

PR #1091, opened 2021-07-08
changes to CMake to check the versions of cuda and hip

### LINKED ISSUE #947: add version check for cuda, maybe also hip etc, in cmake
Our build failure with a version of cuda that's outside our supported range currently prints a big mess of errors that don't point to the cause very well.   It would be good to add some logic to catch it and print a more meaningful error.

brandonneth:
This PR is a build system improvement.
It does the following:
* Checks the CUDA and HIP version numbers against the minimums given in the documentation.
* Fixes #947


rhornung67 (approved):
Thanks @brandonneth

trws (approved):
Looks great, thanks @brandonneth!

Issue #1093, opened 2021-07-16
omptarget build on lassen


liruipeng:
Hi, I found error when building on lassen with 

```
scripts/lc-builds/blueos_xl_omptarget.sh 2021.03.31
```

`2020.09.17` seems to be fine.

trws:
Fixed by #1094.

PR #1094, opened 2021-07-20
OpenMP Target Resource Fixes

### LINKED ISSUE #1092: OpenMP Target Build Issue
When building with OpenMP and Offloading enabled, test-kernel-basic-single-loop-Segments-OpenMPTarget.exe (and others) are hitting errors bc of a missing forall_impl implementation.

I've run into the issue with XL 16.01 and Clang 11.0.1, both on Lassen.

**Steps to Reproduce**
```
mkdir build
cd build
cmake -DENABLE_OPENMP=On -DENABLE_TARGET_OPENMP=On ..
make test-kernel-basic-single-loop-Segments-OpenMPTarget.exe
```

It looks like the problem is that the kernel call is using a Host resource, while the available omp_target forall_impls all expect a Omp resource.

**Fix**

adding the following to include/RAJA/policy/openmp_target/forall.hpp fixes the issue for me. All I did was copy the existing implementation and change the return and first argument's type (and name) from resources::Omp to resources::Host.

```
template <typename Iterable, typename Func>
RAJA_INLINE resources::EventProxy<resources::Host> forall_impl(resources::Host host_res,
                                                              const omp_target_parallel_for_exec_nt&,
                                                              Iterable&& iter,
                                                              Func&& loop_body)
{
  using Body = typename std::remove_reference<decltype(loop_body)>::type;
  Body body = loop_body;

  RAJA_EXTRACT_BED_IT(iter);

#pragma omp target teams distribute parallel for schedule(static, 1) \
    firstprivate(body,begin_it)
  for (decltype(distance_it) i = 0; i < distance_it; ++i) {
    Body ib = body;
    ib(begin_it[i]);
  }

  return resources::EventProxy<resources::Host>(host_res);
}
```

brandonneth:
# Summary

- This PR is a bug fix 
- It does the following:
  - Fixes #1092 and #1093 
  - Fixes resource selection process for nested kernels.

The existence of these bugs also suggests there should be more CI/CD testing of OpenMP target offloading, as without these changes, OpenMP target offloading tests do not build at all. 
 
Thank you @trws for the help figuring out what was causing the issues!

### DIFF
--- a/include/RAJA/pattern/kernel/For.hpp
+++ b/include/RAJA/pattern/kernel/For.hpp
@@ -101,7 +101,8 @@ struct StatementExecutor<
     auto len = segment_length<ArgumentId>(data);
     using len_t = decltype(len);
 
-    auto r = data.res;
+    using EnclosedResourceNeed = resources::resource_from_pol_t<ExecPolicy>;
+    auto r = EnclosedResourceNeed{};
 
     forall_impl(r, ExecPolicy{}, TypedRangeSegment<len_t>(0, len), for_wrapper);
   }
--- a/include/RAJA/pattern/kernel/For.hpp
+++ b/include/RAJA/pattern/kernel/For.hpp
@@ -101,8 +101,7 @@ struct StatementExecutor<
     auto len = segment_length<ArgumentId>(data);
     using len_t = decltype(len);
 
-    using EnclosedResourceNeed = resources::resource_from_pol_t<ExecPolicy>;
-    auto r = EnclosedResourceNeed{};
+    auto r = data.res;
 
     forall_impl(r, ExecPolicy{}, TypedRangeSegment<len_t>(0, len), for_wrapper);
   }
--- a/include/RAJA/policy/loop/forall.hpp
+++ b/include/RAJA/policy/loop/forall.hpp
@@ -52,8 +52,8 @@ namespace loop
 //
 
 
-template <typename Iterable, typename Func>
-RAJA_INLINE resources::EventProxy<resources::Host> forall_impl(RAJA::resources::Host host_res,
+template <typename Iterable, typename Func, typename Resource>
+RAJA_INLINE resources::EventProxy<Resource> forall_impl(Resource res,
                                                     const loop_exec &,
                                                     Iterable &&iter,
                                                     Func &&body)
@@ -63,9 +63,8 @@ RAJA_INLINE resources::EventProxy<resources::Host> forall_impl(RAJA::resources::
   for (decltype(distance_it) i = 0; i < distance_it; ++i) {
     body(*(begin_it + i));
   }
-  return RAJA::resources::EventProxy<resources::Host>(host_res);
+  return RAJA::resources::EventProxy<Resource>(res);
 }
-
 }  // namespace loop
 
 }  // namespace policy
--- a/include/RAJA/policy/openmp_target/policy.hpp
+++ b/include/RAJA/policy/openmp_target/policy.hpp
@@ -35,29 +35,32 @@ struct Collapse {
 
 template <size_t ThreadsPerTeam>
 struct omp_target_parallel_for_exec
-    : make_policy_pattern_t<Policy::target_openmp,
+    : make_policy_pattern_platform_t<Policy::target_openmp,
                             Pattern::forall,
+                            Platform::omp_target,
                             omp::Target,
                             omp::Teams<ThreadsPerTeam>,
                             omp::Distribute> {
 };
 
 struct omp_target_parallel_for_exec_nt
-    : make_policy_pattern_t<Policy::target_openmp,
+    : make_policy_pattern_platform_t<Policy::target_openmp,
                             Pattern::forall,
+                            Platform::omp_target,
                             omp::Target,
                             omp::Distribute> {
 };
 
 struct omp_target_parallel_collapse_exec
-    : make_policy_pattern_t<Policy::target_openmp,
+    : make_policy_pattern_platform_t<Policy::target_openmp,
                             Pattern::forall,
+                            Platform::omp_target,
                             omp::Target,
                             omp::Collapse> {
 };
 
 struct omp_target_reduce
-    : make_policy_pattern_t<Policy::target_openmp, Pattern::reduce> {
+    : make_policy_pattern_platform_t<Policy::target_openmp, Pattern::reduce, Platform::omp_target> {
 };
 
 ///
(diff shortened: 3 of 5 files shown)


rhornung67:
Thanks for the fixes @brandonneth I added some reviewers to your PR.

Part of the problem here may have been the result of spotty compiler support whereby code would work for one compiler version and not another (both XL and clang). Plus, we don't spend much time on the OpenMP target stuff because no one uses it in any serious way. It's not very well tested because our tests for this back-end take a very long time to build.

brandonneth:
@rhornung67 The build time is killer! That was the longest part of figuring out this issue.

include/RAJA/pattern/kernel/For.hpp
```diff
@@ -101,7 +101,8 @@ struct StatementExecutor<
     auto len = segment_length<ArgumentId>(data);
     using len_t = decltype(len);
 
-    auto r = data.res;
```
trws, line 104: Does this change remain necessary? The resource passed in here should be the right one when the user is passing one in, and we probably don't want to disable that.  Will the rest work if this change is pulled out?

trws (changes requested):
Aside from the one spot in For, this looks great.  Thanks again for finding and tracking this down @brandonneth!

include/RAJA/pattern/kernel/For.hpp
brandonneth: Changing it back breaks the following example:
```
Policy = KernelPolicy<
  For<0,seq,
    For<1,omp_target
      Lambda<0>>>>;

kernel<Policy>(...) // no resource specified
```

**The issue:**

kernel<Policy> creates a default resource based on Policy. 
Default resource creates the default based on get_platform<Policy>. 
This should basically resolve to platform_max(omp_target, host) = omp_target.
Default resource for omp_target is resource::Omp
So we have resource::Omp being passed to the LoopData constructor and then used throughout the execution.

When we get to the first StatementExecutor, which is for seq_exec, we're gonna call forall_impl for a seq_exec and resource::Omp. No such forall_impl exists.

**Alternative change:** 
Implement a forall_impl for a seq_exec policy and a omp resource. 

Do you anticipate implementing such a forall_impl to cause any other issues?

trws:
Ouch, I hadn’t thought about that, but it begs an interesting 
question.  Does that case work for a cuda inner policy? If it doesn’t 
then I’d say leave it broken but maybe file an issue about it.

If on the other hand it does work, then we’ll need to solve this.  If 
we do need to solve it I think the right thing to do is going to be 
somewhat complicated, and involve being able to take multiple resources 
so we can get ones of the right types for all the policies inside a 
kernel policy, which could get kinda rough.

brandonneth:
So the example above comes from test-kernel-nested-loop-Basic-OpenMPTarget.exe. Fixing it requires a forall_impl to be included for both loop_exec and seq_exec. I've added those (wrapped in a check for openmp) and they seem to work just fine. 

For CUDA, the current tests don't use this configuration, but if I add an example that does 
```
For<cuda_exec<64>,
  For<seq_exec
...
```
or
```
For<seq_exec
  For<cuda_exec<64>,
...
```

the code no longer compiles. The current tests use cuda_thread_{x,y}_loop, which do not include policy, pattern, launch, or platform information. 

I'm not sure what you mean by "if this case works", but I'm guessing that CUDA code not compiling is something we want to fix.

trws:
I'm ok with fixing it if we can do it consistently, but if it isn't supported for cuda that means nobody is using the ability to do this in production, and it might be better to defer trying to support this kind of multi-kernel case until we need it.  @rhornung67, @MrBurmark, @ajkunen have any of you run into a case where someone wants to be able to use kernel to run a sequential loop on the host over a number of kernel launches such that it would be valuable to allow that corner case?

brandonneth:
for fixing it consistently, I think forall_impl's for the product of (loop_exec,seq_exec) x (omp,cuda,hip) should cover it

MrBurmark:
@trws I don't have a use case for this.

trws:
It would cover all the serial cases, but do we then support the host threading cases?  I'm not sure.  My first thought would be we have a way to tag all the host policies such that we can make a forall_impl that will accept any of the host policies on the outside, and device policies on the inside if we go that way.

brandonneth:
Thats a good point, I think I’m leaning towards changing the policies we use in the test cases and adding some sort of spec requirement that policies and resources need to be compatible throughout the entire kernel policy.

rchen20:
Just ran in to this compilation issue while trying the newest compilers on OpenMP Target; thanks @brandonneth for looking in to this!

> I'm ok with fixing it if we can do it consistently, but if it isn't supported for cuda that means nobody is using the ability to do this in production, and it might be better to defer trying to support this kind of multi-kernel case until we need it. @rhornung67, @MrBurmark, @ajkunen have any of you run into a case where someone wants to be able to use kernel to run a sequential loop on the host over a number of kernel launches such that it would be valuable to allow that corner case?

@trws @rhornung67 I don't think we need this use case at the moment, although the code I work on has the following host threading with CUDA kernels implementation for using concurrent streams:

`#pragma omp parallel do`
`  cudastreamkernel<<<. . .>>>(. . .)`
`#pragma omp end parallel do`

Maybe we can implement this functionality another time, but for now if someone needs this, they can manually create omp host threads over RAJA CUDA kernels.

brandonneth:
The change I ended up going with was to template the resource type for the sequential forall_impl functions so you can have a nested loop that has different policy types (omptarget then seq or  cuda then seq), but doesn't allow mixing of policy types beyond that. If we don't want to have the sequential forall_impl support other resource types, then we should probably just remove the tests, because it's testing a feature we don't support.

trws:
That sounds quite reasonable to me.  Am I reading this right that it will take a resource for the portion that uses offload and just pass it through the outer sequential layers?

brandonneth:
@trws Yes. The default resource for a KernelPolicy is calculated like:
```
highestPriorityPlatform = max( [ExecPol.platform for ExecPol in KernelPolicy] )
return highestPriorityPlatform.defaultResource
```
Sequential is the lowest priority, mapping to resource::Host, while target_openmp is higher priority, mapping to resource::Omp. This highest priority resource is then passed through the entire loop nest as part of the LoopData data structure.  

cuda is a higher priority than omptarget, so the policy
For<..., seq,
  For<... omptarget,
     For<..., cuda,...>>>
would create a resources::Cuda, which would work fine for the outer For (due to the templating I added), but would break for the omptarget For, as those implementations require a resources::Omp.

trws (approved):

PR #1097, opened 2021-07-29
Cuda and Hip block reduction tweak

MrBurmark:
# Summary
Tweak the hip and cuda block reduction to fix the number of steps in the final warp reduction. This was not wrong before but was inconsistent with the maximum amount of data. This saves a couple rounds of warp shfls with hip. This also adds a static assert so we can change the block reduction algorithm if max number of warps exceeds warp size in the future.

- This PR is a bugfix
  - Fixes number of shfl steps in block reduction

### DIFF
--- a/include/RAJA/policy/cuda/reduce.hpp
+++ b/include/RAJA/policy/cuda/reduce.hpp
@@ -408,6 +408,9 @@ RAJA_DEVICE RAJA_INLINE T block_reduce(T val, T identity)
   // reduce per warp values
   if (numThreads > policy::cuda::WARP_SIZE) {
 
+    static_assert(policy::cuda::MAX_WARPS <= policy::cuda::WARP_SIZE,
+        "Max Warps must be less than or equal to Warp Size for this algorithm to work");
+
     // Need to separate declaration and initialization for clang-cuda
     __shared__ unsigned char tmpsd[sizeof(RAJA::detail::SoAArray<T, policy::cuda::MAX_WARPS>)];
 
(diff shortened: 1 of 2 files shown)


include/RAJA/policy/cuda/reduce.hpp
```diff
@@ -408,6 +408,9 @@ RAJA_DEVICE RAJA_INLINE T block_reduce(T val, T identity)
   // reduce per warp values
   if (numThreads > policy::cuda::WARP_SIZE) {
 
+    static_assert(policy::cuda::MAX_WARPS <= policy::cuda::WARP_SIZE,
```
artv3, line 411: Just to double check, do we have a way to test this?

artv3 (approved):

artv3:
This could potentially make things faster right?

include/RAJA/policy/cuda/reduce.hpp
MrBurmark: I'm not sure what you mean exactly, this is a static assert so it will fail at compile time if WARP_SIZE and MAX_WARPS change to invalid values.

MrBurmark:
Yes, hip was doing a 2 extra shfl reduce passes compared to cuda (12 vs 10) and is now doing the same number of passes.

include/RAJA/policy/cuda/reduce.hpp
artv3: Yes, I was wondering under what conditions we would trigger the assert. Doing a quick grep it looks like the values are hard-coded, so to trigger the assert we would have to manually change `MAX_BLOCK_SIZE` and `WARP_SIZE` . I was thinking the values would get picked up when we configure RAJA or something like that.

MrBurmark: Yep, if we ever change them we'll know to come back and fix this.

rhornung67 (approved):

PR #1103, opened 2021-08-12
Add SumTransaction option to halo-exchange example

nselliott:
# Summary

- This PR is a feature in example code
  - Adds SumTransaction class in halo-exchange as an alternative to CopyTransaction
  - Tests the results within the example driver


nselliott:
The last commit does a swap operation that makes the sum transactions do a true accumulation. The initial values are changed to square root of the previous values to make things more challenging for the order of operations.

MrBurmark (approved):

PR #1112, opened 2021-08-30
clangcuda generic script

MrBurmark:
# Summary
Add a generic clangcuda script. The clangcuda build still segfaults in many tests, but at least this lets us try newer versions more easily.

- This PR is a refactoring
- It does the following :
  - Refactors the clangcuda script to be generic like our other scripts


rhornung67 (dismissed):
We should this added to our azure tests and/or gitlab ci

davidbeckingsale (approved):

artv3 (dismissed):

davidbeckingsale (approved):
Awesome! I agree with @rhornung67 that we should get this added to CI.

PR #1120, opened 2021-09-15
We're compiling with pedantic, sorry for the noise

DavidPoliakoff:
# Summary

- This PR is a bugfix for pedantic compilers
  - removes an extra semicolon

Let me know if this is problematic, sorry for the noise

### DIFF
--- a/src/PluginStrategy.cpp
+++ b/src/PluginStrategy.cpp
@@ -7,7 +7,7 @@
 
 #include "RAJA/util/PluginStrategy.hpp"
 
-RAJA_INSTANTIATE_REGISTRY(PluginRegistry);
+RAJA_INSTANTIATE_REGISTRY(PluginRegistry)
 
 namespace RAJA {
 namespace util {


rhornung67 (approved):

ajpowelsnl:
@DavidPoliakoff @trws -- Quick question:  can I pull in all of the changes to fix compiler warnings in from the "develop" branch?

trws:
Once this merges, this would get you the raja one and you'd probably need to update to the new camp version, not sure if we've pulled that here yet.

ajpowelsnl:
> Once this merges, this would get you the raja one and you'd probably need to update to the new camp version, not sure if we've pulled that here yet.

Morning @trws and @DavidPoliakoff -- Tom, many thanks for the update.  I am seeing compiler warnings from both RAJA and camp.  Would it be possible for you to also address compiler warnings from RAJA dependencies , as well?  I am grateful for your help with this.

Issue #1118, opened 2021-09-13
Add ROCTX support 


artv3:
Add ROCTX support to enable kernel naming where applicable for AMD platforms.

artv3:
Resolved in https://github.com/LLNL/RAJA/pull/1122

PR #1122, opened 2021-09-20
Configure roctx for AMD kernel naming

artv3:
Add support for ROCTX to enable kernel naming with RAJA Teams.


artv3:
Resolves https://github.com/LLNL/RAJA/issues/1118

CMakeLists.txt
```diff
@@ -236,6 +237,12 @@ if (ENABLE_HIP)
   set(raja_depends
     ${raja_depends}
     hip)
+  if(ENABLE_ROCTX)
+    include(FindRoctracer)
```
davidbeckingsale, line 241: I think you should move the FindRoctracer call into `cmake/SetupPackages.cmake`

cmake/SetupPackages.cmake
```diff
@@ -61,6 +61,9 @@ endif ()
 
 if (ENABLE_HIP OR ENABLE_EXTERNAL_ROCPRIM)
   find_package(RocPRIM)
+  blt_import_library(NAME roctx
```
davidbeckingsale, line 64: I think this should be in a separate conditional block (not inside the "OR ENABLE_EXTERNAL_ROCPRIM") - it should be guarded with ENABLE_ROCTX.

davidbeckingsale (changes requested):

CMakeLists.txt
artv3: Done!

davidbeckingsale (approved):

tomstitt (approved):
Thank you @artv3 !

PR #1133, opened 2021-10-15
Fix View pointer type.

rchen20:
# Summary

- This PR is a bugfix
- It does the following:
  - Fixes https://github.com/LLNL/Kripke/pull/30 and https://github.com/LLNL/Kripke/pull/32
  - The CHAI ManagedArray type was not being passed through to View. Instead, the raw pointer type was being used. This caused missing CHAI memory transfers. (Thanks @vsrana01 @davidbeckingsale!)
  - Note: For RAJA v0.14.0 and Kripke (and https://github.com/LLNL/RAJAProxies/pull/23), need to use internal::ViewBase to work around this problem. RAJA v0.15.0 should be able to use this current fix.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -38,6 +38,7 @@ variables:
   ALLOC_NAME: ${CI_PROJECT_NAME}_ci_${CI_PIPELINE_ID}
   BUILD_ROOT: ${CI_PROJECT_DIR}
   DEFAULT_TIME: 30
+  MP_BRANCH: "develop"
 
 # Normally, stages are blocking in Gitlab. However, using the keyword "needs" we
 # can express dependencies between job that break the ordering of stages, in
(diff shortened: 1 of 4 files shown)


davidbeckingsale:
This looks good to me. Do you think we should also expose PointerType in the TypedView? https://github.com/LLNL/RAJA/blob/28a1e441eda23c584b4e68edb58888bbef86e3b3/include/RAJA/util/View.hpp#L59

I cannot remember if we used to in previous releases. In CHAI, we currently use TypedViewBase to provide this functionality.

ajkunen:
@rchen20 @davidbeckingsale I think the issue with exposing the pointer type in TypedView was the variadic template list.  Perhaps we should have another TypeView alias... like TypedViewP that takes the type list as a camp::list, and provides the PointerType defaulted like View.
Something like:   
`template <typename ValueType, typename LayoutType, typename IndexTypes, typename PointerType = ValueType*>
using TypedViewP =
    internal::TypedViewBase<ValueType, PointerType, LayoutType, IndexTypes >;`

rchen20:
> This looks good to me. Do you think we should also expose PointerType in the TypedView?
> 
> https://github.com/LLNL/RAJA/blob/28a1e441eda23c584b4e68edb58888bbef86e3b3/include/RAJA/util/View.hpp#L59
> 
> I cannot remember if we used to in previous releases. In CHAI, we currently use TypedViewBase to provide this functionality.

When I looked at RAJA v0.10.0 yesterday, TypedViewBase was not exposed to PointerType, but I can add that in for good measure.

davidbeckingsale:
We should add that to RAJA and to CHAI. We can work with @kab163 to set this up.

PR #1135, opened 2021-10-18
Add a new Lassen CI job with ats disabled

kab163:
# Summary

- This PR adds a new CI job with ats disabled
- Tests a Lassen build with ats disabled (see PR #1133)


kab163:
> You also need to add a job to .gitlab/lassen-jobs.yml that extends your new `.build_and_test_on_lassen_ats_disabled` template

Lol you're moving too fast! I just figured that out haha fixing it now

PR #1155, opened 2021-11-05
Enable hip device functions automatically

MrBurmark:
# Summary
Remove the cmake option RAJA_ENABLE_HIP_INDIRECT_FUNCTION_CALL and choose to define it in the config based on the hip version instead.

- This PR is a feature
  - Adds automatic hip device call support detection

# Design review

I chose to remove the cmake option RAJA_ENABLE_HIP_INDIRECT_FUNCTION_CALL in favor of checking and defining/not defining the var in the config because my original intention was to not allow this to be configurable. It is also possible to make this user configurable in cmake with the default checking the compiler version in cmake to set RAJA_ENABLE_HIP_INDIRECT_FUNCTION_CALL appropriately.


davidbeckingsale (approved):
Nice!

rhornung67 (approved):

PR #1169, opened 2021-12-01
Add basic ND->1D holder class

MrBurmark:
# Summary
Add a class to turn a lambda meant to be used in nested loops into a lambda usable with a non-nested api.

NOTE: this PR requires c++14 so it will fail some tests until we require c++14.

This is intended to be used with RAJA workgroup to fuse multi-dimensional kernels.
The performance of using this holder class with forall is expected to be worse than using the lambda directly with a multi-dimensional loop abstraction like kernel.

The class stores the lambda and information about the given ranges. It will take ownership of the lambda, but stores the ranges as a tuple of spans and does not take ownership of the ranges.
It provides a 1D call operator that calls the lambda's ND call operator. The class uses division to break a one-dimensional index into n-dimensional indices.

- This PR is a feature
- It does the following:
  - Adds adapter to do multidimensional work in single dimensional loop abstractions at the request of myself and Samrai for use with fusible kernels


MrBurmark:
@nselliott

davidbeckingsale (changes requested):
Three questions:
- should we namespace this under `util` or something?
- is there a snappier name we can use?
- can you add tests/example showing the holder used with a forall?

MrBurmark:
- We have other classes in the util directory intended for user consumption like SPan that are not in the util namespace. Its a good idea, but we haven't done it consistently in the past.
- I agree the name needs to be improved, looking for suggestions.
- I plan to add more examples and tests next. There is an example of using this with forall in the class documentation.

trws:
I need to look over this some more, but my initial impressions are that this is basically a bind or collapse, maybe use one of those as a basis for the name?

Implementation-wise, why is this manually re-implementing big chunks of layout? The one used in View already supports translating both directions, and we have various conveniences for creating them that would remove the need for the extra range segments. Is there a need to support an indexset in one dimension or something?

-Tom (mobile)

MrBurmark:
I was thinking about using collapse, I'll change the name.

I was not aware that layout did this direction of translation already, I'll re-implement it. I only foresee using range segments, so no need for indexset.

MrBurmark:
@trws I've been trying to use layouts instead and hit a snag, why do layout convert 0 to 1 when converting size?

trws:
I’m not sure what you mean.  It uses 1 as the stride value when a size is zero, but it will store the zero size.  What’s causing you trouble?

MrBurmark:
The size() method will return a non-zero number of if there are one or more dimensions with size 0. I'll just make make another method that returns the actual size.

trws:
That's extremely peculiar.  It does beg the question though, why a dimension with zero size?

trws:
Also, it sounds like a bug... does fixing it break any tests?

MrBurmark:
I haven't tried fixing it. In my case I intend to use the dimensions like range segments where 0 is a valid size.

trws:
I suppose it is a valid size, it just means the total size is zero if any of the dimensions is zero right?

MrBurmark:
Right, that's what I was expecting.
Take a look at this test, it uses 0 in one of the dimensions. It doesn't call size, but it basically treats the 0 dimension as if it wasn't there. https://github.com/LLNL/RAJA/blob/develop/test/unit/view-layout/test-standard-layout.cpp#L129.

trws:
Ah, I’m betting that’s to support the projection stuff.  Subsetting basically by fixing a dimension, maybe that’s currently working by setting the size and/or stride to zero.  That actually does get used, so messing with it might cause issues, but then it might be worth untangling as well.

rhornung67:
Our current kernel implementation does no kernel operations if one of the segments in a nested loop, say, has zero length. For example, https://github.com/LLNL/RAJA/blob/develop/test/functional/kernel/nested-loop-segment-types/tests/test-kernel-nested-loops-segment-types.hpp#L194

A projection capability would be useful. However, it seems clearer to me if you can specify which index in the dimension you want to flatten (?); e.g., (i, 3, k) where i and k are iterates in some ranges in the first and third dimensions.

trws:
The projection thing is (I think) used for View. IIRC @ajkunen implemented it to ease passing say one plane of a 3D space for example, or one column or row of the plane.

trws:
Yeah, this stuff: #273 

I had the impression at the time that this was a common Ardra pattern, but can't speak to it now.

Then size was added later, and to support the notion of projections triggered by zero-length dimensions it has to treat zero-sized dimensions as size one.  So, options right now seem to be:

1. As @MrBurmark mentioned, add a different size method that gives the size of the represented iteration space rather than the size of the projection.
1. Give layout a flag, probably a template parameter, to tell it whether to do projection or not.  This would be relatively straightforward and if we add it with a default of true it shouldn't break anything, but then if a user wants that behavior they have to ask for it.
2. Do the above and add a `make_index_layout` or similar that creates a layout of that type, possibly with options on the sub-types of layouts.
3. Make the default that projection is *not* supported.  This is probably what people would expect from working directly with a layout, but would break an unknown amount of code.

ajkunen:
right... it's nice to have a layout that allows projections.  I can see arguments for supporting that or not... 

But perhaps it should return a 0-size if all dimensions are 0-size?  @MrBurmark would that fix the issue you're seeing? or are you expecting the size to be 0 if any dimension is 0?

MrBurmark:
I think number 2/3 might be a reasonable way to move forward for the moment.
I'm using this to run multidimensional loops in 1D abstractions so if any dimension is 0 I would expect size to return 0.

trws:
It probably should return 0 if all are 0 regardless now that I think about it.  Even the projected version should be empty if all dimensions are projected out right?

MrBurmark:
On another note, should OffsetLayout support a stride one dimension, it appears to be missing it?

rhornung67:
@MrBurmark I think that's a gap that should be filled. @ajkunen is there a reason that the stride-1 dim was left out of offset layout?

PR #1197, opened 2022-01-15
Kernel warp tests.

rchen20:
# Summary

- This PR is a refactoring
- It does the following:
  - Refactors kernel warp tests into modern testing framework


test/functional/kernel/warp-thread/tests/warp-thread-ReduceMask-impl.hpp
```diff
@@ -0,0 +1,185 @@
...
+  RAJA::ReduceSum<REDUCE_POL, RAJA::Index_type> worksum(0);
+
+  call_kernel_param<EXEC_POLICY, USE_RESOURCE>(
+                            RAJA::make_tuple(RAJA::TypedRangeSegment<RAJA::Index_type>(0, directlen), RAJA::TypedRangeSegment<RAJA::Index_type>(0, looplen)),
+                            RAJA::make_tuple((RAJA::Index_type)0, (RAJA::Index_type)0),
+                            work_res,
+                            [=] RAJA_DEVICE (RAJA::Index_type i, RAJA::Index_type j, RAJA::Index_type x, RAJA::Index_type y) {
+                              trip_count += 1;
+                              worksum += y; // y should only be 0..3
+                              max_thread.max(threadIdx.x);
```
artv3, line 132: Looking ahead towards SYCL (assuming it gets supported) I don't think `threadIdx.x` will be the right syntax. Can't we use one of the variables listed above to get the threadIdx? This would enable us  to insulate programing model specifics inside of RAJA.

rchen20: That's a good point, this particular test case exercises CUDA & HIP thread masking specifically, which are not going to applicable to other things like SYCL. I was going to name this set of tests something like "gpu-specific", but figured that "warp-thread" would be a bit more accurate.

artv3: Interesting, I figured since we could run SYCL on the GPU there would be something analogous, like threadIdx.x on a threadblock is a workItem on a workgroup.

artv3 (approved):

rhornung67 (approved):

PR #1242, opened 2022-04-04
Add cmake dependencies if none were specified in raja_add_executable.

rchen20:
# Summary

- This PR is a bugfix
  - While building CoMD in the RAJAProxies repo with HIP, RAJA attempts to add an extraneous `-lopenmp` link flag if RAJA was already installed with OpenMP enabled. HIP does not recognize this flag, and if the HIP build does not require OpenMP, this flag causes an error. Furthermore, the CoMD use of `raja_add_executable()` already passes in the necessary dependencies it requires.
- It does the following:
  - Changes RAJA's cmake setup to only add library dependencies if none are passed to `raja_add_executable()`.
  - NOTE: This would need to be patched into the v2022.03.x release.

### DIFF
--- a/cmake/RAJAMacros.cmake
+++ b/cmake/RAJAMacros.cmake
@@ -12,28 +12,29 @@ macro(raja_add_executable)
 
   cmake_parse_arguments(arg
     "${options}" "${singleValueArgs}" "${multiValueArgs}" ${ARGN})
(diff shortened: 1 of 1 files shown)


rchen20:
Looks like this change breaks everything. Closing this in favor of using `blt_add_executable` instead of `raja_add_executable` in CoMD, which also solves the problem.

PR #1243, opened 2022-04-04
Fix bug for cuda + host unrolling

artv3:
Need to undef RAJA_UNROLL_COUNT when inside a cuda kernel. This PR fixes a warning that comes up with clang + nvcc.


MrBurmark:
Would we need a similar fix for hip?

MrBurmark (approved):

artv3:
> Would we need a similar fix for hip?

It doesn't look we have the same case for HIP. From previous conversations, it seems unrolling in a hip kernel could actually hurt performance.

MrBurmark:
I'm not sure I understand why we wouldn't want to unroll in hip if the compiler could do it well, but if its faster I guess that's ok.

PR #1251, opened 2022-04-20
RAJA-Teams-Sycl

artv3:
# Summary

Code seems to stall when calling  `ctx.team->parallel_for_work_item` from a RAJA::expt::loop. Reproducer in `sycl_teams.cpp`


artv3:
Calling for each work item from an external function is not supported in SYCL.

PR #1260, opened 2022-05-05
Add GCC 8 and CUDA 11 job

davidbeckingsale:
- This PR adds a GCC 8 + CUDA 11 job to CI


davidbeckingsale:
I think this is unnecessary due to #1269

PR #1350, opened 2022-10-19
Feature/trws/clang cuda

trws:
Moving here so we run tests for #1293.  @MrBurmark, could you take a look at the workgroup plugin test file here? I'm not 100% sure the merge is as it should be.


MrBurmark:
@trws I don't see anything obviously wrong in there.

PR #1362, opened 2022-11-01
inline KernelName to avoid multiply defined symbol

MrBurmark:
# inline KernelName to avoid multiply defined symbol

- This PR is a bugfix
  - Fixes (https://github.com/LLNL/RAJAPerf/pull/272)


artv3 (approved):

rhornung67 (approved):
D'oh! We must not be testing this....

PR #1363, opened 2022-11-01
Update build scripts for later use with RAJAPerf

MrBurmark:
# Update build scripts far later use with RAJAPerf

Update scripts with some minor features from RAJAPerf with the intention of using these scripts in RAJAPerf instead of duplicating them.
- Default to using c++14.
- fix build directory name prints.
- Add rocm asan script.

- This PR is a feature
  - Adds build script RAJAPerf compatability at the request of (project or person)


rhornung67:
How are these scripts going to work in RAJAPerf since the host-config file location is hard-coded when passed to cmake via '-C' option?

MrBurmark:
Adding a host-configs symlink in the the top level of RAJAPerf worked.

rhornung67:
> Adding a host-configs symlink in the the top level of RAJAPerf worked.

I get it now.

rhornung67 (approved):

artv3 (approved):

PR #1382, opened 2022-11-14
Use compile time blocksize in forall

MrBurmark:
# Use compile time blocksize in forall
This improves performance of hip forall kernels.

- This PR is a feature
- It does the following:
  - Adds better hip forall performance at the request of me


rhornung67 (approved):
Looks good @MrBurmark Thanks for tracking this down and fixing.

PR #1386, opened 2022-11-17
Replace DEVICE macro with using alias

artv3:
Fixes name conflicts reported with MFEM in: https://github.com/mfem/mfem/pull/3293

### DIFF
--- a/include/RAJA/pattern/params/reducer.hpp
+++ b/include/RAJA/pattern/params/reducer.hpp
@@ -5,11 +5,11 @@
 #include "RAJA/util/SoAPtr.hpp"
 
 #if defined(RAJA_CUDA_ACTIVE)
-#define RAJA_DEVICE_BACKEND cuda
 #include "RAJA/policy/cuda/MemUtils_CUDA.hpp"
+using device_mem_pool_t = RAJA::cuda::device_mempool_type;
 #elif defined(RAJA_HIP_ACTIVE)
-#define RAJA_DEVICE_BACKEND hip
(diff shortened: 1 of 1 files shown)


include/RAJA/pattern/params/reducer.hpp
```diff
@@ -91,7 +91,7 @@ namespace detail
 #if defined(RAJA_CUDA_ACTIVE) || defined(RAJA_HIP_ACTIVE)
     // Device related attributes.
     value_type * devicetarget = nullptr;
-    RAJA::detail::SoAPtr<value_type, RAJA::DEVICE::device_mempool_type> device_mem;
+    RAJA::detail::SoAPtr<value_type, RAJA::RAJA_DEVICE_BACKEND::device_mempool_type> device_mem;
```
davidbeckingsale, line 94: Why were you using a macro here rather than a 'using' statement for the complete type in the same place you are defining the macro?

artv3: I'll defer to the author - ping @mdavis36

artv3: But I like your idea @davidbeckingsale, I propose we use that.

MrBurmark: We can do that for now, long term we should remove the mempool_type from this type.

MrBurmark (approved):

rhornung67 (approved):
Thanks @artv3

davidbeckingsale (approved):

PR #1396, opened 2022-12-01
move device_mem_pool_t to raja detail namespace

artv3:
Addresses: https://github.com/LLNL/RAJA/pull/1393#discussion_r1037288810

### DIFF
--- a/include/RAJA/pattern/params/reducer.hpp
+++ b/include/RAJA/pattern/params/reducer.hpp
@@ -6,10 +6,8 @@
 
 #if defined(RAJA_CUDA_ACTIVE)
 #include "RAJA/policy/cuda/MemUtils_CUDA.hpp"
-using device_mem_pool_t = RAJA::cuda::device_mempool_type;
 #elif defined(RAJA_HIP_ACTIVE)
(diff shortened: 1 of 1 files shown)


include/RAJA/pattern/params/reducer.hpp
```diff
@@ -66,6 +64,14 @@ struct limits<RAJA::expt::ValLoc<T>> {
...
 namespace RAJA
 {
+namespace detail
+{
+#if defined(RAJA_CUDA_ACTIVE)
+using device_mem_pool_t = RAJA::cuda::device_mempool_type;
+#elif defined(RAJA_HIP_ACTIVE)
+using device_mem_pool_t = RAJA::hip::device_mempool_type;
+#endif
+} //namespace detail
```
artv3, line 74: @MrBurmark I moved it closer to where it is being used, let me know if there is a preference.

rhornung67 (approved):

include/RAJA/pattern/params/reducer.hpp
MrBurmark: If its only used here you could put it in the expt and detail namespaces right above Reducer.

MrBurmark (approved):

PR #1398, opened 2022-12-05
Add a build script for hipcc 

CRobeck:
This PR adds a script to build with hipcc as a temporary work around for RAJA test failures from missing symbols when building with amdclang. Building with hipcc resolves the missing symbol runtime test errors.


MrBurmark:
@CRobeck How about forking the file so we keep both options?

PR #1414, opened 2022-12-28
Fix MultiLambda GPU test

### LINKED ISSUE #988: Known-Hip-Failures
New HIP CI tests are going to be added to the RAJA gitlab CI. Although most tests pass, there are a couple of tests with known HIP failures. In order to have a test that gives reasonable information about whether or not the majority of the HIP functionality is working ok, I decided to label those tests which have known HIP failures with "Known-Hip-Failure" in the corresponding CMakeLists.txt. So, every test that we know won't work with HIP, the name of the test exe has the label.

When running the CI, the script runs a ctest command with:
`ctest --output-on-failure -T test -E Known-Hip-Failure`

This means that ctest will exclude (with "-E") all tests labeled with "Known-Hip-Failure". The ctest output doesn't actually show any tests as being "skipped", so if not tracked right, these tests could accidentally be phased out, even after the known failures are fixed. However, as long as the ctest command has a "-E" argument, something is being skipped.

Currently, PR #964 shows the CMakeLists files that were modified to make the changes to the test exe names. Those tests that have known HIP failures include:
         23 - test-plugin-workgroup-Hip.exe 
	116 - test-kernel-nested-loop-MultiLambda-Hip.exe 
	189 - test-algorithm-util-sort-Heap-Hip.exe 
	191 - test-algorithm-util-sort-Insertion-Hip.exe

See test output here: https://lc.llnl.gov/gitlab/radiuss/RAJA/-/jobs/174629

MrBurmark:
# Fix race condition in MultiLambda GPU policy

The gpu policy ran two dependent nested loops in the same kernel.
This is a race condition so it now uses two kernels so one nested loop is done before the other.

This fixes a known failure with hip.
This was passing for cuda, but it also had the same race condition.

- This PR is a bugfix
  - Fixes #988

### DIFF
--- a/test/functional/kernel/nested-loop/tests/nested-loop-MultiLambda-impl.hpp
+++ b/test/functional/kernel/nested-loop/tests/nested-loop-MultiLambda-impl.hpp
@@ -172,7 +172,9 @@ struct MultiLambdaNestedLoopExec<DEVICE_DEPTH_2, POLICY_DATA> {
           RAJA::statement::For<1, typename camp::at<POLICY_DATA, camp::num<1>>::type,
             RAJA::statement::Lambda<0>
           >
-        >,
+        >
+      >,
+      RAJA::statement::DEVICE_KERNEL<
         RAJA::statement::For<0, typename camp::at<POLICY_DATA, camp::num<0>>::type,
           RAJA::statement::For<1, typename camp::at<POLICY_DATA, camp::num<1>>::type,
             RAJA::statement::Lambda<1>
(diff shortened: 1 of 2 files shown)


rchen20 (approved):

test/functional/kernel/nested-loop/tests/nested-loop-MultiLambda-impl.hpp
```diff
@@ -172,7 +172,9 @@ struct MultiLambdaNestedLoopExec<DEVICE_DEPTH_2, POLICY_DATA> {
           RAJA::statement::For<1, typename camp::at<POLICY_DATA, camp::num<1>>::type,
             RAJA::statement::Lambda<0>
           >
-        >,
+        >
+      >,
+      RAJA::statement::DEVICE_KERNEL<
```
artv3, line 177: what was the issue? was it essentially launching instructions for two kernels?

artv3 (approved):

test/functional/kernel/nested-loop/tests/nested-loop-MultiLambda-impl.hpp
MrBurmark: I updated the description, but the two loop nests are dependent.

PR #1415, opened 2022-12-29
Fix cuda and hip reduceloc reset and make interface consistent

MrBurmark:
# Fix cuda and hip reduceloc reset and make interface consistent

There was a problem with the cuda and hip minloc and maxloc reducers. If you called reset you would change the identity value leading to incorrect results in some cases. This was caused by accidentally calling a base class reset method that treated the initial location argument as the identity. This PR adds a reset method in the derived class that hides the base class method.

In addition the arguments to all reduce loc constructors and resets were inconsistent. They now all take an initial value and location and an identity value and location.

The forall segment multiple reduce tests had some duplication and some potential for error in places so I refactored those to reduce duplication and fixed some potential for errors like the test assumed that first would always be 0.

  - Modifies/refactors
    - reduce loc constructors and reset methods 
    - the forall segment multiple reduce tests
  - Fixes cuda and hip loc reducer reset

### DIFF
--- a/include/RAJA/pattern/detail/reduce.hpp
+++ b/include/RAJA/pattern/detail/reduce.hpp
@@ -355,9 +355,18 @@ public:
   constexpr BaseReduceMinLoc() : Base(value_type(T(), IndexType())) {}
 
   constexpr BaseReduceMinLoc(T init_val, IndexType init_idx,
-                             T identity_ = reduce_type::identity())
-    : Base(value_type(init_val, init_idx), identity_)
+                             T identity_val_ = reduce_type::identity(),
+                             IndexType identity_loc_ = DefaultLoc<IndexType>().value())
+    : Base(value_type(init_val, init_idx), value_type(identity_val_, identity_loc_))
   {
+  }
+
+  void reset(T init_val, IndexType init_idx = DefaultLoc<IndexType>().value(),
+             T identity_val_ = reduce_type::identity(),
+             IndexType identity_loc_ = DefaultLoc<IndexType>().value())
+  {
+    operator T(); // automatic get() before reset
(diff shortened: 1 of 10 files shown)


artv3 (approved):

include/RAJA/pattern/detail/reduce.hpp
```diff
@@ -355,11 +355,20 @@ class BaseReduceMinLoc
...
   constexpr BaseReduceMinLoc(T init_val, IndexType init_idx,
-                             T identity_ = reduce_type::identity())
-    : Base(value_type(init_val, init_idx), identity_)
+                             T identity_val_ = reduce_type::identity(),
+                             IndexType identity_loc_ = DefaultLoc<IndexType>().value())
+    : Base(value_type(init_val, init_idx), value_type(identity_val_, identity_loc_))
   {
   }
 
+  void reset(T init_val, IndexType init_idx = DefaultLoc<IndexType>().value(),
```
rchen20, line 364: Thanks for catching these hidden problems! For the loc reductions, should we require the `init_idx` as a second reset argument to distinguish it from a normal non-loc reduction?

MrBurmark: I think that's a good idea but it would be a change.

rchen20: Yeah, it would change the interface subtly. I'm fine with we have now too.

MrBurmark: How about we leave it as it was for now, then change it for the next release?

rchen20: > How about we leave it as it was for now, then change it for the next release?

Sounds good to me.

rchen20 (approved):

PR #1416, opened 2022-12-29
Add launch shared memory test (dynamic and static)

artv3:
This PR also fixes a bug, in renaming Teams->Launch I missed a few items and certain test were not being triggered.


test/functional/launch/shared_mem/tests/test-launch-DynamicMem.hpp
```diff
@@ -0,0 +1,111 @@
...
+                                     working_res,
+                                     &working_array,
+                                     &check_array,
+                                     &test_array);
+  
+  
+  for(int b=0; b<RAJA::stripIndexType(block_range); ++b) {
+    for(int c=0; c<RAJA::stripIndexType(thread_range); ++c) {
+      int idx = c + RAJA::stripIndexType(thread_range)*b;
+      test_array[idx] = INDEX_TYPE(b);
```
MrBurmark, line 37: Would it make more sense to set them to idx instead of b so they are each unique?

test/functional/launch/shared_mem/tests/test-launch-DynamicMem.hpp
```diff
@@ -0,0 +1,111 @@
...
+                        RAJA::Threads(RAJA::stripIndexType(thread_range)), shared_mem_size),
+     [=] RAJA_HOST_DEVICE(RAJA::LaunchContext ctx) {
+
+      RAJA::loop<TEAM_POLICY>(ctx, outer_range, [&](INDEX_TYPE bid) {
+
+          INDEX_TYPE * tile_ptr = ctx.getSharedMemory<INDEX_TYPE>(1);
+          RAJA::View<INDEX_TYPE, RAJA::Layout<1>> Tile(tile_ptr, 1);
+
+
+          RAJA::loop<THREAD_POLICY>(ctx, inner_range, [&](INDEX_TYPE tid) {
```
MrBurmark, line 54: tid unused here

artv3: ah yes let me fix that. Thanks.

test/functional/launch/shared_mem/tests/test-launch-DynamicMem.hpp
artv3: Well the test I had in mind was to do a broadcast with shared memory. For example in shared memory store the block number (that would use an array of size one) and then each thread can read from it. That would test that the threads can indeed read from a shared location unique to each block (so threads from different blocks would be reading a different value).

artv3: Just noticed what was pushed was not what I had in mind, just pushed the correct implementation.

MrBurmark: That is a good idea to actually read from memory set by a different thread. How about you use idx, but set it in a different thread?

test/functional/launch/shared_mem/tests/test-launch-DynamicMem.hpp
```diff
@@ -0,0 +1,110 @@
...
+     [=] RAJA_HOST_DEVICE(RAJA::LaunchContext ctx) {
+
+      RAJA::loop<TEAM_POLICY>(ctx, outer_range, [&](INDEX_TYPE bid) {
+
+          INDEX_TYPE * tile_ptr = ctx.getSharedMemory<INDEX_TYPE>(1);
+          RAJA::View<INDEX_TYPE, RAJA::Layout<1>> Tile(tile_ptr, 1);
+
+
+          RAJA::loop<THREAD_POLICY>(ctx, RAJA::TypedRangeSegment<INDEX_TYPE>(0,1), [&](INDEX_TYPE ) {
+              Tile(0) = bid;
```
MrBurmark, line 55: For example if you got a buffer of size thread_range,
you could do `Tile(thread_range-tid-1) = thread_range-tid-1 + RAJA::stripIndexType(thread_range)*b` here,
then use `working_array[RAJA::stripIndexType(idx)] = Tile(tid);` below.

artv3: Sure, that works, just pushed it up.

test/functional/launch/shared_mem/tests/test-launch-DynamicMem.hpp
artv3: Done, I used your idea from below.

artv3:
I am seeing an odd warning here, and I think it has to do with the fact that we are using a strong index type: 

```
test/functional/launch/shared_mem/tests/test-launch-StaticMem.hpp(51): warning #3019-D: dynamic initialization is not supported for a function-scope static __shared__ variable within a __device__/__global__ function
```

after some reading I think it has to do with a potential race condition in initializing the shared memory. The question ends up being "which thread initializes the shared memory of the INDEX_TYPE" . Seems to be an issue with the static shared memory type, not dynamic.

MrBurmark:
@artv3 I think you are right, I've seen this before with static shared memory and types with non-trivial constructors/destructors.
Dynamic shared memory doesn't have this problem because its just memory instead of an array of objects.
I've worked around this by making a shared array of chars then making a pointer to the memory of an appropriate type and manually constructing or destructing objects if necessary.

test/functional/launch/shared_mem/tests/test-launch-DynamicMem.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+  RAJA::launch<LAUNCH_POLICY>
+    (RAJA::LaunchParams(RAJA::Teams(RAJA::stripIndexType(block_range)),
+                        RAJA::Threads(RAJA::stripIndexType(thread_range)), shared_mem_size),
+     [=] RAJA_HOST_DEVICE(RAJA::LaunchContext ctx) {
+
+      RAJA::loop<TEAM_POLICY>(ctx, outer_range, [&](INDEX_TYPE bid) {
+
+          INDEX_TYPE * tile_ptr = ctx.getSharedMemory<INDEX_TYPE>(1);
+          RAJA::View<INDEX_TYPE, RAJA::Layout<1>> Tile(tile_ptr, 1);
```
MrBurmark, line 51: should get the right amount of memory.

test/functional/launch/shared_mem/tests/test-launch-StaticMem.hpp
```diff
@@ -0,0 +1,116 @@
...
+      RAJA::loop<TEAM_POLICY>(ctx, outer_range, [&](INDEX_TYPE bid) {
+
+          //Since we are using custom index type we have to first use a 
+          //type that the device compiler can intialize, we can then use a
+          //pointer to recast the shared memory to our desired type.
+          //This enables us to work around the following warning: 
+          // warning #3019-D: dynamic initialization is not supported for
+          //a function-scope static __shared__ variable within a __device__/__global__ function
+          RAJA_TEAM_SHARED char char_Tile[THREAD_RANGE*sizeof(INDEX_TYPE)];
+          INDEX_TYPE *Tile = &((INDEX_TYPE *)char_Tile)[0];
```
MrBurmark, line 58: If you've cast it it is the pointer type you need, no need to then get a pointer to the first element.
```suggestion
          INDEX_TYPE *Tile = (INDEX_TYPE *)char_Tile;
```

test/include/RAJA_test-launch-execpol.hpp
```diff
@@ -28,7 +28,7 @@ using Sequential_launch_policies = camp::list<
 #if defined(RAJA_ENABLE_OPENMP)
 using omp_policies = camp::list<
          RAJA::LaunchPolicy<RAJA::omp_launch_t>,
-         RAJA::LoopPolicy<RAJA::omp_parallel_for_exec>  
+         RAJA::LoopPolicy<RAJA::omp_for_exec>
```
rchen20, line 31: Are we using `omp_for_exec` because the parallel region is instantiated in the top level launch?

artv3: yes, omp_launch_t opens up the parallel region.

rchen20 (approved):

artv3:
Did anyone else want to review this? Or are we ready for auto-merge?

MrBurmark (approved):

PR #1427, opened 2023-01-13
Point to new shared-ci branch to test

adrienbernede:
## Fix CI not running on develop in GitLab

Branches like develop are considered "draft" on GitHub because they don’t have an associated PR.
A fix in RADIUSS Shared CI allows to define branches that will skip the "draft test" and run anyway.

It is possible to override the list of branches always running, with the variable ALWAYS_RUN_LIST.
This will be used to validate that the fix works, without waiting to run on develop.


adrienbernede:
Tested, vetted, closed !

PR #1428, opened 2023-01-13
Generalize forall GPU indexing

MrBurmark:
# Generalize GPU indexing in forall

This is just the forall indexing changes from #1334.
This adds an indexing class in the GPU policies that allows more general behaviors for indexing.
For example `hip_exec<block_size>` is now an alias to `HipForallDirect<HipIndexGlobal<0, block_size, 0>>` to map 0 or 1 indices directly to global threads in the x direction in blocks of block_size.

- This PR is a feature
- It does the following:
  - More generic GPU indexing at my request, others have asked about using the occupancy calculator with forall.

1. First idea or goal
2. Second idea
3. Third idea

- (impractical)
- (too big)
- (not a good idea anyway)


artv3 (approved):

MrBurmark:
Contained in #1334

PR #1441, opened 2023-02-07
Attempt to share flux allocation

adrienbernede:
Trying to finally share flux allocation on corona and tioga.


adrienbernede:
@rhornung67 I think doing so is good practice: we allocate resource only once and run all the jobs in it. Basically, I would argue that it improves the throughput by only allocating once.

We do that on ruby (with slurm), but I didn’t know how to do the same with flux (tried but failed) until I used the slurm-like wrappers to figure it out.

Reminder:
Once the ressource is allocated, there are 2 strategies: "overload" the node with parallel builds, or run the builds sequentially. You always preferred running builds one after the other in RAJA, and the same happens here. In order to speed-up the build, we car increase the number on nodes in the top-level allocation.

- [ ] I marked this WIP because I want to get your approval first, then I’ll merge the associated branch in radiuss-shared-ci.

adrienbernede:
Closed in favor of #1457.

PR #1497, opened 2023-06-15
Adding a barrier synchronization after asynchronous copy for Cuda and…

clarkse:
… HIP platforms in forall-AtomicView unit test.

# Summary

- This PR is a bugfix
  - Fixes unit test failure in test-forall-AtomicView when using CUDA on RTX A6000 with slower memory bandwidth and increased memory transfer latency.  Asynchronous copying was not complete prior to subsequent kernel launch.


artv3 (approved):
Yes, I just ran into an issue related to this.

rhornung67 (approved):
I'm surprised this is needed. Thanks for the contribution.

PR #1513, opened 2023-07-07
Fix Compilation Warnings in Vectorization and Hyperplane

rchen20:
# Summary

- This PR is a bugfix
- It does the following:
  - Fixes https://github.com/LLNL/RAJA/issues/1508 and https://github.com/LLNL/RAJA/issues/1509


rhornung67 (approved):
Thank you @rchen20 I verified that warnings are now gone for clang and gcc (multiple versions of each compiler).

PR #1560, opened 2023-10-27
Use coarse grained pinned memory with Hip

MrBurmark:
# Use coarse grained pinned memory with Hip

This improves performance because it can be cached on device

- This PR is a  feature
- It does the following:
  - Adds more speed at the request of me


adayton1 (approved):
LGTM

rhornung67 (approved):

PR #1563, opened 2023-11-01
fix desul config, refer to new BLT

trws:
# Summary

- This PR is an update of blt and desul and corresponding adjustment of raja's config for desul to fix an issue reported by @rhornung67 

Depends on desul/desul#117 and LLNL/blt#664


davidbeckingsale:
It's pinned here (to the commit with John's TPL export changes) - if you change that to the appropriate hash it should work: https://github.com/LLNL/RAJA/blob/develop/.gitlab/custom-jobs-and-variables.yml#L62C1-L62C83

rhornung67:
@davidbeckingsale good catch. I forgot about that. Trying some things....

trws:
> It's pinned here (to the commit with John's TPL export changes) - if you change that to the appropriate hash it should work: https://github.com/LLNL/RAJA/blob/develop/.gitlab/custom-jobs-and-variables.yml#L62C1-L62C83

Thanks for this, I noticed the issue but had no idea where to fix it, hopefully this will take care of it.

rhornung67:
@trws I'm on it.

rhornung67:
@trws We stepped on each other. I think this is correct now. I hope your toes aren't bruised.

trws:
LoL, no problem, I had no idea there were all those other places to update, thanks for pushing the update!

rhornung67:
> LoL, no problem, I had no idea there were all those other places to update, thanks for pushing the update!

Yes. This will all be squared when we get a new BLT release. Let's see what the checks say....

.gitlab/jobs/lassen.yml
```diff
@@ -58,5 +58,5 @@ clang_13_0_1_libcpp:
 
 gcc_8_3_1_cuda_10_1_243_desul_atomics:
   variables:
-    SPEC: " ~shared +openmp +tests +cuda +desul %gcc@8.3.1 cuda_arch=70 ^cuda@10.1.243+allow-unsupported-compilers ^blt@git.a7f0a6ecc4fdfa1724399b1454c3909b9ee02e81=develop"
+    SPEC: " ~shared +openmp +tests +cuda +desul %gcc@8.3.1 cuda_arch=70 ^cuda@10.1.243+allow-unsupported-compilers ^blt@git.6dc747989504adee075bbe2ebf703a230232bea5=develop"
```
trws, line 61: BLT is updated, but it's pulling the old desul now.  
```suggestion
    SPEC: " ~shared +openmp +tests +cuda +desul %gcc@8.3.1 cuda_arch=70 ^cuda@10.1.243+allow-unsupported-compilers ^blt@git.6dc747989504adee075bbe2ebf703a230232bea5=develop ^desul@git.a0b1b2fadded49ef526e31aa296218f565a6f5d1=develop"
```

trws: No, this is not what's happening because desul isn't split into a spack package in the package version we have... how is this failing...

rhornung67: I do not know. Test output indicates CUDA kernels with atomics aren't doing anything.  In a local config on rzansel, ccmake tells me that `DESUL_ENABLE_OPENMP` is ON. That's not what we want, correct? Is something off by trying to shut that off?

I'm trying something else locally on rzansel now.....

rhornung67: @trws when I set `ENABLE_OPENMP=Off` in my CMake recipe, all tests pass. Also, all the compiler warnings coming from the CUDA atomic tests about calling a host function from a host-device function go away. 

The issue seems to be that building with both CUDA and OpenMP with desul atomics enabled breaks the CUDA atomics.

trws:
That’s exactly right, and it should be getting set to Off by raja’s cmake unless enable target OpenMP is on.  I’m not sure how it would be avoiding that.

---
Sent from Workspace ONE Boxer<https://whatisworkspaceone.com/boxer>
On November 2, 2023 at 2:20:39 PM PDT, Rich Hornung ***@***.***> wrote:

@rhornung67 commented on this pull request.

In .gitlab/jobs/lassen.yml<https://urldefense.us/v3/__https://github.com/LLNL/RAJA/pull/1563*discussion_r1380793122__;Iw!!G2kpM7uM-TzIFchu!28IIZ93zdkh8n-v-bmpycoVrV57HEc0T-CVanfPImwhB_cPrIOoc-enQFx8D3FznsORDLal17tSme6-yxY_fAiT8ZFU$>:

> @@ -58,5 +58,5 @@ clang_13_0_1_libcpp:

 gcc_8_3_1_cuda_10_1_243_desul_atomics:
   variables:
-    SPEC: " ~shared +openmp +tests +cuda +desul ***@***.*** cuda_arch=70 ***@***.***+allow-unsupported-compilers ***@***.***=develop"
+    SPEC: " ~shared +openmp +tests +cuda +desul ***@***.*** cuda_arch=70 ***@***.***+allow-unsupported-compilers ***@***.***=develop"


I do not know. Test output indicates CUDA kernels with atomics aren't doing anything. In a local config on rzansel, ccmake tells me that DESUL_ENABLE_OPENMP is ON. That's not what we want, correct? Is something off by trying to shut that off?

I'm trying something else locally on rzansel now.....

rhornung67:
@trws this seems insufficient to do what we need https://github.com/trws/desul/blob/desul-config-control/CMakeLists.txt#L26  because this `ENABLE_OPENMP=On` triggers this https://github.com/trws/desul/blob/desul-config-control/atomics/CMakeLists.txt#L35

Is there a way to turn on *host support only* for desul atomics?

trws:
Yes, `ENABLE_OPENMP=Off` leaves the compiler builtin atomics, which are the ones we want.  This: https://github.com/LLNL/RAJA/pull/1563/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR231-R233
is what should do the job though.  RAJA should be explicitly setting DESUL_ENABLE_OPENMP to off before loading the directory.  Is desul getting built some other way such that we miss that?  An earlier version did basically the same thing except it set `ENABLE_OPENMP` to off and restored it afterward, that also works but I don't like mucking with enables that we need if we can avoid it, hence adding the DESUL-specific ones.

trws:
Ok, clearly this has something to do with it, but I haven't seen this locally:

```
-- Desul Atomics support is ON
CMake Warning (dev) at /usr/tce/packages/cmake/cmake-3.23.1/share/cmake/Modules/CMakeDependentOption.cmake:75 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'DESUL_ENABLE_OPENMP'.
  tpl/desul/CMakeLists.txt:26 (cmake_dependent_option)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found OpenMP_C: -fopenmp (found version "4.5") 
```

Maybe it needs to be a switch to a cache variable, testing now to see if I can repro and fix this locally.

CMakeLists.txt
```diff
@@ -229,7 +229,7 @@ message(STATUS "Desul Atomics support is ${RAJA_ENABLE_DESUL_ATOMICS}")
 if (RAJA_ENABLE_DESUL_ATOMICS)
   # NOTE: ENABLE_OPENMP in desul implies OpenMP OFFLOAD
   if (NOT RAJA_ENABLE_TARGET_OPENMP)
-    set(DESUL_ENABLE_OPENMP Off)
+    set(DESUL_ENABLE_OPENMP Off CACHE BOOL "Disable OpenMP offload atomics to select builtin host atomics")
```
rhornung67, line 232: @trws how does this play with this? It seems odd to me because RAJA will be configured with ENABLE_OPENMP=On.

https://github.com/trws/desul/blob/desul-config-control/CMakeLists.txt#L26

trws: I had to look it up, and test it, to be honest.  The dependent option setup does this:

* if the main variable, ENABLE_OPENMP
   * exists: then the dependent variable exists as a cache variable with
      * if the dependent variable is already defined, the value it has
      * it it isn't, then the default (On) 
   * doesn't exist: then the dependent variable *does not* exist as a cache variable and is created as a local variable
     * if it was already defined, then with the value it has
     * if it isn't, then the undefined default, Off

So, if it's explicitly set before we hit that, the dependent option doesn't override it, it just changes whether it's a treated as a cache variable or not.  If it _isn't_ explicitly set then the values in there take effect.  Not _at all_ obvious.  Seems to work though.

We use the same mechanism to have ENABLE_TESTS on and disable them for say CAMP though, it's something we actually do in a bunch of places it just looks extra odd here because we aren't overriding from the command line.

rhornung67: I guess I had only a partial understanding of CMake dependent options. Definitely a delicate dance...  Thank you for your time on this.

rhornung67:
@trws it works! I restarted the poodle job that timed out. I will talk to Chris about ETA for a BLT release. I don't think he wants the BLT changes in this and prefers to make the cuda namespacing real.

trws:
Ok, I'd be happy to talk to him about it.  The funny thing about cmake is that what's in here is functionally equivalent to whatever he'd want to do, except it doesn't factor out the scary string matching stuff, so I'm not sure it makes sense to delay much if that's going to be a while.  This isn't a bandaid like the original alias in desul was, it's exposing the cuda target under two names, which I thought would be a good way to support a deprecation cycle across a BLT release.

rhornung67:
Closing. We have moved to a newer Desul and and BLT.

PR #1573, opened 2023-12-04
Pass mpi option through raja_add_test

MrBurmark:
Let the NUM_MPI_TASKS and NUM_OMP_THREADS options through the raja_add_test and raja_add_benchmark cmake macros so they can be passed to BLT.

This is intended to be used in https://github.com/LLNL/RAJAPerf/pull/332.

# Summary

- This PR is a feature,
- It does the following:
  - Adds ability to use BLT's mpi testing feature at the request of me

### DIFF
--- a/cmake/RAJAMacros.cmake
+++ b/cmake/RAJAMacros.cmake
@@ -141,7 +141,7 @@ endfunction()
 
 macro(raja_add_test)
   set(options )
-  set(singleValueArgs NAME)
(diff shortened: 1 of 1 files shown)


artv3 (approved):
Looks reasonable to me, but not sure why the CI timed out -- I don't think it is due to changes here.

rchen20 (approved):

PR #1576, opened 2023-12-07
PR from fork/1575

MrBurmark:
See #1575


artv3 (approved):
I would think our test would catch this

rchen20:
> I would think our test would catch this

We may have turned off much of the compilation of the omp target backend because it was taking so long . . .

rchen20 (approved):

PR #1585, opened 2023-12-20
[Woptim] update rocm and Spack

adrienbernede:
# Summary

This PR updates external packages use in CI.

Tests now run with rocm 5.7.1 on tioga and rocm 5.7.0 on corona, cce 16.0.1 is used with both.
MPI was updated for both machines.

Spack was updated.
The CI reservation is used on lassen.


adrienbernede:
@rhornung67
corona failing because we can’t get an alloc on it a.t.m.
lassen failing possibly because of an issue with camp@main. Who should I ping ?
Otherwise this should be ready...

adrienbernede:
@rhornung67 @kab163 I triggered Tioga jobs, they should succeed. Then this can be merged. It will incidentally "fix" time out issues on poodle.

rhornung67 (approved):

PR #1592, opened 2024-01-24
removing module map stuff

kab163:
# Summary

- This PR is a bugfix, I think we'd like to include this in next release?
- It does the following:
  - Fixes modulemap stuff described in Issue #1150


davidbeckingsale:
I think you should also delete the files themselves.

kab163:
@rhornung67 - We talked about this issue yesterday at the RAJA meeting. I totally forgot about it! I removed what I thought I needed to and couldn't find any other mentions of the modulemap stuff - is this pretty much what is needed for fix this issue or is there something else to add?

kab163:
> I think you should also delete the files themselves.

Haha yes that makes sense :p thanks

rhornung67 (approved):
Thank you @kab163 I also forgot about this and the issue was assigned to me 👎

PR #1596, opened 2024-02-07
Remove qlanglvl=extended0x flags from xl build command

adrienbernede:
Test a fix in XL build.


adrienbernede:
xl builds... old times.

PR #1616, opened 2024-03-19
Draft: Desul ordered atomic policies + litmus tests

publixsubfan:
# Summary

- When the Desul atomic backend is enabled, adds atomic policies in the form of `RAJA::atomic_{mem_policy}_{scope}`, where:
  - `mem_policy` is one of `relaxed`, `acquire`, `release`, `acq_rel`, or `seq_cst`
  - `scope` is either empty (device-scope), `system` for a system-wide atomic, or `block` for a block-wide atomic
- Adds 2-thread litmus tests to test the ability of acquire-release and SC atomics to restore sequential memory behavior on relaxed memory platforms

## Motivation

On architectures which adopt relaxed memory models (ARM, PowerPC, most GPU architectures), the order in which memory modifications on one thread are observed on another thread may differ from the "program order" of the memory operations. This may lead to unexpected results if, for example, an atomic variable is used as a mutex; writes done in the critical section may not be visible to another thread due to the memory subsystem reordering the writes.

x86 implements a much stronger, but not entirely sequentially-consistent memory model (x86-TSO). The only observable reordering between threads is Store->Load reordering, where an earlier (in "program-order") store can be reordered after the load. The "Store Buffer" litmus test demonstrates this behavior, where without fencing, it can appear as if the store instruction in both threads happen after their corresponding load instructions.

Desul supports specifying a memory order policy, which can restore a consistent view of memory operations between threads with a stronger memory ordering pair between the two threads.

## Litmus testing

The added GPU litmus tests are based off of the work here: https://gpuharbor.ucsc.edu/webgpu-mem-testing/ and in a paper ["Foundations of Empirical Memory Consistency Testing"](https://users.soe.ucsc.edu/~tsorensen/files/oopsla2020.pdf), Kirkham et al. (OOPSLA 2020).

Litmus testing allows us to probe the existence of relaxed memory behavior on GPU platforms. We implement a family of 2-thread tests, where each thread attempts to write data to or read data from a thread residing on a different block.

## More references

["A Tutorial Introduction to the ARM and POWER Relaxed Memory Models"](https://www.cl.cam.ac.uk/~pes20/ppc-supplemental/test7.pdf)


trws:
One comment based on yours @publixsubfan, all previously existing raja atomics were relaxed.  It's the stronger ones, and really the scopes, that are most interesting because they can mean we can pass data without having to do all loads and stores with atomics.  Block scope atomics, with device scope fences only when necessary, also are likely to help us greatly on El Cap.  The atomics themselves wont be faster, but there will be substantially less expensive cache invalidation.

The code looks good to me with a cursory look over it. I'm not sure what we want to do with respect to these interfaces longer term, but this looks good to me as a place to explore.

MrBurmark:
Can you add something about how you get back non-atomic memory operations using this interface?

include/RAJA/policy/desul/atomic.hpp
```diff
@@ -26,153 +25,208 @@ using raja_default_desul_scope = desul::MemoryScopeDevice;
...
 RAJA_SUPPRESS_HD_WARN
 template <typename AtomicPolicy, typename T>
-RAJA_HOST_DEVICE
-RAJA_INLINE T
-atomicAdd(AtomicPolicy, T volatile *acc, T value) {
-  return desul::atomic_fetch_add(const_cast<T*>(acc),
+RAJA_HOST_DEVICE RAJA_INLINE T atomicAdd(AtomicPolicy, T volatile *acc, T value)
+{
+  using desul_order =
+      typename detail::DesulAtomicPolicy<AtomicPolicy>::memory_order;
```
MrBurmark, line 49: Does this mean that AtomicPolicy has to be `detail_atomic_t<...>` instead of `DesulAtomicPolicy<...>`

trws:
> Can you add something about how you get back non-atomic memory operations using this interface?

When designing desul, we realized that the sequential option behaved more like a scope than a memory order.  The scope is `MemoryScopeCaller` in the desul interface [here](https://github.com/desul/desul/blob/47cb1acc746c0ac0d83f9a90331fbafc3f6e7b82/atomics/include/desul/atomics/Common.hpp#L55).  The main reason for this is that we only need one implementation for essentially all backends to support it, and it makes no sense to give a different scope when there's no coherence.  Of course, there's no point in giving a memory order either, but if the scope is the caller they at least all _make sense_.

PR #1643, opened 2024-05-13
Use GitLab container registry as a Spack build cache

adrienbernede:
In this PR, we set the GitLab CI to push Spack built binaries (RAJA dependencies) to the builtin OCI container registry.

This required an update of Spack, RADIUSS Spack Configs and Uberenv.


rhornung67 (approved):
Thank you @adrienbernede This looks very useful.

PR #1663, opened 2024-06-04
Add atomicLoad and atomicStore

adayton1:
# Summary

- This PR is a feature
- It does the following:
  - Adds atomicLoad and atomicStore at the request of myself
  - Adds *.swp files to gitignore list

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 /install_*/
 /install-*/
 /Debug/
+*.swp
--- a/include/RAJA/pattern/atomic.hpp
+++ b/include/RAJA/pattern/atomic.hpp
@@ -81,6 +81,19 @@ namespace RAJA
 
 
 /*!
+ * @brief Atomic load
+ * @param acc Pointer to location of value
+ * @return Returns value at acc
+ */
+RAJA_SUPPRESS_HD_WARN
+template <typename Policy, typename T>
+RAJA_INLINE RAJA_HOST_DEVICE T atomicLoad(T volatile *acc)
+{
+  return RAJA::atomicAdd(Policy{}, acc);
+}
+
+
+/*!
  * @brief Atomic add
  * @param acc Pointer to location of result value
  * @param value Value to add to *acc
--- a/include/RAJA/pattern/atomic.hpp
+++ b/include/RAJA/pattern/atomic.hpp
@@ -97,7 +97,6 @@ RAJA_INLINE RAJA_HOST_DEVICE T atomicLoad(T volatile *acc)
  * @brief Atomic store
  * @param acc Pointer to location of value
  * @param value Value to store at *acc
- * @return Stores value at acc
  */
 RAJA_SUPPRESS_HD_WARN
 template <typename Policy, typename T>
--- a/include/RAJA/pattern/atomic.hpp
+++ b/include/RAJA/pattern/atomic.hpp
@@ -89,7 +89,7 @@ RAJA_SUPPRESS_HD_WARN
 template <typename Policy, typename T>
 RAJA_INLINE RAJA_HOST_DEVICE T atomicLoad(T volatile *acc)
 {
-  return RAJA::atomicAdd(Policy{}, acc);
+  return RAJA::atomicLoad(Policy{}, acc);
 }
 
 
--- a/include/RAJA/pattern/atomic.hpp
+++ b/include/RAJA/pattern/atomic.hpp
@@ -94,6 +94,20 @@ RAJA_INLINE RAJA_HOST_DEVICE T atomicLoad(T volatile *acc)
 
 
 /*!
+ * @brief Atomic store
+ * @param acc Pointer to location of value
+ * @param value Value to store at *acc
+ * @return Stores value at acc
+ */
+RAJA_SUPPRESS_HD_WARN
+template <typename Policy, typename T>
+RAJA_INLINE RAJA_HOST_DEVICE void atomicStore(T volatile *acc, T value)
+{
+  return RAJA::atomicStore(Policy{}, acc, value);
+}
+
+
+/*!
  * @brief Atomic add
  * @param acc Pointer to location of result value
  * @param value Value to add to *acc
--- a/include/RAJA/pattern/atomic.hpp
+++ b/include/RAJA/pattern/atomic.hpp
@@ -83,7 +83,7 @@ namespace RAJA
 /*!
  * @brief Atomic load
  * @param acc Pointer to location of value
- * @return Returns value at acc
+ * @return Value at acc
  */
 RAJA_SUPPRESS_HD_WARN
 template <typename Policy, typename T>
--- a/include/RAJA/pattern/atomic.hpp
+++ b/include/RAJA/pattern/atomic.hpp
@@ -103,7 +103,7 @@ RAJA_SUPPRESS_HD_WARN
 template <typename Policy, typename T>
 RAJA_INLINE RAJA_HOST_DEVICE void atomicStore(T volatile *acc, T value)
 {
-  return RAJA::atomicStore(Policy{}, acc, value);
+  RAJA::atomicStore(Policy{}, acc, value);
 }
 
 
--- a/include/RAJA/policy/desul/atomic.hpp
+++ b/include/RAJA/policy/desul/atomic.hpp
@@ -41,10 +41,10 @@ template <typename AtomicPolicy, typename T>
 RAJA_HOST_DEVICE
 RAJA_INLINE void
 atomicStore(AtomicPolicy, T volatile *acc, T value) {
-  return desul::atomic_store(const_cast<T*>(acc),
-                             value,
-                             raja_default_desul_order{},
-                             raja_default_desul_scope{});
+  desul::atomic_store(const_cast<T*>(acc),
+                      value,
+                      raja_default_desul_order{},
+                      raja_default_desul_scope{});
 }
 
 RAJA_SUPPRESS_HD_WARN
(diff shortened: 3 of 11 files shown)


MrBurmark:
How about atomicStore too?

adayton1:
> How about atomicStore too?

I was going to ask about that. It looks like we already have atomicExchange - should atomicStore just be an alias to atomicExchange? Or should atomicStore not return the old value? We have the old value, so I see no reason not to return it, though.

MrBurmark:
Ya the non-returning part of atomicStore is what makes it different. Some architectures have an atomicStore that is a bit faster because it doesn't return anything.

adayton1:
> Ya the non-returning part of atomicStore is what makes it different. Some architectures have an atomicStore that is a bit faster because it doesn't return anything.

I've added atomicStore now.

test/functional/forall/atomic-basic/tests/test-forall-atomic-basic.hpp
```diff
@@ -78,8 +78,6 @@ void ForallAtomicBasicTestImpl( IdxType seglimit )
                               &check_array,
                               &test_array );
 
-  work_res.memcpy( work_array, test_array, sizeof(T) * len );
```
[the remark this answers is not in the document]
MrBurmark: You can remove them.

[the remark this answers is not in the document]
MrBurmark: in general we should be doing work_res.wait() instead of a backend specific sync anyway.

include/RAJA/policy/hip/atomic.hpp
```diff
@@ -474,6 +474,32 @@ RAJA_INLINE __device__ T hip_atomicCAS( T volatile *acc, T compare, T val)
...
+RAJA_SUPPRESS_HD_WARN
+template <typename T, typename host_policy>
+RAJA_INLINE RAJA_HOST_DEVICE void
+atomicStore(hip_atomic_explicit<host_policy>, T volatile *acc, T value)
+{
+#if defined(__HIP_DEVICE_COMPILE__)
+  detail::hip_atomicExchange(acc, value);
+  return;
+#else
+  return RAJA::atomicStore(host_policy{}, acc, value);
```
MrBurmark, line 499: Do we really need to write returns here?

include/RAJA/policy/hip/atomic.hpp
```diff
@@ -474,6 +474,32 @@ RAJA_INLINE __device__ T hip_atomicCAS( T volatile *acc, T compare, T val)
...
  * These are atomic in hip device code and non-atomic otherwise
  */
+
+RAJA_SUPPRESS_HD_WARN
+template <typename T, typename host_policy>
+RAJA_INLINE RAJA_HOST_DEVICE T
+atomicLoad(hip_atomic_explicit<host_policy>, T volatile *acc)
+{
+#if defined(__HIP_DEVICE_COMPILE__)
+  return detail::hip_atomicAdd(acc, (T)0);
```
MrBurmark, line 484: I'm pretty sure there is a builtin for hip atomic load and store, https://github.com/ROCm/rocPRIM/blob/4458244e6dc78e666997d7552052798d069d18aa/rocprim/include/rocprim/intrinsics/atomic.hpp#L80.

include/RAJA/policy/hip/atomic.hpp
adayton1: Done.

include/RAJA/policy/hip/atomic.hpp
MrBurmark: I tried these out and they were a bit faster on some platforms, but they also aren't exactly documented so I don't know what level of support they have long term.

adayton1: I can't find any documentation for it, or when it was added. Though it looks like it wasn't used in rocprim until rocm 6. Do you have an example of conditionally doing something for a specific rocm version? Would I just need to check for a specific value of `__HIP_DEVICE_COMPILE__`?

MrBurmark: That would be pretty good. Maybe a check on `__has_builtin` would be useful as hip always uses clang https://releases.llvm.org/3.1/tools/clang/docs/LanguageExtensions.html.

MrBurmark: For checking the hip version, https://github.com/LLNL/RAJA/blob/develop/include/RAJA/config.hpp.in#L274.

adayton1: I'm now using the hip builtin atomics if they are available.

include/RAJA/policy/hip/atomic.hpp
```diff
@@ -260,6 +273,35 @@ using hip_atomicExch_builtin_types = list<
...
+  return hip_atomic_CAS_oper(acc, [=] __device__(T a) {
+    return a;
+  });
+}
+
+template <typename T, enable_if_is_any_of<T, hip_atomicLoad_builtin_types>* = nullptr>
+RAJA_INLINE __device__ T hip_atomicLoad(T volatile *acc)
+{
+  return __hip_atomic_load((T *)acc, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT);
+}
```
MrBurmark, line 288: I think this is too constrained, we should be able to load a double via unsigned long long.

adayton1: I've checked and it compiles with all builtin arithmetic types (integral and floating point), as well as enums. So I've changed this to be enabled if std::is_arithmetic<T>::value || std::is_enum<T>::value, and then let all other types fall back to the compare and swap.

MrBurmark (approved):

.gitignore
```diff
@@ -14,3 +14,4 @@
 /install_*/
 /install-*/
 /Debug/
+*.swp
```
rhornung67, line 17: 😄

include/RAJA/pattern/atomic.hpp
```diff
@@ -80,6 +80,33 @@ namespace RAJA
  */
 
 
+/*!
+ * @brief Atomic load
+ * @param acc Pointer to location of value
+ * @return Returns value at acc
```
rhornung67, line 86:
```suggestion
 * @return Value at acc
```

include/RAJA/pattern/atomic.hpp
```diff
@@ -80,6 +80,33 @@ namespace RAJA
...
+{
+  return RAJA::atomicLoad(Policy{}, acc);
+}
+
+
+/*!
+ * @brief Atomic store
+ * @param acc Pointer to location of value
+ * @param value Value to store at *acc
+ * @return Stores value at acc
```
rhornung67, line 100:
```suggestion

```
No return statement.

include/RAJA/policy/desul/atomic.hpp
```diff
@@ -26,6 +26,27 @@ using raja_default_desul_scope = desul::MemoryScopeDevice;
 namespace RAJA
 {
 
+RAJA_SUPPRESS_HD_WARN
+template <typename AtomicPolicy, typename T>
+RAJA_HOST_DEVICE
+RAJA_INLINE T
+atomicLoad(AtomicPolicy, T volatile *acc) {
```
rhornung67, line 33:
```suggestion
atomicLoad(AtomicPolicy, T volatile *acc)
```

include/RAJA/policy/desul/atomic.hpp
```diff
@@ -26,6 +26,27 @@ using raja_default_desul_scope = desul::MemoryScopeDevice;
...
+  return desul::atomic_load(const_cast<T*>(acc),
```
rhornung67, line 34:
```suggestion
  {
    return desul::atomic_load(const_cast<T*>(acc),
```

include/RAJA/policy/desul/atomic.hpp
rhornung67, line 33:
```suggestion
atomicLoad(AtomicPolicy, T volatile *acc)
{
```

Format consistency with other routine.

include/RAJA/policy/desul/atomic.hpp
```diff
@@ -26,6 +26,27 @@ using raja_default_desul_scope = desul::MemoryScopeDevice;
...
+                            raja_default_desul_order{},
+                            raja_default_desul_scope{});
+}
+
+RAJA_SUPPRESS_HD_WARN
+template <typename AtomicPolicy, typename T>
+RAJA_HOST_DEVICE
+RAJA_INLINE void
+atomicStore(AtomicPolicy, T volatile *acc, T value) {
```
rhornung67, line 43:
```suggestion
atomicStore(AtomicPolicy, T volatile *acc, T value)
{
```

Format consistency with other routines.

rhornung67 (commented):
A few minor nits to address.

include/RAJA/policy/desul/atomic.hpp
adayton1: Fixed.

include/RAJA/policy/desul/atomic.hpp
adayton1: Fixed.

rhornung67 (approved):

adayton1:
It's taking forever to get an allocation on poodle. Is that normal?

rhornung67:
> It's taking forever to get an allocation on poodle. Is that normal?

It could be bogged down with folks running jobs. Poodle is usually pretty quick.

rhornung67:
@adayton1 I restarted the tioga CI jobs when the machine came back up. All checks are green and this is approved. Please feel free to merge.

PR #1678, opened 2024-06-21
(WIP) Atomic clean up

adayton1:
# Summary

- This PR is a refactoring:
- It does the following:
  - Refactors hip and cuda atomics to have a cleaner implementation

### DIFF
--- a/include/RAJA/policy/hip/atomic.hpp
+++ b/include/RAJA/policy/hip/atomic.hpp
@@ -97,126 +97,107 @@ RAJA_INLINE __device__ T hip_atomicExchange(T *acc, T value)
 }
 
 
+#if defined(__has_builtin) && \
+    (__has_builtin(__hip_atomic_load) || __has_builtin(__hip_atomic_store))
+
+/*!
+ * Type trait for determining if the operator should be implemented
+ * using an intrinsic
+ */
+template <typename T>
+struct builtin_useIntrinsic {
+  static constexpr bool value =
+    (std::is_integral<T>::value || std::is_enum<T>::value) &&
+    (sizeof(T) == 1 || sizeof(T) == 2 || sizeof(T) == 4 || sizeof(T) == 8);
+};
+
+
+/*!
+ * Type trait for determining if the operator should be implemented
+ * by reinterpreting inputs to types that intrinsics support
+ */
+template <typename T>
+struct builtin_useReinterpret {
+  static constexpr bool value =
+    !std::is_integral<T>::value &&
+    !std::is_enum<T>::value &&
+    ((sizeof(T) == 1
+#if !defined(UINT8_MAX)
+      && sizeof(unsigned char) == 1
+#endif
+     ) ||
+     (sizeof(T) == 2
+#if !defined(UINT16_MAX)
+      && sizeof(unsigned short) == 2
+#endif
+     ) ||
+     (sizeof(T) == 4
+#if !defined(UINT32_MAX)
+      && sizeof(unsigned int) == 4
+#endif
+     ) ||
+     (sizeof(T) == 8
+#if !defined(UINT64_MAX)
+      && sizeof(unsigned long long) == 8
+#endif
+     ));
+
+  using type =
+    std::conditional_t<sizeof(T) == 1,
+#if defined(UINT8_MAX)
+                       uint8_t,
+#else
+                       unsigned char,
+#endif
+    std::conditional_t<sizeof(T) == 2,
+#if defined(UINT16_MAX)
+                       uint16_t,
+#else
+                       unsigned short,
+#endif
+    std::conditional_t<sizeof(T) == 4,
+#if defined(UINT32_MAX)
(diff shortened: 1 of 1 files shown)


include/RAJA/policy/hip/atomic.hpp
```diff
@@ -55,200 +55,244 @@ using hip_atomicCommon_builtin_types = ::camp::list<
...
+using hip_useReinterpretLoadStore_t = typename hip_useReinterpretLoadStore<T>::type;
 
 #endif
 
-#if defined(UINT64_MAX)
+
+/*!
+ * Atomic load
+ */
+#if defined(__has_builtin) && __has_builtin(__hip_atomic_load)
```
MrBurmark, line 260: This conditional is a bit different than above where store is also checked.

include/RAJA/policy/hip/atomic.hpp
```diff
@@ -55,200 +55,244 @@ using hip_atomicCommon_builtin_types = ::camp::list<
...
-  return ::atomicOr(acc, static_cast<T>(0));
+  return hip_atomicOr(acc, static_cast<T>(0));
 }
 
 template <typename T,
-          std::enable_if_t<!std::is_same<T, int>::value &&
-                           !std::is_same<T, unsigned int>::value &&
-                           !std::is_same<T, unsigned long long int>::value &&
-                           sizeof(T) == sizeof(unsigned int), bool> = true>
+          std::enable_if_t<hip_useReinterpretCommon<T>::value, bool> = true>
```
MrBurmark, line 289: I wonder if it makes sense combining the reinterpret cases here since the implementation is the same and doing the ifdef just around the enable_if in the template.

include/RAJA/policy/hip/atomic.hpp
adayton1: That's intentional - this section is just for atomic load. There is a section further down for atomic store. But the type traits are the same for both atomic load and store, so if only one intrinsic is provided, we still want the type traits defined.

MrBurmark: Whoops, I read the one above as an && instead of an ||.

include/RAJA/policy/hip/atomic.hpp
adayton1: Done.

test/include/RAJA_test-atomic-types.hpp
```diff
@@ -24,19 +21,10 @@
 using AtomicDataTypeList =
   camp::list< RAJA::Index_type,
               int,
-#if defined(RAJA_TEST_EXHAUSTIVE)
-              unsigned,
+              unsigned int,
```
MrBurmark, line 24: Do we still want to separate out some for the exhaustive tests?

include/RAJA/policy/cuda/atomic.hpp
```diff
@@ -556,17 +623,15 @@ template <typename T,
...
-template <typename T,
-          RAJA::util::enable_if_is_any_of<T, cuda_atomicBit_builtin_types>* = nullptr>
-RAJA_INLINE __device__ T cuda_atomicOr(T *acc, T value)
-{
-  return ::atomicOr(acc, value);
-}
+/*!
+ * Atomic or via builtin functions was implemented much earlier since atomicLoad
+ * may depend on it.
+ */
```
MrBurmark, line 634: You could forward declare it instead of moving it up, not sure if that will look nicer or not.

test/include/RAJA_test-atomic-types.hpp
adayton1: I'm torn. But thinking about it, int and double cover the majority of our use cases.

include/RAJA/policy/cuda/atomic.hpp
adayton1: I think I'd rather just keep a comment here.

test/include/RAJA_test-atomic-types.hpp
adayton1: OK, I'll leave int and double testing on for now and let the rest be enabled during exhaustive testing.

MrBurmark (approved):

rhornung67 (approved):

include/RAJA/policy/cuda/atomic.hpp
```diff
@@ -164,104 +232,98 @@ RAJA_INLINE __device__ void cuda_atomicStore(T *acc, T value)
...
+     sizeof(T) == sizeof(unsigned short) ||
 #endif
+     sizeof(T) == sizeof(unsigned int) ||
+     sizeof(T) == sizeof(unsigned long long)
+    );
 
-template <typename T,
-          std::enable_if_t<
+  using type =
+#if __CUDA_ARCH__ >= 700
```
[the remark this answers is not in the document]
adayton1: It took me a bit to figure it out haha

PR #1689, opened 2024-07-01
Use c version of memcpy in device code

MrBurmark:
Only the c version of memcpy is supported
in device code.
I'm not sure why this compiles in our tests.

# Summary

- This PR is a bugfix
- It does the following:
  - Fixes hip compile


rhornung67 (approved):
Maybe nvcc does some namespace shenanigans for things like this??

PR #1690, opened 2024-07-12
Draft: Valloc Interface for New Reducer

rchen20:
# Summary

- This PR is a refactoring
- It does the following:
  - Refactors new reduction interface to have consistent operator semantics
  - Introduces type safety to reduction operators
  - Will require the use of additional types for reduction targets

- [x] sequential
- [ ] CUDA
- [ ] HIP
- [ ] SYCL
- [ ] OpenMP


examples/forall-param-reductions.cpp
```diff
@@ -146,6 +146,62 @@ int main(int RAJA_UNUSED_ARG(argc), char** RAJA_UNUSED_ARG(argv[]))
...
 
 
+//
+// Define RefLoc Type
+//
+
+  using REF_INT_SUM = RAJA::expt::ValOp<int, RAJA::operators::plus>;
+  using REF_INT_MIN = RAJA::expt::ValOp<int, RAJA::operators::minimum>;
+  using REF_INT_MAX = RAJA::expt::ValOp<int, RAJA::operators::maximum>;
+  using REFLOC_INT_MIN = RAJA::expt::ValLocOp<int, RAJA::operators::minimum>;
```
MrBurmark, line 156: We should have an option to change the loc type.

include/RAJA/pattern/params/reducer.hpp
```diff
@@ -23,29 +25,115 @@ struct ValLoc {
...
+  RAJA_HOST_DEVICE constexpr ValLoc(value_type v, RAJA::Index_type l) : val(v), loc(l) {}
 
-  bool constexpr operator<(const ValLoc& rhs) const { return val < rhs.val; }
-  bool constexpr operator>(const ValLoc& rhs) const { return val > rhs.val; }
+  RAJA_HOST_DEVICE constexpr bool operator<(const ValLoc& rhs) const { return val < rhs.val; }
+  RAJA_HOST_DEVICE constexpr bool operator>(const ValLoc& rhs) const { return val > rhs.val; }
 
-  value_type getVal() {return val;}
-  RAJA::Index_type getLoc() {return loc;}
+  RAJA_HOST_DEVICE constexpr value_type getVal() const {return val;}
```
MrBurmark, line 35: If this is constexpr everything here should probably be constexpr.

include/RAJA/pattern/params/reducer.hpp
```diff
@@ -23,29 +25,115 @@ struct ValLoc {
...
+  RAJA_HOST_DEVICE constexpr ValOp & operator&=(const value_type& rhs) { val &= rhs; return *this; }
+
+  template <typename U = op_type, std::enable_if_t<std::is_same<U, RAJA::operators::bit_or<T,T,T>>::value> * = nullptr>
+  RAJA_HOST_DEVICE constexpr ValOp & operator|=(const value_type& rhs) { val |= rhs; return *this; }
+
+  template <typename U = op_type, std::enable_if_t<std::is_same<U, RAJA::operators::bit_and<T,T,T>>::value> * = nullptr>
+  RAJA_HOST_DEVICE ValOp & operator&=(value_type& rhs) { val &= rhs; return *this; }
+
+  template <typename U = op_type, std::enable_if_t<std::is_same<U, RAJA::operators::bit_or<T,T,T>>::value> * = nullptr>
+  RAJA_HOST_DEVICE ValOp & operator|=(value_type& rhs) { val |= rhs; return *this; }
```
MrBurmark, line 69: duplicates?

include/RAJA/pattern/params/reducer.hpp
```diff
@@ -85,14 +173,19 @@ namespace detail
...
     RAJA_HOST_DEVICE Reducer() {}
     Reducer(value_type *target_in) : target(target_in), val(op::identity()) {}
+    //Reducer(value_type *target_in) : target(target_in), val(op::identity()), vop(value_op(*target_in)) {}
+    //Reducer(value_op *target_in) : target(target_in->get()), val(op::identity()), vop(value_op(*target_in)) {}
 
     value_type *target = nullptr;
     value_type val = op::identity();
+    //value_op *vop = nullptr;
 
 #if defined(RAJA_CUDA_ACTIVE) || defined(RAJA_HIP_ACTIVE) || defined(RAJA_SYCL_ACTIVE)
```
MrBurmark, line 190: This reminds me that we should specialize SoAPtr on ValLoc, etc.

include/RAJA/policy/sequential/params/reduce.hpp
```diff
@@ -7,6 +7,50 @@ namespace RAJA {
 namespace expt {
 namespace detail {
 
+  // Init
+  template<typename EXEC_POL, template <typename, typename, typename> class OP, typename T>
+  camp::concepts::enable_if< std::is_same< EXEC_POL, RAJA::seq_exec> >
+  init(Reducer<OP<ValOp<ValLoc<T>,OP>, ValOp<ValLoc<T>,OP>, ValOp<ValLoc<T>,OP>>, ValOp<ValLoc<T>,OP>>& red) {
+    red.val = OP<ValOp<ValLoc<T>,OP>, ValOp<ValLoc<T>,OP>, ValOp<ValLoc<T>,OP>>::identity();
```
MrBurmark, line 14: We should figure out why using `Op<ValOp<ValLoc<T>,OP>, ...>::identity()` works here but `Op<ValOp<T,OP>, ...>::identity()` doesn't work below.

include/RAJA/pattern/params/reducer.hpp
rchen20: Thanks, good catch.

include/RAJA/pattern/params/reducer.hpp
rchen20: Everything here should already be `constexpr`; am I missing something in particular?

examples/forall-param-reductions.cpp
rchen20: Do you mean you want something like `ValLocOp<double, int, minimum>`, where `int` represents the loc type? I don't think we've had that in the past.

include/RAJA/pattern/params/reducer.hpp
MrBurmark: no, looks good now.

examples/forall-param-reductions.cpp
MrBurmark: yep

rchen20:
After some discussion with @MrBurmark, we've both concluded that the GPU backends for a valloc interface have enough complications that will require a re-design of valloc. The problem lies in the fact that the GPU backend implementation of `grid_reduce` (e.g. policy/cuda/reduce.hpp) relies on the reduction object and the val-x type (T). In the current valloc design, the val-x type can become a complex type (ValLoc, ValOp, ValOp<ValLoc>), which was not the intended use for `grid_reduce` which primarily should handle basic data types. To remedy this, I will change the valloc interface requirements, and attempt to hide the various val-x types within RAJA, rather than exposing them to the user. Here is an example of the new valloc we have in mind:

Original Valloc Design -
```
using REF_INT_SUM = RAJA::expt::ValOp<int, RAJA::operators::plus>;
using REFLOC_INT_MIN = RAJA::expt::ValLocOp<int, RAJA::operators::minimum>;
REF_INT_SUM rcuda_sum(0);
REFLOC_INT_MIN rcuda_minloc(std::numeric_limits<int>::max(), -1);
RAJA::forall<EXEC_POL3R>(cuda_res, arange,
  RAJA::expt::Reduce(&rcuda_sum),
  RAJA::expt::Reduce(&rcuda_minloc),
  [=] RAJA_DEVICE (int i, REF_INT_SUM &_rcuda_sum, REFLOC_INT_MIN &_rcuda_minloc){ // lambda body } );
```

New Valloc Design -
```
int rcuda_sum(0);
RAJA::expt::ValLoc<int> rcuda_minloc(std::numeric_limits<int>::max(), -1);
RAJA::forall<EXEC_POL3R>(cuda_res, arange,
  RAJA::expt::Reduce<RAJA::operators::plus>(&rcuda_sum),
  RAJA::expt::Reduce<RAJA::operators::minimum>(&rcuda_minloc),
  [=] RAJA_DEVICE (int i, RAJA::expt::reduceref &_rcuda_sum, RAJA::expt::reduceref &_rcuda_minloc){ // lambda body } );
```

Notes on new design -
It looks like the first new reducer design, except that the lambda data types are references to the underlying data. The lambda data type `RAJA::expt::reduceref` will internally generate the various val-x types we need, depending on the Reduce policies and concrete data types passed in. `reduceref` will ensure the type safety and interface consistency that we wanted in the beginning.

artv3:
@rchen20 @MrBurmark Will we update all the other reducers as well?

rchen20:
> @rchen20 @MrBurmark Will we update all the other reducers as well?

No, I'm fairly certain this interface change can be contained to the new reductions. Although, later on when we unify the old and new reducers, we can consider how that interface will look, and how we want to implement it. For this implementation, I will probably use some of the principles from @MrBurmark's multi-reducer design.

artv3:
> > @rchen20 @MrBurmark Will we update all the other reducers as well?
> 
> No, I'm fairly certain this interface change can be contained to the new reductions. Although, later on when we unify the old and new reducers, we can consider how that interface will look, and how we want to implement it. For this implementation, I will probably use some of the principles from @MrBurmark's multi-reducer design.

Right that is what I meant, like the SumReducer etc...

rchen20:
Closing in favor of the more updated version here https://github.com/LLNL/RAJA/pull/1732.

PR #1699, opened 2024-07-24
Missing GPU Info Headers in KernelName

rchen20:
# Summary

- This PR is a bugfix
- It does the following:
  - Adds missing GPU info headers.

PR #1702, opened 2024-07-24
Add kernel names in forall tests

artv3:
This PR will add further testing to the kernel naming API found in RAJA::forall.


MrBurmark:
Is the kernel name exposed to plugins?

artv3:
> Is the kernel name exposed to plugins?

I don't think so.

MrBurmark:
> > Is the kernel name exposed to plugins?
> 
> I don't think so.

Something we could think about during our next meeting.

rchen20 (approved):

artv3:
Did others want to review? or is this good to go when green?

adayton1 (approved):
LGTM

MrBurmark:
This is just testing the mechanics, right? There are no things that actually use/test the fact that the loops are named?

rchen20:
> This is just testing the mechanics, right? There are no things that actually use/test the fact that the loops are named?

Yes, we just want to make sure it compiles correctly, and avoid problems like in #1699. Previously, the kernel names were only used in examples/forall-param-reductions.cpp, and this would make sure there are some automated builds of kernelname.

artv3:
> > This is just testing the mechanics, right? There are no things that actually use/test the fact that the loops are named?
> 
> Yes, we just want to make sure it compiles correctly, and avoid problems like in #1699. Previously, the kernel names were only used in examples/forall-param-reductions.cpp, and this would make sure there are some automated builds of kernelname.

Right, at the moment we only support ROCTX/NVTX -- an area of future investigation is how to better integrate with profiling tools that allow kernel naming.

MrBurmark (approved):

rhornung67 (approved):

Issue #1718, opened 2024-08-22
Slowdown observed with Raja View


rahulb1218:
**Describe the bug**

Slowdown observed with Raja View when compared to accessing memory directly

**To Reproduce**

```
#include <RAJA/RAJA.hpp>
#include <iostream>
#include <cuda_runtime.h>
#include <caliper/cali.h>

int main() {
    CALI_CXX_MARK_FUNCTION;
    const int N = 10000;  
    const int K = 17; 

    int* array = new int[N * N];
    int* array_copy = new int[N * N];

    //big array, or image
    for (int i = 0; i < N * N; ++i) {
        array[i] = 1;
        array_copy[i] = 1;
    }

    //small array that acts as the blur
    int* kernel = new int[K * K];
    for (int i = 0; i < K * K; ++i) {
        kernel[i] = 2;
    }

    // copying to gpu
    int* d_array;
    int* d_array_copy;
    int* d_kernel;
    cudaMalloc((void**)&d_array, N * N * sizeof(int));
    cudaMalloc((void**)&d_array_copy, N * N * sizeof(int));
    cudaMalloc((void**)&d_kernel, K * K * sizeof(int));
    cudaMemcpy(d_array, array, N * N * sizeof(int), cudaMemcpyHostToDevice);
    cudaMemcpy(d_array_copy, array_copy, N * N * sizeof(int), cudaMemcpyHostToDevice);
    cudaMemcpy(d_kernel, kernel, K * K * sizeof(int), cudaMemcpyHostToDevice);


    constexpr int DIM = 2;
    RAJA::View<int, RAJA::Layout<DIM, int, 1>> array_view(d_array, N, N);
    RAJA::View<int, RAJA::Layout<DIM, int, 1>> array_view_copy(d_array_copy, N, N);
    RAJA::View<int, RAJA::Layout<DIM, int, 1>> kernel_view(d_kernel, K, K);


    using EXEC_POL5 = RAJA::KernelPolicy<
        RAJA::statement::CudaKernelFixed<256,
            RAJA::statement::For<1, RAJA::cuda_global_size_y_direct<16>,
                RAJA::statement::For<0, RAJA::cuda_global_size_x_direct<16>,
                    RAJA::statement::Lambda<0>
                >
            >
        >
    >;

    RAJA::RangeSegment range_i(0, N);
    RAJA::RangeSegment range_j(0, N);


    CALI_MARK_BEGIN("RAJA_View_Kernel");
[...]

PR #1728, opened 2024-09-02
Add RAJA view performance test to benchmark

artv3:
After resolving  issue #1718, this PR now adds the performance test into the bench mark folder.

//------------

This PR adds the code provided in issue #1718 in an effort to reproduce the slow down. 
I don't  have access to pascal but on lassen I see comparable performance: 

Elapsed time with RAJA view : 0.0951086
Elapsed time with NO RAJA view : 0.0952884 

To avoid measuring stream initialization I added an basic forall  at the start of the program. 

Compiler setup: 
nvcc: nvcc11.2.0, cuda_arch=70, gcc8.3.1

---- 
I also tried: nvcc11.8.0, cuda_arch=70, gcc8.3.1
Elapsed time with RAJA view : 0.0949394
Elapsed time with NO RAJA view : 0.0949237


artv3:
@artv3 move to benchmark folder.

rahulb1218:
I ran the code on Pascal and got the following results:
```
Elapsed time with RAJA view : 16.6409
Elapsed time with NO RAJA view : 2.26529
```

MrBurmark:
What gpus are on pascal? It seems strange that they are ~24x slower than the V100s on lassen?

rahulb1218:
Tesla P100-PCIE-16GB I believe.

rahulb1218:
We found that the issue was that my code built for debugging which was causing the slowdown.

benchmark/raja_view_blur.cpp
```diff
@@ -0,0 +1,170 @@
...
+  int * array_copy = def_host_res.allocate<int>(N * N);
+  
+  //big array, or image
+  for (int i = 0; i < N * N; ++i) {
+    array[i] = 1;
+    array_copy[i] = 1;
+  }
+
+  //small array that acts as the blur
+  //int* kernel = new int[K * K];
```
johnbowen42, line 78: dead code?

benchmark/CMakeLists.txt
```diff
@@ -18,3 +18,7 @@ raja_add_benchmark(
 raja_add_benchmark(
   NAME ltimes
   SOURCES ltimes.cpp)
+
+raja_add_bench_mark(
```
johnbowen42, line 22: maybe add a guard here to only compile the benchmark for GPU or serial backends?  Or add kernel definitions for OpenMP

johnbowen42 (approved):
LGTM, just some small nits about making sure the benchmark makes sense for all backends

benchmark/CMakeLists.txt
artv3: I just took a pass at the driver, I tried to follow the ltimes version in which we can use define which version to run

artv3:
@johnbowen42 @rhornung67 can I get another review? I just pushed up the changes I thought I had pushed up.

rhornung67 (approved):

adayton1 (approved):

PR #1729, opened 2024-09-03
Adding Radeon support by controlling wave size

seanofthemillers:
# Summary

This is a feature. It adds a configuration time control parameter for the default wave size. On AMD MI cards this is generally 64, however on Radeon (gaming) cards this is usually 32. These changes will require the user to know if the card is setup for Wave32 or Wave64.

I also added a fix for the dynamically sized memory allocation which seems to trigger a lot of warnings for ROCm 6.2.

1. First idea or goal
2. Second idea
3. Third idea

- (impractical)
- (too big)
- (not a good idea anyway)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,6 +171,7 @@ endif()
 
 if(RAJA_ENABLE_HIP)
   message(STATUS "HIP version: ${hip_VERSION}")
+  set(RAJA_HIP_WAVESIZE "64" CACHE STRING "Set the wave size for GPU architecture. E.g. MI200/MI300 this is 64.")
   if("${hip_VERSION}" VERSION_LESS "3.5")
     message(FATAL_ERROR "Trying to use HIP/ROCm version ${hip_VERSION}. RAJA requires HIP/ROCm version 3.5 or newer. ")
   endif()
--- a/include/RAJA/pattern/kernel/InitLocalMem.hpp
+++ b/include/RAJA/pattern/kernel/InitLocalMem.hpp
@@ -77,23 +77,16 @@ struct StatementExecutor<statement::InitLocalMem<RAJA::cpu_tile_mem,camp::idx_se
     using varType = typename camp::tuple_element_t<Pos, typename camp::decay<Data>::param_tuple_t>::value_type;
 
     // Initialize memory
-#ifdef RAJA_COMPILER_MSVC
-    // MSVC doesn't like taking a pointer to stack allocated data?!?!
     varType *ptr = new varType[camp::get<Pos>(data.param_tuple).size()];
     camp::get<Pos>(data.param_tuple).set_data(ptr);
-#else
-    varType Array[camp::get<Pos>(data.param_tuple).size()];
-    camp::get<Pos>(data.param_tuple).set_data(&Array[0]);
-#endif
+
 
     // Initialize others and execute
     exec_expanded<others...>(data);
 
     // Cleanup and return
     camp::get<Pos>(data.param_tuple).set_data(nullptr);
-#ifdef RAJA_COMPILER_MSVC
     delete[] ptr;
-#endif
   }
   
 
--- a/include/RAJA/pattern/kernel/InitLocalMem.hpp
+++ b/include/RAJA/pattern/kernel/InitLocalMem.hpp
@@ -77,16 +77,23 @@ struct StatementExecutor<statement::InitLocalMem<RAJA::cpu_tile_mem,camp::idx_se
     using varType = typename camp::tuple_element_t<Pos, typename camp::decay<Data>::param_tuple_t>::value_type;
 
     // Initialize memory
+#ifdef RAJA_COMPILER_MSVC
+    // MSVC doesn't like taking a pointer to stack allocated data?!?!
     varType *ptr = new varType[camp::get<Pos>(data.param_tuple).size()];
     camp::get<Pos>(data.param_tuple).set_data(ptr);
-
+#else
+    varType Array[camp::get<Pos>(data.param_tuple).size()];
+    camp::get<Pos>(data.param_tuple).set_data(&Array[0]);
+#endif
 
     // Initialize others and execute
     exec_expanded<others...>(data);
 
     // Cleanup and return
     camp::get<Pos>(data.param_tuple).set_data(nullptr);
+#ifdef RAJA_COMPILER_MSVC
     delete[] ptr;
+#endif
   }
   
 
--- a/include/RAJA/policy/tensor/arch/hip/hip_wave.hpp
+++ b/include/RAJA/policy/tensor/arch/hip/hip_wave.hpp
@@ -57,7 +57,7 @@ namespace expt
 
 		public:
 
-      static constexpr int s_num_elem = 64;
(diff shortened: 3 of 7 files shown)


include/RAJA/policy/tensor/arch/hip/hip_wave.hpp
```diff
@@ -780,8 +780,8 @@ namespace expt
 
         // Third: mask off everything but output_segment
         //        this is because all output segments are valid at this point
-        // (5-segbits), the 5 is since the warp-width is 32 == 1<<5
-        int our_output_segment = get_lane()>>(6-segbits);
+        const int log2_warp_size = 32-1-__builtin_clz(warpSize);
```
MrBurmark, line 783: By using warpSize it appear that this is making a runtime value now instead of a compile time value, do we need to worry about the performance implications of this?

seanofthemillers: It will add a few instructions and possibly a sgpr load. It probably won't be noticeable, but we can move the log2_warp_size as a constexpr member of the DeviceConstants. Or just wrap these two places in ifdefs. If we keep it dynamic then it should work for Nvidia/other as well - I'm not sure how important that is.

MrBurmark: Given that we're already relying on getting the right warpSize at configure time I think we should keep this compile time if possible. There is a constexpr RAJA::log2 that we can use instead of clz here to keep it constexpr.

seanofthemillers: I finally had a chance to update this. I switched it over to the DeviceConstants/WARP_SIZE value using the RAJA::log2 call.

include/RAJA/pattern/kernel/InitLocalMem.hpp
```diff
@@ -77,23 +77,16 @@ struct StatementExecutor<statement::InitLocalMem<RAJA::cpu_tile_mem,camp::idx_se
...
 
     // Initialize memory
-#ifdef RAJA_COMPILER_MSVC
-    // MSVC doesn't like taking a pointer to stack allocated data?!?!
     varType *ptr = new varType[camp::get<Pos>(data.param_tuple).size()];
     camp::get<Pos>(data.param_tuple).set_data(ptr);
-#else
-    varType Array[camp::get<Pos>(data.param_tuple).size()];
-    camp::get<Pos>(data.param_tuple).set_data(&Array[0]);
-#endif
```
MrBurmark, line 87: Is this related to this PR?

CMakeLists.txt
```diff
@@ -171,6 +171,7 @@ endif()
 
 if(RAJA_ENABLE_HIP)
   message(STATUS "HIP version: ${hip_VERSION}")
+  set(RAJA_DEFAULT_WAVESIZE "64" CACHE STRING "Default wave size for GPU architecture. E.g. MI200/MI300 this is 64.")
```
MrBurmark, line 174: This should probably have HIP in the name.

include/RAJA/policy/tensor/arch/hip/hip_wave.hpp
```diff
@@ -780,8 +780,8 @@ namespace expt
 
         // Third: mask off everything but output_segment
         //        this is because all output segments are valid at this point
-        // (5-segbits), the 5 is since the warp-width is 32 == 1<<5
-        int our_output_segment = get_lane()>>(6-segbits);
+        const int log2_warp_size = RAJA::log2(RAJA::policy::hip::device_constants.WARP_SIZE);
```
MrBurmark, line 783:
```suggestion
        constexpr int log2_warp_size = RAJA::log2(RAJA::policy::hip::device_constants.WARP_SIZE);
```

CMakeLists.txt
seanofthemillers: Switched it to RAJA_HIP_WAVESIZE - I'm not sure 'default' was needed.

include/RAJA/policy/tensor/arch/hip/hip_wave.hpp
seanofthemillers: Fixed in both places.

include/RAJA/pattern/kernel/InitLocalMem.hpp
seanofthemillers: This was to address a warning that pops up in clang 18 for dynamic stack variables. I can move to a separate issue, but it was small enough I thought I would just throw it in here.

MrBurmark: I was wondering if anyone knew if we'd gotten past the MSVC problem from the comment.

seanofthemillers: It would appear that the compiler is not treating it as compile time knowledge - I'm not sure why. Should I remove this change from this branch to be worked on separately?

MrBurmark: Let's do that for now.

include/RAJA/policy/tensor/arch/hip/hip_wave.hpp
```diff
@@ -780,8 +780,8 @@ namespace expt
 
         // Third: mask off everything but output_segment
         //        this is because all output segments are valid at this point
-        // (5-segbits), the 5 is since the warp-width is 32 == 1<<5
-        int our_output_segment = get_lane()>>(6-segbits);
+        constexpr int log2_warp_size = RAJA::log2(RAJA::policy::hip::device_constants.WARP_SIZE);
```
MrBurmark, line 783:
```suggestion
        static constexpr int log2_warp_size = RAJA::log2(RAJA::policy::hip::device_constants.WARP_SIZE);
```

include/RAJA/policy/tensor/arch/hip/hip_wave.hpp
```diff
@@ -828,8 +828,9 @@ namespace expt
 
         // First: tree reduce values within each segment
         element_type x = m_value;
+        constexpr int log2_warp_size = RAJA::log2(RAJA::policy::hip::device_constants.WARP_SIZE);
```
MrBurmark, line 831:
```suggestion
        static constexpr int log2_warp_size = RAJA::log2(RAJA::policy::hip::device_constants.WARP_SIZE);
```

test/include/RAJA_test-tensor.hpp
```diff
@@ -87,7 +87,9 @@ struct TensorTestHelper<RAJA::expt::hip_wave_register>
     void exec(BODY const &body){
       hipDeviceSynchronize();
 
-      RAJA::forall<RAJA::hip_exec<64>>(RAJA::RangeSegment(0,64),
+      constexpr int warp_size = RAJA::policy::hip::device_constants.WARP_SIZE;
```
MrBurmark, line 90:
```suggestion
      static constexpr int warp_size = RAJA::policy::hip::device_constants.WARP_SIZE;
```

MrBurmark (approved):

rhornung67 (approved):

PR #1760, opened 2024-11-05
Reshape method

artv3:
# Summary

This PR adds helper methods to views to create Reshape methods with compile time options C and Fortran style indexing


examples/reshape.cpp
```diff
@@ -0,0 +1,728 @@
...
+ *
+ *  This example performs batched matrix multiplication
+ *  for matrices of dimension 3 x 3 using two different
+ *  data layouts.
+ *
+ *  Matrices are stored in arrays A and B. Results
+ *  are stored in a third array, C.
+ *  We introduce the notation A^{e}_rc
+ *  to correspond to the matrix entry in the row, r,
+ *  column, c, of matrix, e. Below we describe the two
```
MrBurmark, line 30: I'm not sure what the right way to put commas here is, it feels a bit awkward to have so many commas so close together.

examples/reshape.cpp
```diff
@@ -0,0 +1,728 @@
...
+//
+  double *A = memoryManager::allocate<double>(N_c * N_r * N);
+  double *B = memoryManager::allocate<double>(N_c * N_r * N);
+  double *C = memoryManager::allocate<double>(N_c * N_r * N);
+
+//
+// Layout 1
+//
+// make_permuted_layout takes the number of entries in each dimension and a
+// templated array indicating index arguments with slowest to fastest stride.
```
MrBurmark, line 119: Should this just be an array? Do you mean to say that any array-like class can be used or that it has to be a std::array but the types stored in the array can vary but have to be convertible to integers, or something else?

examples/reshape.cpp
```diff
@@ -0,0 +1,728 @@
...
+// Standard C++ arrays are used to hold the number of entries in each component.
+// This example uses double braces to initalize the array and its subobjects.
+// The layout object will index into the array as the following C macro would
+// #define Aview(e, r, c) A[c + N_c*(r + N_r*e)].
+//
+// RAJA::Layout objects may be templated on dimension, argument type, and
+// index with unit stride. Here, the column index has unit stride (argument 2).
```
MrBurmark, line 126: Can you supply the permutation at compile time and have it produce a view with the right unit stride index?

examples/reshape.cpp
```diff
@@ -0,0 +1,728 @@
...
+  double *C2 = memoryManager::allocate<double>(N_c * N_r * N);
+
+//
+// Permuted layout - equivalent to indexing using the following macro
+// #define Aview2(e, r, c) A2[e + N*(c + N_c*r)]
+// In this case the element index has unit stride (argument 0).
+//
+  // _permutedlayout_permviews_start
+#if 0
+    std::array<RAJA::idx_t, 3> perm2 {{0, 1, 2}};
```
MrBurmark, line 160: Isn't perm 2 the same as perm 1?

examples/reshape.cpp
```diff
@@ -0,0 +1,728 @@
...
+  for (int i = 0; i < NITER; ++i) {
+
+    timer.start();
+    // _permutedlayout_batchedmatmult_loop_start
+    RAJA::forall<RAJA::seq_exec>(RAJA::TypedRangeSegment<int>(0, N),
+      [=](int e) {
+
+        Cview(e, 0, 0) = Aview(e, 0, 0) * Bview(e, 0, 0)
+                         + Aview(e, 0, 1) * Bview(e, 1, 0)
+                         + Aview(e, 0, 2) * Bview(e, 2, 0);
```
MrBurmark, line 219: This spacing does not spark joy.

include/RAJA/util/View.hpp
```diff
@@ -296,6 +297,43 @@ RAJA_INLINE AtomicViewWrapper<ViewType, AtomicPolicy> make_atomic_view(
...
+  }
+};
+
+template<>
+struct Reshape<layout_left>
+{
+  template<typename T, typename...Ts>
+  static auto get(T *ptr, Ts... s)
+  {
+  constexpr int N = sizeof...(Ts);
```
MrBurmark, line 323: spacing

include/RAJA/util/View.hpp
```diff
@@ -296,6 +297,43 @@ RAJA_INLINE AtomicViewWrapper<ViewType, AtomicPolicy> make_atomic_view(
...
+
+  std::array<RAJA::idx_t, N> extent{s...};
+
+  ///Should be a away to do this at compile time...
+  std::array<RAJA::idx_t, N> reverse_indices_array;
+  for(int i = N-1, j=0; i>-1; --i, j++) {reverse_indices_array[j] = i; }
```
MrBurmark, line 329: Make an array of `{s...}`, and an integer_sequence of `0...(N-1)`, pass them into a function and then you'll have all you need to do the indexing.

include/RAJA/util/View.hpp
```diff
@@ -296,6 +297,43 @@ RAJA_INLINE AtomicViewWrapper<ViewType, AtomicPolicy> make_atomic_view(
...
+struct Reshape;
+
+template<>
+struct Reshape<layout_right>
+{
+  template<typename T, typename...Ts>
+  static auto get(T *ptr, Ts... s)
+  {
+    constexpr int N = sizeof...(Ts);
+    return RAJA::View<T, RAJA::Layout<N, RAJA::Index_type, N-1>>(ptr, s...);
```
MrBurmark, line 313: Can you add an alias for the view_type?

include/RAJA/util/View.hpp
```diff
@@ -296,6 +297,43 @@ RAJA_INLINE AtomicViewWrapper<ViewType, AtomicPolicy> make_atomic_view(
   return RAJA::AtomicViewWrapper<ViewType, AtomicPolicy>(view);
 }
 
+struct layout_left{};
+struct layout_right{};
+
+template<typename LAYOUT>
```
MrBurmark, line 303: Can we add a specialization for a generic index sequence like std::index_sequence?

include/RAJA/util/View.hpp
artv3: @MrBurmark I just pushed up a new version, would you be able to provide some feedback?

include/RAJA/util/View.hpp
```diff
@@ -296,6 +297,48 @@ RAJA_INLINE AtomicViewWrapper<ViewType, AtomicPolicy> make_atomic_view(
...
+struct Reshape<layout_left>
+{
+  template<typename T, typename...Ts>
+  static auto get(T *ptr, Ts... s)
+  {
+  constexpr int N = sizeof...(Ts);
+
+  std::array<RAJA::idx_t, N> extent{s...};
+
+  constexpr auto reverse_array = make_reverse_array(std::make_index_sequence<N>{});
```
artv3, line 334: @MrBurmark this is the revised implementation

MrBurmark: Nice, this is what I was thinking.

examples/reshape.cpp
rhornung67: Why not just simplify is like this?

We use the notation A^{e}_rc to indicate the entry in matrix e in row r at column c.

examples/reshape.cpp
rhornung67: Also, since it's an example, would making it a doubly-nested loop to compute each CView(e, i, j) entry make it clearer and easier to read?

examples/reshape.cpp
```diff
@@ -0,0 +1,728 @@
...
+
+  minRun = std::numeric_limits<double>::max();
+  for (int i = 0; i < NITER; ++i) {
+
+    timer.start();
+    // _permutedlayout2_batchedmatmult_loop_start
+    RAJA::forall<RAJA::seq_exec>(RAJA::TypedRangeSegment<int>(0, N),
+      [=](int e) {
+
+        Cview2(e, 0, 0) = Aview2(e, 0, 0) * Bview2(e, 0, 0)
```
rhornung67, line 272: Similar comment to above

examples/reshape.cpp
```diff
@@ -0,0 +1,728 @@
...
+
+  minRun = std::numeric_limits<double>::max();
+  for (int i = 0; i < NITER; ++i) {
+
+    timer.start();
+    // _permutedlayout_batchedmatmult_omp_start
+    RAJA::forall<RAJA::omp_parallel_for_exec>(RAJA::TypedRangeSegment<int>(0, N),
+      [=](int e) {
+
+        Cview(e, 0, 0) = Aview(e, 0, 0) * Bview(e, 0, 0)
```
rhornung67, line 331: Similar comment

artv3:
@MrBurmark , @rhornung67 I think the example for the reshape method can be simplified. I just pushed up a new version. I was a bit worried the previous version might have too much going on.

include/RAJA/util/View.hpp
artv3: Sure, I'll take a pass at this and then I'll remove the "Draft" marking from this PR.

examples/reshape.cpp
```diff
@@ -0,0 +1,134 @@
...
+  memoryManager::deallocate(Lptr);
+
+  std::cout << "\n DONE!...\n";
+  return 0;
+}
+
+//
+// check result
+//
+void checkResult(int *ptr, int N, int M, int K)
```
MrBurmark, line 113: NMK here?

examples/reshape.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+
+  bool status = true;
+
+  for(int k = 0; k < K; ++k) {
+    for(int n = 0; n < N; ++n) {
+      for(int m = 0; m < M; ++m) {
+        const int idx = m + M * (n + N * k);
+        if (std::abs(ptr[idx] - idx) > 1.0e-12) {
```
MrBurmark, line 122: These are ints.

examples/reshape.cpp
artv3: yes, thanks!

examples/reshape.cpp
artv3: yes, thanks!

examples/reshape.cpp
artv3: yes, I pushed a version for that just now. Let me know what you think.

include/RAJA/util/View.hpp
```diff
@@ -296,6 +297,76 @@ RAJA_INLINE AtomicViewWrapper<ViewType, AtomicPolicy> make_atomic_view(
...
+struct Reshape;
+
+template<std::size_t Id, std::size_t...Ids>
+constexpr auto get_first_index()
+{
+  return Id;
+}
+
+template<std::size_t...Is>
+struct Reshape<std::index_sequence<Is...>>
```
artv3, line 316: @MrBurmark , this is the specialization that takes an index_sequence and identifies the correct unit stride

examples/reshape.cpp
```diff
@@ -0,0 +1,162 @@
...
+  int *Rptr = memoryManager::allocate<int>(K * N * M);
+  int *Lptr = memoryManager::allocate<int>(K * N * M);
+  int *Cptr = memoryManager::allocate<int>(K * N * M);
+
+//----------------------------------------------------------------------------//
+//
+// Initialize memory using right most unit stride
+//
+//----------------------------------------------------------------------------//
+  std::cout << "\n\nInitialize array with right most indexing...\n";
```
rhornung67, line 64: I think these implementations are much easier to understand if your intent is just to show the mechanics of View reshaping. 👍

include/RAJA/util/View.hpp
```diff
@@ -296,6 +297,76 @@ RAJA_INLINE AtomicViewWrapper<ViewType, AtomicPolicy> make_atomic_view(
...
+  {
+    constexpr int N = sizeof...(Ts);
+    std::array<RAJA::idx_t, N> extent{s...};
+
+    auto custom_layout =
+      RAJA::make_permuted_layout(extent, std::array<RAJA::idx_t, N>{Is...});
+
+    constexpr auto unit_stride = get_first_index<Is...>();
+
+    return RAJA::View<T, RAJA::Layout<N, RAJA::Index_type, unit_stride>>
```
artv3, line 329: Bug found here! unit_stride should be the last value of the prameter pack -- TODO: FIX

artv3:
I fixed a bug with the custom sequence, thanks @rhornung67 for point it out. This PR is now ready for review.

include/RAJA/util/View.hpp
artv3: Now fixed.

examples/reshape.cpp
```diff
@@ -0,0 +1,163 @@
...
+void checkResult(int *ptr, int K, int N, int M)
+{
+
+  bool status = true;
+
+  for(int k = 0; k < K; ++k) {
+    for(int n = 0; n < N; ++n) {
+      for(int m = 0; m < M; ++m) {
+        const int idx = m + M * (n + N * k);
+        if (std::abs(ptr[idx] - idx) > 0) {
```
MrBurmark, line 151: How about `ptr[idx] == idx`?

include/RAJA/util/View.hpp
```diff
@@ -296,6 +297,82 @@ RAJA_INLINE AtomicViewWrapper<ViewType, AtomicPolicy> make_atomic_view(
...
+  {
+    constexpr int N = sizeof...(Ts);
+    using view_t = RAJA::View<T, RAJA::Layout<N, RAJA::Index_type, N-1>>;
+
+    return view_t(ptr, s...);
+  }
+};
+
+template<std::size_t... Is>
+constexpr std::array<RAJA::idx_t, sizeof...(Is)> make_reverse_array(std::index_sequence<Is...>) {
```
MrBurmark, line 354: The helpers like get_last_index and make_reverse_array should be in a detail namespace.

include/RAJA/util/View.hpp
```diff
@@ -296,6 +297,82 @@ RAJA_INLINE AtomicViewWrapper<ViewType, AtomicPolicy> make_atomic_view(
...
 }
 
+struct layout_left{};
+struct layout_right{};
+
+template<typename LAYOUT>
+struct Reshape;
+
+template<typename LAYOUT>
+struct Reshape;
```
MrBurmark, line 307: declare once

include/RAJA/util/View.hpp
```diff
@@ -296,6 +297,82 @@ RAJA_INLINE AtomicViewWrapper<ViewType, AtomicPolicy> make_atomic_view(
...
+};
+
+template<>
+struct Reshape<layout_right>
+{
+  template<typename T, typename...Ts>
+  static auto get(T *ptr, Ts... s)
+  {
+    constexpr int N = sizeof...(Ts);
+    using view_t = RAJA::View<T, RAJA::Layout<N, RAJA::Index_type, N-1>>;
```
MrBurmark, line 347: In general spacing and when aliases or temporaries are used could be more consistent.

include/RAJA/util/View.hpp
MrBurmark: Makes sense, I see an example is documentation coming?

include/RAJA/util/View.hpp
artv3: good call!

examples/reshape.cpp
artv3: Sounds good to me!

include/RAJA/util/View.hpp
artv3: ah! missed this, thanks

include/RAJA/util/View.hpp
artv3: yes, good call. I just hit with a focused version of clang-format and opted for using aliases when the view types became long.

include/RAJA/util/View.hpp
```diff
@@ -296,6 +295,89 @@ RAJA_INLINE AtomicViewWrapper<ViewType, AtomicPolicy> make_atomic_view(
...
+namespace detail
+{
+template <typename T>
+constexpr auto get_last_index(T last)
+{
+  return last;
+}
+
+template <typename T, typename... Args>
+constexpr auto get_last_index(T, Args... args)
```
MrBurmark, line 315: I think this should be like this so it can't be called with 1 argument.
```suggestion
template <typename T0, typename T1, typename... Args>
constexpr auto get_last_index(T0, T1, Args... args)
```

include/RAJA/util/View.hpp
artv3: Yes, thanks for reminding me. I'll make something in the read the docs page within the next few days.

include/RAJA/util/View.hpp
MrBurmark: I didn't fix the function call inside though

include/RAJA/util/View.hpp
```diff
@@ -312,9 +313,9 @@ constexpr auto get_last_index(T last)
 }
 
 template <typename T0, typename T1, typename... Args>
-constexpr auto get_last_index(T0, T1, Args... args)
+constexpr auto get_last_index(T0, T1 t1, Args... args)
 {
-  return get_last_index(args...);
+  return get_last_index(t1, args...);
```
artv3, line 318: @MrBurmark is this what you had in mind?

rhornung67:
@artv3 I think you need to run the clang-format style check on this.

artv3:
@LLNL/raja-core could I get a review? I pushed up docs --- we have some users who would like to start using this in the next release.

docs/sphinx/user_guide/feature/view.rst
```diff
@@ -92,21 +92,57 @@ accesses array entries with unit stride. The loop::
 
 access array entries with stride N :subscript:`n` * N :subscript:`(n-1)` * ... * N :subscript:`(j+1)`.
 
+Reshape
+^^^^^^^
+The ``RAJA::Reshape`` method is designed as a simplified multi-dimensional pointer abstraction
+only requiring the pointer, dimensions, and prefered layout type based on a template specilization.
```
MrBurmark, line 98: This seems too focused on how its implemented via specialization than how its used. Could this just say that you need to provide a pointer, dimensions, and a layout type?

artv3: Sure, I think we can make that works. I coined the term meta-layout as predefined layout types as suggested.

johnbowen42 (approved):
LGTM

artv3:
All green! anyone else wants to take a last look? otherwise I can merge at the end of the day

docs/sphinx/user_guide/feature/view.rst
```diff
@@ -92,21 +92,56 @@ accesses array entries with unit stride. The loop::
 
 access array entries with stride N :subscript:`n` * N :subscript:`(n-1)` * ... * N :subscript:`(j+1)`.
 
+Reshape
+^^^^^^^
+The ``RAJA::Reshape`` method will provide developers a ``RAJA::View`` given a pointer, dimentions and a meta-layout.
```
rhornung67, line 97:
```suggestion
The ``RAJA::Reshape`` method takes a pointer to data and extents in each dimension as arguments and returns a ``RAJA::View`` object. The reshape method is specialized on a `meta-layout` template parameter, which specifies the stride ordering for the dimensions.

```

docs/sphinx/user_guide/feature/view.rst
```diff
@@ -92,21 +92,56 @@ accesses array entries with unit stride. The loop::
 
 access array entries with stride N :subscript:`n` * N :subscript:`(n-1)` * ... * N :subscript:`(j+1)`.
 
+Reshape
+^^^^^^^
+The ``RAJA::Reshape`` method will provide developers a ``RAJA::View`` given a pointer, dimentions and a meta-layout.
+Meta-layouts are predefined types which specify stride ordering. The first meta-layout is
```
rhornung67, line 98:
```suggestion
The first meta-layout is

```
after applying the change above.

docs/sphinx/user_guide/feature/view.rst
```diff
@@ -92,21 +92,56 @@ accesses array entries with unit stride. The loop::
...
+The second supported meta-layout is the ``RAJA::layout_left`` which will follow a Fortran style indexing
+in which the left most index has unit stride and the right most has the longest stride. The configuration
+is demonstrated in the following example.
+
+.. literalinclude:: ../../../../examples/reshape.cpp
+   :start-after: _Reshape_left_start
+   :end-before: _Reshape_left_end
+   :language: C++
+
+Finally, users may specify a custom index ordering by providing a ``std::index_sequence`` list as a
```
rhornung67, line 120:
```suggestion
Finally, users may specify a custom index ordering by providing a ``std::index_sequence`` type as a

```

docs/sphinx/user_guide/feature/view.rst
```diff
@@ -92,21 +92,56 @@ accesses array entries with unit stride. The loop::
...
+template argument.
+
+.. literalinclude:: ../../../../examples/reshape.cpp
+   :start-after: _Reshape_custom_start
+   :end-before: _Reshape_custom_end
+   :language: C++
+
+The example above will give unit stride to the right most index of the sequence while the left most index will have
```
rhornung67, line 128: The example code for this has entry '0' for the second entry in the sequence. I think that requires a sentence of explanation.

rhornung67 (changes requested):
Please consider my documentation suggestions and removing unused code from the example.

docs/sphinx/user_guide/feature/view.rst
```diff
@@ -96,7 +96,8 @@ Reshape
 ^^^^^^^
 The ``RAJA::Reshape`` method takes a pointer to data and extents in each dimension as arguments and returns a ``RAJA::View`` object. The reshape method is specialized on a `meta-layout` template parameter, which specifies the stride ordering for the dimensions.
 
-Meta-layouts are predefined types which specify stride ordering. The first meta-layout is
+The first meta-layout is
+
```
rhornung67, line 100: Remove the blank line 100. Otherwise, Sphinx thinks line 99 is a separate paragraph.

artv3:
Thanks @rhornung67 , I believe I have address all review comments.

rhornung67 (approved):

docs/sphinx/user_guide/feature/view.rst
```diff
@@ -92,21 +92,59 @@ accesses array entries with unit stride. The loop::
...
+.. literalinclude:: ../../../../examples/reshape.cpp
+   :start-after: _Reshape_left_start
+   :end-before: _Reshape_left_end
+   :language: C++
+
+Finally, users may specify a custom index ordering by providing a ``std::index_sequence`` type as a
+template argument for the ``RAJA::Reshape`` method. A ``RAJA::View`` enumerates the parathesis accessor
+from left to right starting at zero. The ``std::index_sequence`` will hold an ordering of slowest (left most)
+to fastest (right most) index for the ``RAJA::View``. In the following example, unit stride will be given to
+index 1 while index 2 is assumed to have the unit stride.
```
MrBurmark, line 122: What is this saying? What's the parathesis accessor? Which has the unit stride?

rhornung67: I was confused by this also. The middle entry in the `std::index_sequence` is zero in the example, which requires explanation, IMO.

docs/sphinx/user_guide/feature/view.rst
```diff
@@ -92,21 +92,62 @@ accesses array entries with unit stride. The loop::
...
+   :start-after: _Reshape_left_start
+   :end-before: _Reshape_left_end
+   :language: C++
+
+Finally, users may specify a custom index ordering by providing a ``std::index_sequence`` type as a
+template argument to the ``RAJA::Reshape`` method. The members of the ``std::index_sequence`` enumerate
+the striding order of the arguments of the ``RAJA::View`` parenthesis operator.
+
+For example to use the C index ordering (``RAJA::layout_right``) with a 3-dimensional ``RAJA::View`` use the following 
+sequence type ``std::index_sequence<0U,1U,2U>``.  In the following example, we consider a permutation
```
rhornung67, line 123: I think this would be clearer if you added a sentence that states exactly what the indexing order is for this case.

artv3:
Just pushed up an iteration which ties Reshape to the permuted layout usage in view.

examples/reshape.cpp
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  checkResult(Cptr, K, N, M);
+
+
+//
+// Clean up.
+//
+  memoryManager::deallocate(Rptr);
+  memoryManager::deallocate(Lptr);
```
rchen20, line 120: Should we also deallocate `Cptr`?

rchen20:
I like this idea: it makes permutations and layouts much easier to use. However, I think it should be called something like `ViewMaker`, or something else. When I see `Reshape`, I think that this will take an existing View I've already defined, and will alter its dimensions. Instead, this PR is creating a wrapper that returns a new View, and can be used to construct Views from scratch. I realize that `ReShape` is used in a similar way in MFEM as well, so if everyone else is on board with this then that's fine with me too.

rhornung67:
> I like this idea: it makes permutations and layouts much easier to use. However, I think it should be called something like `ViewMaker`, or something else. When I see `Reshape`, I think that this will take an existing View I've already defined, and will alter its dimensions. Instead, this PR is creating a wrapper that returns a new View, and can be used to construct Views from scratch. I realize that `ReShape` is used in a similar way in MFEM as well, so if everyone else is on board with this then that's fine with me too.

That's actually a good point. What about `make_permuted_view`?  We have several `make_*_view` helpers and this would be consistent with what we already have.

artv3:
> > I like this idea: it makes permutations and layouts much easier to use. However, I think it should be called something like `ViewMaker`, or something else. When I see `Reshape`, I think that this will take an existing View I've already defined, and will alter its dimensions. Instead, this PR is creating a wrapper that returns a new View, and can be used to construct Views from scratch. I realize that `ReShape` is used in a similar way in MFEM as well, so if everyone else is on board with this then that's fine with me too.
> 
> That's actually a good point. What about `make_permuted_view`? We have several `make_*_view` helpers and this would be consistent with what we already have.

Maybe we can debate this at the RAJA Hackathon, this falls in line with documentation and our desire to perform general clean up.

rhornung67:
Coming up with good, intuitive names is hard.

MrBurmark:
I second make_permuted_view. That's more in line with our existing API.

rchen20:
Another idea would be to add this functionality as partial specializations of the existing `make_view` function, and get rid of the `ReShape` object entirely. What's implemented here is a way to create both permuted and non-permuted Views, so the functionality is generic enough to sit under the `make_view` umbrella.

examples/reshape.cpp
```diff
@@ -0,0 +1,151 @@
...
+  int *Cptr = memoryManager::allocate<int>(K * N * M);
+
+//----------------------------------------------------------------------------//
+//
+// Initialize memory using right most unit stride
+//
+//----------------------------------------------------------------------------//
+  std::cout << "\n\nInitialize array with right most indexing...\n";
+  // _Reshape_right_start
+  auto Rview = RAJA::Reshape<RAJA::layout_right>::get(Rptr, K, N, M);
```
artv3, line 49: consider: auto Rview = RAJA::ViewMaker<RAJA::layout_right>::get(Rptr, K, N, M);

artv3:
Thank you all, through teamwork we have improved the API and implementation details. I believe we are ready for another review.

rchen20 (approved):

artv3:
@MrBurmark, if you have a second can you provide a review here? I think we are good to go once you take a look.

MrBurmark (approved):

PR #1768, opened 2024-11-14
[WIP] From RSC: Use llvm-amdgpu instead of hip to determine ROCM_PATH

adrienbernede:
Test https://github.com/LLNL/radiuss-spack-configs/pull/114 from RADIUSS Spack Configs


adrienbernede:
Testing completed.

PR #1791, opened 2025-02-11
Fix kernel to not launch a kernel if there is no work

### LINKED ISSUE #1733: Kernel may launch GPU kernel with no work

MrBurmark:
# Fix kernel to not launch a kernel if there is no work

Track which dimensions are active when calculating dimensions in kernel. This allows us to know that we have no work to do in some cases where there was ambiguity before and kernels were launched despite having no work.

- This PR is a bugfix
  - Fixes #1733

### DIFF
--- a/include/RAJA/policy/cuda/kernel/internal.hpp
+++ b/include/RAJA/policy/cuda/kernel/internal.hpp
@@ -46,19 +46,20 @@ namespace internal
 struct LaunchDims
 {
 
-  CudaDims dims;
-  CudaDims min_dims;
+  CudaDims active{0};
+  CudaDims dims{0};
+  CudaDims min_dims{0};
 
   LaunchDims()                             = default;
   LaunchDims(LaunchDims const&)            = default;
+  LaunchDims(LaunchDims &&)                = default;
   LaunchDims& operator=(LaunchDims const&) = default;
+  LaunchDims& operator=(LaunchDims &&)     = default;
 
   RAJA_INLINE
-  LaunchDims(CudaDims _dims) : dims {_dims}, min_dims {} {}
-
-  RAJA_INLINE
-  LaunchDims(CudaDims _dims, CudaDims _min_dims)
-      : dims {_dims},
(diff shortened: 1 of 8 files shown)


include/RAJA/policy/cuda/kernel/internal.hpp
```diff
@@ -259,16 +286,23 @@ struct KernelDimensionCalculator<RAJA::policy::cuda::cuda_indexer<
...
           "len exceeds the size of the directly mapped index space");
     }
-    set_cuda_dim<dim>(dims.threads, static_cast<IdxT>(IndexMapper::block_size));
-    set_cuda_dim<dim>(min_dims.threads,
-                      static_cast<IdxT>(IndexMapper::block_size));
+
+    LaunchDims dims;
+
+    set_cuda_dim<dim>(dims.  active.threads, static_cast<cuda_dim_member_t>(true));
+    set_cuda_dim<dim>(dims.    dims.threads, static_cast<cuda_dim_member_t>((len > zero) ? IndexMapper::block_size : 0));
```
adayton1, line 302: Can you use "zero" instead of "0"?

adayton1: Oh, I didn't realize it was an int in these cases. I don't think it really adds any clarity to make changes, so it's probably not worth worrying about.

PR #1792, opened 2025-02-18
Rename param implementation

MrBurmark:
# Rename param implementation

Prepend param_ to the init, combine, and resolve overloads used to implement the params interface.
This avoids having to make a call to a specific function in detail and should allow the interface to be more extensible. Perhaps even to the point of letting users add params that could look at things like the cuda/hip launch parameters.
Also pass the policy by value to avoid having to specifying the template parameter explicitly.

- This PR is a refactoring
  - refactors params to be more easily greppable and more extensible

### DIFF
--- a/include/RAJA/pattern/params/forall.hpp
+++ b/include/RAJA/pattern/params/forall.hpp
@@ -43,45 +43,45 @@ struct ForallParamPack
 private:
   // Init
   template<typename EXEC_POL, camp::idx_t... Seq, typename... Args>
-  static constexpr void params_init(EXEC_POL,
+  static constexpr void params_init(EXEC_POL const& pol,
                                     camp::idx_seq<Seq...>,
                                     ForallParamPack& f_params,
                                     Args&&... args)
   {
-    CAMP_EXPAND(param_init<EXEC_POL>(camp::get<Seq>(f_params.param_tup),
-                                     std::forward<Args>(args)...));
+    CAMP_EXPAND(param_init(pol, camp::get<Seq>(f_params.param_tup),
+                                std::forward<Args>(args)...));
   }
 
   // Combine
   template<typename EXEC_POL, camp::idx_t... Seq>
   RAJA_HOST_DEVICE static constexpr void params_combine(
-      EXEC_POL,
+      EXEC_POL const& pol,
       camp::idx_seq<Seq...>,
       ForallParamPack& out,
       const ForallParamPack& in)
   {
-    CAMP_EXPAND(param_combine<EXEC_POL>(camp::get<Seq>(out.param_tup),
-                                        camp::get<Seq>(in.param_tup)));
+    CAMP_EXPAND(param_combine(pol, camp::get<Seq>(out.param_tup),
+                                   camp::get<Seq>(in.param_tup)));
   }
 
   template<typename EXEC_POL, camp::idx_t... Seq>
   RAJA_HOST_DEVICE static constexpr void params_combine(
-      EXEC_POL,
+      EXEC_POL const& pol,
       camp::idx_seq<Seq...>,
       ForallParamPack& f_params)
   {
-    CAMP_EXPAND(param_combine<EXEC_POL>(camp::get<Seq>(f_params.param_tup)));
+    CAMP_EXPAND(param_combine(pol, camp::get<Seq>(f_params.param_tup)));
   }
 
   // Resolve
   template<typename EXEC_POL, camp::idx_t... Seq, typename... Args>
-  static constexpr void params_resolve(EXEC_POL,
+  static constexpr void params_resolve(EXEC_POL const& pol,
                                        camp::idx_seq<Seq...>,
                                        ForallParamPack& f_params,
                                        Args&&... args)
   {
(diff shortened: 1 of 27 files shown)


include/RAJA/policy/sycl/launch.hpp
```diff
@@ -149,14 +150,14 @@ struct LaunchExecute<RAJA::sycl_launch_t<async, 0>>
 
 
       auto combiner = [](ReduceParams x, ReduceParams y) {
-        RAJA::expt::ParamMultiplexer::combine<EXEC_POL>(x, y);
+        RAJA::expt::ParamMultiplexer::params_combine(EXEC_POL {}, x, y);
```
adayton1, line 153: Could you pass in `pol`?

include/RAJA/policy/sycl/launch.hpp
```diff
@@ -316,7 +318,7 @@ struct LaunchExecute<RAJA::sycl_launch_t<async, 0>>
...
```
adayton1, line 321: Use `pol`?

include/RAJA/util/macros.hpp
```diff
@@ -141,9 +141,10 @@ RAJA_HOST_DEVICE RAJA_INLINE void RAJA_UNUSED_VAR(T&&...) noexcept
  */
 #if defined(RAJA_ENABLE_OPENMP)
 #define RAJA_OMP_DECLARE_REDUCTION_COMBINE                                     \
+  RAJA_UNUSED_VAR(EXEC_POL {});                                                \
```
adayton1, line 144: Is this needed?

include/RAJA/policy/sycl/launch.hpp
MrBurmark: I think the combiner is not supposed to capture anything.

include/RAJA/util/macros.hpp
MrBurmark: Without this you get lots of warnings about an unused local alias, probably because the alias is only used in the pragma.

adayton1 (approved):

include/RAJA/policy/sycl/launch.hpp
MrBurmark: Right now there isn't any runtime info in the policies so it only really matters that the type is right. It would be nice to pass the same policy object in though.

include/RAJA/pattern/params/forall.hpp
```diff
@@ -43,45 +43,45 @@ struct ForallParamPack
...
   template<typename EXEC_POL, camp::idx_t... Seq>
-  RAJA_HOST_DEVICE static constexpr void detail_combine(
-      EXEC_POL,
+  RAJA_HOST_DEVICE static constexpr void params_combine(
+      EXEC_POL const& pol,
       camp::idx_seq<Seq...>,
       ForallParamPack& f_params)
   {
-    CAMP_EXPAND(detail::combine<EXEC_POL>(camp::get<Seq>(f_params.param_tup)));
+    CAMP_EXPAND(param_combine(pol, camp::get<Seq>(f_params.param_tup)));
```
rchen20, line 73: Is the difference between `params_combine` and `param_combine` just that the former is at the loop level, and the latter is in the backend? This naming can be a little confusing and possibly error prone if we mistype a recursive call. Maybe something like `local_params_combine` or `params_combine_impl` for the backend one?

include/RAJA/pattern/params/forall.hpp
```diff
@@ -43,45 +43,45 @@ struct ForallParamPack
 private:
   // Init
   template<typename EXEC_POL, camp::idx_t... Seq, typename... Args>
-  static constexpr void detail_init(EXEC_POL,
+  static constexpr void params_init(EXEC_POL const& pol,
```
rchen20, line 46: What are we planning on doing with this execution policy object?

MrBurmark: The plan is to pass it to the implementation functions as a normal argument instead of a template argument. This makes the call to the implementation normal overloading without any explicit template parameters. That is helpful as until c++20 you have to be more explicit about which functions you might be calling when explicitly passing template parameters, hence the detail:: that was at the call site of the implementations.

include/RAJA/pattern/params/forall.hpp
MrBurmark: Ya, one calls combine on a parameter pack and one is the implementation for a single parameter. More clear differentiation is probably a good idea. How about something like `parampack_combine`?

include/RAJA/pattern/params/forall.hpp
rchen20: That makes sense, and does clean up the call sites quite a bit.

include/RAJA/pattern/params/forall.hpp
rchen20: `parampack_comine` and `param_combine`? That should be much more clear.

MrBurmark: done

rchen20 (approved):

mdavis36 (approved):

PR #1808, opened 2025-03-21
Update packages

adrienbernede:
- Update RSC (see https://github.com/LLNL/radiuss-spack-configs/pull/126 for changes)
- Optimize submodule usage in GitLab.
- Use radiuss-spack-configs version of ctest-to-junit conversion utility to avoid having to fetch BLT as a submodule.
- Point at specific commit in Spack _before compiler as nodes feature_.
- Update Spack mirror oci options syntax
- Set BLT variable to allow mpi tests to overlap parent allocation (do not run tests on another node).


johnbowen42 (approved):
LGTM

rhornung67 (approved):

PR #1817, opened 2025-04-14
Memory arena for launch

artv3:
# Summary
This PR introduces the concept of a memory arena of launch. The underlying idea is to allow users to allocate static shared memory (or could be an external memory type?) and manage it through a bump style allocator.

This PR is a work in progress.

### DIFF
--- a/include/RAJA/pattern/launch/launch_core.hpp
+++ b/include/RAJA/pattern/launch/launch_core.hpp
@@ -222,7 +222,7 @@ public:
   {
 
     // Calculate offset in bytes with a char pointer
-    void* mem_ptr = static_cast<char*>(arena_mem_ptr) + shared_arena_offset;
+    void* mem_ptr = static_cast<char*>(mem_arena_ptr) + mem_arena_offset;
 
     arena_mem_offset += bytes * sizeof(T);
 


artv3:
@YohannDudouit , I'm thinking the shared memory arena that we talked about could look be implemented something like this? At the moment I'm not sure how we can differentiate between static or dynamically allocated shared memory. What do you think?

artv3:
@tomstitt , adding you in the loop since I think we talked about this as well.

include/RAJA/pattern/launch/launch_core.hpp
```diff
@@ -208,6 +210,26 @@ class LaunchContext
...
+    void * mem_arena_ptr = static_cast<char*>(&ptr[0]);
+  }
+
+  // TODO handle alignment
+  template<typename T>
+  RAJA_HOST_DEVICE T* getArenaMemory(size_t bytes)
+  {
+
+    // Calculate offset in bytes with a char pointer
+    void* mem_ptr = static_cast<char*>(arena_mem_ptr) + shared_arena_offset;
```
YohannDudouit, line 225: Shouldn't it be this instead?
```suggestion
    void* mem_ptr = static_cast<char*>(mem_arena_ptr) + mem_arena_offset;
```

include/RAJA/pattern/launch/launch_core.hpp
```diff
@@ -208,6 +210,26 @@ class LaunchContext
...
 
+  template<typename T>
+  RAJA_HOST_DEVICE createMemoryArena(T *ptr, size_t bytes)
+  {
+    void * mem_arena_ptr = static_cast<char*>(&ptr[0]);
+  }
+
+  // TODO handle alignment
+  template<typename T>
+  RAJA_HOST_DEVICE T* getArenaMemory(size_t bytes)
```
YohannDudouit, line 221: I don't think `bytes` reflects well how this is used, it seems to be a number of values, not a number of bytes.

PR #1825, opened 2025-04-21
Added exclusive prefix sum to examples

SavannaSpayd:
# Summary

- This PR is a new example
  - Adds a new exclusive prefix sum example at the request of Savanna Spayd

### DIFF
--- a/examples/prefix_sum.cpp
+++ b/examples/prefix_sum.cpp
@@ -1,0 +1,156 @@
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+// Copyright (c) 2016-25, Lawrence Livermore National Security, LLC
+// and RAJA project contributors. See the RAJA/LICENSE file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#include <cstdlib>
+#include <iostream>
+#include <iomanip>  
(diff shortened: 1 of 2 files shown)


examples/prefix_sum.cpp
```diff
@@ -0,0 +1,156 @@
...
+  std::cout << "\n\nRAJA prefix sum (exclusive_scan) example using a series of " << N << " length...\n";
+
+  std::vector<int> input(N, 1);
+  std::vector<int> reference_output(N, 0);
+  std::vector<int> test_output(N, 0);
+
+  //----------------------------------------------------------------------------
+  std::cout << "\n Running C-style prefix sum...\n";
+
+  auto start = std::chrono::high_resolution_clock::now();
```
artv3, line 67: There is a RAJA timer that might be useful here. See example: 
https://github.com/LLNL/RAJA/blob/develop/exercises/permuted-layout-batch-matrix-multiply_solution.cpp#L105

examples/prefix_sum.cpp
```diff
@@ -0,0 +1,156 @@
...
+  duration = std::chrono::duration_cast<std::chrono::microseconds>(end - start);
+  
+  std::cout << "   Result: " 
+            << (check_equal(reference_output, test_output) ? "PASS" : "FAIL")
+            << " | Time: " << duration.count() << " us\n";
+#endif
+  
+
+  //----------------------------------------------------------------------------
+#if defined(RAJA_ENABLE_CUDA)
```
artv3, line 119: Maybe add HIP variant as well?

SavannaSpayd: Unfortunately, I don't have a HIP-compatible environment. Since I wouldn’t be able to verify the results or debug any issues, I’ve opted to leave it out to avoid introducing untested code.

SavannaSpayd:
@artv3 Thanks for your feedback. I've implemented the requested change. Ready for another review.

PR #1832, opened 2025-04-28
Feature: Device messages

bechols97:
## Summary

This PR adds a feature for device messages (i.e. store function arguments on device to be handled by a host callback at a later time). The original idea for the feature is to have better error handling on the device with handling any output on with a host callback. This moves the code from Camp to RAJA due dependency on atomic operations.

## Design review

For the design, there are some open questions. (regarding these open questions please see the `Design notes` below)
1. Currently, this design supports a MPSC model with the expectation that the device side will produce messages while a single host thread consumes messages. 
2. Additionally, before consuming messages, the current implementation forces the stream to synchronize. 
3. If there are more messages than the buffer size, then those messages are lost. This is to avoid waits on the device side. However, would it be beneficial to have a configurable option to use a circular buffer?

### Design notes

Based on discussion from a meeting:
- As suggested in the meeting, there should be policies for the message queue to support different use cases and allows the message queue to be extended if there are any future use cases. 
- Based on current use cases, the default policy of the message queue should support a MPSC model, wait all should synchronize the stream, and not storing additional messages when the buffer is full. 
- The `message_handler` class that stores the callback should be move-only to prevent accidental copies to a lambda with supporting a view-like queue to copy to device kernels (as mentioned below by @MrBurmark). 

### Additional design notes

Based on further discussions from a meeting:
- [x] Ability to get range of messages (so that user can sort)
- [x] Ability to unsubscribe a callback
- [ ] ~~Ability to bind args for a callback, such as source location or strings.~~
  - If moving to C++20, `std::bind_front` can be used. However, if needed, this can be added in a future PR
- [ ] ~~Store number of missed messages~~
  - Posting messages returns a bool. Users can use this to get number of missed messages.
- [ ] ~~Remove duplicate messages~~
  - This would likely be a separate queue policy, which can be added in a future PR
- [ ] ~~Move interface to something similar to the Reducer interface, where kernel/forall loop gives correct resource.~~
  - For now, a `message_manager` will just have one resource. If need be in the future, this can be expanded to support multiple resources


test/unit/messages/test-messages.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+#include "gtest/gtest.h"
+
+TEST(message_handler, initialize) {
+  int test = 0;
+  RAJA::message_handler<void(int)> msg(1, [&](int val) {
+    test = val;   
+  });
+
+  ASSERT_EQ(msg.test_any(), false);
```
MrBurmark, line 19: You might also assert that test is 0 here.

MrBurmark:
I think the design would benefit from a view like object that can be used in device kernels.

artv3:
@bechols97 , can you add an example in the examples folder? I have a use case where this may be handy. In my application if I a thread gets a negative value, I  want to take note and output it at the end of the kernel by the root rank. Currently I am using printf and every thread that encounters the negative value spews information onto the screen.

rcarson3:
@bechols97 so one of the libraries I maintain has a failure macro that on the host side throw an error with some useful error messages associated with it which can provide useful context for users / developers for why something failed. Do you think we could emulate something like with this framework if we could provide the absolute max size of the char array / string literal that we'd like and then at the error site have that passed into your message class?

MrBurmark:
I think it would be possible to add a fixed capacity string-like object that could be passed through this interface.

bechols97:
@rcarson3 Yes, the original intention behind this idea is to be used as a device side error handler (though left to be more generic in case there are other use cases). As @MrBurmark mentioned, it would be possible to use a fixed-string object with the message handler. For string literals, you would want some type of fixed-string object / char array to store the string that is later passed to a host side callback. 

In addition to the fixed-string object, any type that is trivially destructible should work as well.

MrBurmark:
The current state looks good if you're trying to handle a single kind of error.
```
auto logger = RAJA::message_handler<void(int*, int, int)>(num_messages, host, 
  [](int* ptr, int idx, int value) {
    std::cout << "\n pointer " << ptr << " a[" << idx << "] = " << value << "\n";
  }
);

auto cpu_msg_queue = logger.get_queue<RAJA::mpsc_queue>();
RAJA::forall<RAJA::seq_exec>(host, RAJA::RangeSegment(0, N), [=] (int i) { 
  if (a[i] < 0) { 
    cpu_msg_queue.try_post_message(a, i, a[i]); 
  }
});

logger.wait_all();
```

MrBurmark:
Another use case that I'm considering is having a long lived logger with an allocation. Then I can enqueue multiple types of messages in that while keeping the gpu running and check for messages occasionally to avoid extra synchronizes.
I am not sure this is feasible however as most of our logging use cases involve catching an error and stopping. If we did continue running we would likely encounter a hard error like a seg fault later.

examples/messages-forall.cpp
```diff
@@ -0,0 +1,461 @@
...
+ *  RAJA features shown:
+ *    - `forall` loop iteration with messages
+ *    -  Create message handler with Resource argument
+ *
+ */
+
+// This is a simplified example fixed string to show how
+// custom types can be used with the message queue.
+template <std::size_t N>
+class my_string
```
rcarson3, line 38: Thanks for adding an example with a "string-like" type in it :) It's helps provide me a road map for how I could replace my current Warning / Error reporting on the GPUs with this framework.

artv3:
@bechols97 , can we bring this up to date?

include/RAJA/policy/msg_queue/spsc_queue.hpp
```diff
@@ -0,0 +1,109 @@
...
+ */
+
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+// Copyright (c) 2016-25, Lawrence Livermore National Security, LLC
+// and RAJA project contributors. See the RAJA/LICENSE file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#ifndef RAJA_mpsc_queue_HPP
```
gberg617:
```suggestion
#ifndef RAJA_spsc_queue_HPP
```

gberg617: Same for line 20 below and for line 6.

include/RAJA/policy/msg_queue/spsc_queue.hpp
```diff
@@ -0,0 +1,109 @@
+/*!
+ ******************************************************************************
+ *
+ * \file
+ *
+ * \brief   Header file containing implementation for a MPSC
```
gberg617:
```suggestion
 * \brief   Header file containing implementation for a SPSC
```
Would it be worth defining what these acronyms (MPSC/SPSC) mean somewhere?

include/RAJA/policy/msg_queue/mpsc_queue.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+  /// Posts message to queue. This is marked `const` to pass to lambda by
+  /// copy. This throws away messages that are over the capacity of the
+  /// container.
+  template<typename... Ts>
+  RAJA_HOST_DEVICE bool try_post_message(Ts&&... args) const
+  {
+    if (m_container != nullptr)
+    {
+      auto local_size = RAJA::atomicInc<auto_atomic>(&(m_container->m_size));
```
gberg617: Is the only difference between MPSC and SPSC the atomic add? I wonder if there's a way to combine these somehow to reduce code duplication.

MrBurmark: I believe if you switch to the seq_atomic policy it will just do an add.

bechols97: Yes, the `seq_atomic` policy will just do an add. My original thought process was that the SPSC policies could have some additional optimizations with storing messages due to knowing that their aren't multiple publishers (though the default policy doesn't show a good example of this). However, I don't know if there would be any use cases for the SPSC policies. If there aren't any use cases at the moment for the SPSC, then these can be removed.

The MPSC policy was looking at using `auto_atomic` to support this with builds that have parallel sections with both GPUs and OpenMP in the same build.

include/RAJA/policy/msg_queue/mpsc_queue.hpp
```diff
@@ -0,0 +1,127 @@
...
+template<typename Container, typename... Args>
+class queue<Container, RAJA::mpsc_queue, RAJA::msg_args<Args...>>
+{
+public:
+  using policy = RAJA::mpsc_queue;
+
+  using args_type = camp::tuple<Args...>;
+  using size_type = typename Container::size_type;
+
+  queue(int id, Container& container) : m_id {id}, m_container {&container} {}
```
MrBurmark: It would be nice if the user didn't have to supply the id.

include/RAJA/util/messages.hpp
```diff
@@ -0,0 +1,332 @@
...
+    using size_type      = unsigned long long;
+    using pointer        = value_type*;
+    using const_pointer  = const value_type*;
+    using iterator       = value_type*;
+    using const_iterator = const value_type*;
+
+    size_type m_begin {0};
+    size_type m_end {0};
+    size_type m_capacity {0};
+    size_type m_size {0};
```
MrBurmark: Do you need size if you have begin and end?

include/RAJA/util/messages.hpp
```diff
@@ -0,0 +1,332 @@
...
+  using value_type     = char;
+  using size_type      = unsigned long long;
+  using pointer        = value_type*;
+  using const_pointer  = const value_type*;
+  using iterator       = msg_iterator;
+  using const_iterator = const iterator;
+  using resource_type  = resource_deleter::resource_type;
+
+  message_bus()
+      : m_res {camp::resources::Host()},
```
MrBurmark: Do we want a default constructor? If so it should probably delegate to the other constructor.

include/RAJA/util/messages.hpp
```diff
@@ -0,0 +1,332 @@
...
+private:
+  // TODO: create small wrapper for callables
+  template<typename Policy, typename Callable, typename R, typename... Args>
+  auto get_queue_impl(msg_id id, Callable&& c, std::function<R(Args...)>)
+  {
+    m_callbacks[id] = callback_type {[=](char* msg_args_buf) {
+      msg_args<msg_decay_t<Args>...>& aligned_args = *std::launder(
+          reinterpret_cast<msg_args<msg_decay_t<Args>...>*>(msg_args_buf));
+      camp::apply(c, aligned_args);
+      aligned_args.~msg_args<msg_decay_t<Args>...>();
```
MrBurmark: Is the header destroyed somewhere?

include/RAJA/policy/msg_queue/mpsc_queue.hpp
MrBurmark: SPSC will probably be the most used as its what the normal cpu only build would use.

include/RAJA/policy/msg_queue/mpsc_queue.hpp
```diff
@@ -0,0 +1,137 @@
...
+        new (buf) msg_header {args_sz, m_id, buf + header_sz};
+        new (buf + header_sz)
+            msg_args<Args...> {args_type(std::forward<Ts>(args)...)};
+
+        return true;
+      }
+
+      // If message is not stored, update end to have the correct number of
+      // bytes and/or allow for additional smaller messages to be stored
+      RAJA::atomicSub<auto_atomic>(&(m_container->m_end), msg_sz);
```
MrBurmark: There is a question here of if you ever want to allow end to be past capacity. There is the possibility of overflow to consider. There is the possibility that small messages will never actually see that there is unused capacity because of the ordering. It would be easier to reason about this if this used a CAS loop style atomic as then end would never go past capacity as updates would only ever be applied if the message would fit.

include/RAJA/policy/msg_queue/mpsc_queue.hpp
bechols97: While users can use this class directly, the preferred way would be for uses to keep the queue type returned when subscribing to a new message type. This will set the id automatically.

include/RAJA/util/messages.hpp
bechols97: Size is now removed

include/RAJA/policy/msg_queue/mpsc_queue.hpp
bechols97: This has been updated to use a CAS atomic loop for the updating the end size of the buffer.

examples/messages-forall.cpp
```diff
@@ -0,0 +1,464 @@
...
+  checkResult(c, N);
+  logger.wait_all();
+
+//----------------------------------------------------------------------------//
+// RAJA::sind_exec policy enforces simd execution.... 
+//----------------------------------------------------------------------------//
+
+  std::cout << "\n Running RAJA simd_exec vector addition...\n";
+
+  RAJA::forall<RAJA::simd_exec>(host, RAJA::RangeSegment(0, N), [=] (int i) {
```
MrBurmark: Probably note that its not compatible with simd, if so.

include/RAJA/pattern/messages/msg_header.hpp
```diff
@@ -0,0 +1,52 @@
...
+ *  however, the device side can get a different value compared to the
+ *  host.
+ *
+ * @return The size with proper alignment
+ */
+{
+  return ((size + alignment - 1) / alignment) * alignment;
+}
+
+struct msg_header
```
MrBurmark, line 29: Instead of util you probably want to put the generic msg stuff in a directory in pattern.

examples/messages-forall.cpp
bechols97: Since this feature is not compatible with simd, I just removed the loop from this example to avoid confusion.

include/RAJA/pattern/messages/msg_header.hpp
bechols97: The msg headers have been moved to the pattern directory (the specific queue policies are still under the policy directory).

examples/messages-forall.cpp
```diff
@@ -0,0 +1,452 @@
...
+  checkResult(c, N);
+  logger.wait_all();
+
+//----------------------------------------------------------------------------//
+// RAJA::omp_parallel_for_static_exec policy execution.... 
+//----------------------------------------------------------------------------//
+
+  std::cout << "\n Running RAJA omp_parallel_for_static_exec (default chunksize) vector addition...\n";
+
+  RAJA::forall<RAJA::omp_parallel_for_static_exec< >>(host, RAJA::RangeSegment(0, N),
```
artv3: I think you may have extra `< >` in the policy?

artv3: oh I see, it has a default chunk size that you can modify: 
https://github.com/llnl/RAJA/blob/1d2ed09f2345497b41c4a13d8116c0244bddc38a/include/RAJA/policy/openmp/policy.hpp#L242-L244

examples/messages-forall.cpp
```diff
@@ -0,0 +1,452 @@
...
+  RAJA::resources::Cuda res_gpu2;
+  using EXEC_POLICY = RAJA::cuda_exec_async<GPU_BLOCK_SIZE>;
+#elif defined(RAJA_ENABLE_HIP)
+  RAJA::resources::Hip res_gpu1;
+  RAJA::resources::Hip res_gpu2;
+  using EXEC_POLICY = RAJA::hip_exec_async<GPU_BLOCK_SIZE>;
+#elif defined(RAJA_ENABLE_SYCL)
+  RAJA::resources::Sycl res_gpu1;
+  RAJA::resources::Sycl res_gpu2;
+  using EXEC_POLICY = RAJA::sycl_exec<GPU_BLOCK_SIZE>;
```
artv3: I think you can simplify this by templating the resource on policy type. See: 
https://github.com/llnl/RAJA/blob/1d2ed09f2345497b41c4a13d8116c0244bddc38a/examples/forall-jit.cpp#L60

include/RAJA/pattern/messages/msg_header.hpp
```diff
@@ -0,0 +1,55 @@
...
+#ifndef RAJA_MSG_HEADER_HPP
+#define RAJA_MSG_HEADER_HPP
+
+#include <cstddef>
+#include "camp/tuple.hpp"
+
+namespace RAJA
+{
+RAJA_HOST_DEVICE
+constexpr std::size_t align(std::size_t size, std::size_t alignment = 16)
```
artv3: Should this go under utils? seems pretty useful beyond the MSG feature

bechols97: If preferred, I can move the `align` function to a separate header under the utils folder. Is there a particular header that this should belong in that currently exists under utils or just create a new header for it?

MrBurmark: You might want to take a look around the utils dir and see if there is a header that this kind of fits with already.

rhornung67: @bechols97 please take a look in here https://github.com/llnl/RAJA/blob/develop/include/RAJA/util/align.hpp#L1

examples/messages-forall.cpp
bechols97: In the purpose of this example, the hope was to show that this feature is able to work with non-default resources as well as work with multiple GPU resources. However, there are two examples showing the multiple non-default resources. If preferred, I can update the first example to only use the default stream (since most use cases will likely be using RAJA's default stream)?

examples/messages-forall.cpp
bechols97: The chunk size doesn't really matter in this example, So, if you prefer having an explicit value in the `< >` to be more clear, then I don't mind updating this.

include/RAJA/pattern/messages/msg_callback.hpp
```diff
@@ -0,0 +1,114 @@
...
+    camp::apply(m_callable, msg.args);
+    msg.~msg_args<std::decay_t<Args>...>();
+  }
+
+private:
+  Callable m_callable;
+};
+
+template<typename Fn>
+struct get_signature;
```
johnbowen42: We should move these into a generic `function_signature_helper.hpp` header.  See https://github.com/llnl/RAJA/pull/1949/changes#diff-72533564b1cbd49c320bfd7981489ca5e5a08143353955c70c3ef535e38fc4ccR56.  Arturo recently added similar methods for deducing the index of template parameters.  These types of metaprogramming utilities are broadly useful and should live in a more visible header

johnbowen42: It would be nice to move both to a generic header, maybe in `internal`.  I think it's good to consolidate stuff like this so we don't end up re-implementing the methods when we need it.  For example we have type_trait helper headers like https://github.com/llnl/RAJA/blob/develop/include/RAJA/pattern/kernel/TypeTraits.hpp

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,422 @@
...
+  template<typename Resource>
+  message_bus(const size_type bus_sz, Resource res) : message_bus {res}
+  {
+    reserve(bus_sz);
+  }
+
+  ~message_bus() { reset(); }
+
+  // Copy ctor/operator
+  message_bus(const message_bus&)            = delete;
```
johnbowen42, line 172: I think we might also need `message_bus
(message_bus)            = delete;`

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,422 @@
...
+///
+/// Provides a way to handle messages from a GPU. This currently
+/// stores messages from the GPU and then calls a callback
+/// function from the host.
+///
+/// Note:
+/// Currently, this forces a synchronize prior to calling
+/// the callback function or testing if there are any messages.
+///
+class message_manager
```
johnbowen42, line 253: Do we want this case for the new classes @llnl/raja-core ?  `kernel` is all the `MessageManager` naming convention for classes

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,422 @@
...
+    m_callback_map.emplace_back();
+
+    return get_queue_impl<Policy>(
+        id, RAJA::msg_callback {std::forward<Callable>(c)});
+  }
+
+  template<typename Callable>
+  void subscribe(msg_id id, Callable&& c)
+  {
+    auto callback = RAJA::msg_callback {std::forward<Callable>(c)};
```
johnbowen42: it might be better to use an explicit constructor here

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,422 @@
...
+          std::make_unique<msg_callback_t>(std::move(callback)));
+    }
+  }
+
+  template<typename Callable>
+  void unsubscribe(msg_id id, Callable&&)
+  {
+    auto& fn_list = m_callback_map.at(id);
+    auto it = std::find_if(fn_list.begin(), fn_list.end(), [](const auto& fn) {
+      return typeid(Callable).hash_code() == fn->hash();
```
johnbowen42: I don't know if this is how we want to be hashing values.  For one, I think it's possible to have hashing collisions here, to `find_if` could just match the first of several possible matches. I think it might be better to hash together the `msg_id` with `std::type_index`, and just make `m_callback_map` a `std::unordered_map<HashCode, vector<msg_callback_t>>`

include/RAJA/policy/msg_queue/mpsc_queue.hpp
```diff
@@ -0,0 +1,153 @@
...
+      // to reserve space. Otherwise, message doesn't fit and no space is
+      // reserved. In other words, the CAS-loop below performs the follwing
+      // operation:
+      // (*address + msg_sz <= capacity) ?  (*address + msg_sz) : *address;
+      do
+      {
+        size_type new_sz = old_sz + msg_sz;
+        local_sz         = old_sz;  // offset to start of message
+        // Checks if fits in queue
+        if (new_sz <= m_container->m_capacity)
```
johnbowen42, line 82: do we ever resize the `m_container`?  this could be a race condition if so

bechols97: The `mpsc_queue` and `spsc_queue` are view-like containers to the owning version of `message_bus`. All resizing type of operations on `message_bus` will end up forcing the resource in `message_bus` to synchronize prior to resizing along with ignoring any messages that are currently stored.

include/RAJA/pattern/messages/msg_callback.hpp
bechols97: Is there a preferred location for these types of metaprogramming utitlies? Would https://github.com/llnl/RAJA/tree/develop/include/RAJA/util be a good place for the more generic header?

johnbowen42: I think so yes, see for example https://github.com/llnl/RAJA/blob/develop/include/RAJA/util/EnableIf.hpp is already there.  I would try to move Arturo's helpers from [github.com/llnl/RAJA/pull/1949/changes#diff-72533564b1cbd49c320bfd7981489ca5e5a08143353955c70c3ef535e38fc4ccR56](https://github.com/llnl/RAJA/pull/1949/changes#diff-72533564b1cbd49c320bfd7981489ca5e5a08143353955c70c3ef535e38fc4ccR56.) there as well, and name it `FunctionSignatureUtil.hpp` or something

examples/messages-forall.cpp
```diff
@@ -0,0 +1,452 @@
...
+#include <cstring>
+#include <iostream>
+
+#include "memoryManager.hpp"
+
+#include "RAJA/RAJA.hpp"
+#include "RAJA/util/resource.hpp"
+
+/*
+ *  Vector Addition Example
```
artv3: Maybe we can rename it to "RAJA::messages example" and update the description

bechols97: I updated the example to simplify and better showcase the different features. Let me know if anything else needs adjusting.

include/RAJA/pattern/messages/msg_callback.hpp
bechols97: This now updated and tried to de-duplicate similar code from Arturo's helpers.

docs/sphinx/user_guide/feature/messages.rst
```diff
@@ -0,0 +1,167 @@
...
+
+===============================
+Messages
+===============================
+
+RAJA provides a portable interface and type-safe way to store function arguments that
+passed to a function at a later time. For example, from a GPU, arguments can be stored
+and passed to function that prints to a file on the CPU. 
+
+.. warning:: This capability is new and should be considered experimental.
```
artv3: We have had feedback on how we advertise features as experimental and I would like share some thoughts: 
If this feature is meant for power users, I recommend we place under the expt namespace same if we expect API breaking changes in the near term. If this is pretty robust and we don't expect any API changes then maybe we don't need to advertise as experimental.  I share this because code projects have experienced development slow downs when updating to a RAJA version that introduces an API breaking change for a TPL. At any rate, maybe also expanding on considerations when integrating this feature into a project could mean. Even moving something from `RAJA::expt::` to standard `RAJA::` is a breaking change that users aren't a fan of.

examples/messages-forall.cpp
```diff
@@ -0,0 +1,393 @@
...
+// Functions for checking and printing results
+//
+void checkResult(int* res, int len); 
+void printResult(int* res, int len);
+
+
+int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
+{
+
+  std::cout << "\n\nRAJA vector addition example...\n";
```
artv3:
```suggestion
  std::cout << "\n\nRAJA RAJA::messages with vector addition example...\n";
```

examples/messages-forall.cpp
```diff
@@ -0,0 +1,393 @@
...
+    };
+  msg_manager.subscribe(cpu_msg_queue.get_id(), err_callback);
+// _raja_msg_subscribe_end
+
+
+// _raja_msg_unsubscribe_start
+  msg_manager.unsubscribe(cpu_msg_queue.get_id(), err_callback);
+// _raja_msg_unsubscribe_end
+  
+  const int N = 100000;
```
artv3:
```suggestion
  constexpr int N = 100000;
```

examples/messages-forall.cpp
```diff
@@ -0,0 +1,393 @@
...
+//
+// Clean up.
+//
+  host.deallocate(a);
+  host.deallocate(b);
+  host.deallocate(c);
+
+  host.deallocate(a_);
+  host.deallocate(b_);
+  host.deallocate(c_);
```
artv3: minor nitpicking, I think we should call this `res_host`  for consistency with the above example. It also helps give people give context.

test/unit/messages/test-messages.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+  msg_manager.wait_all();
+
+  ASSERT_EQ(test[0], 1);
+  ASSERT_EQ(test[1], 2);
+  ASSERT_EQ(test[2], 3);
+}
+
+TEST(message_handler, wait_all_overflow) {
+  // TODO: implement
```
artv3: Will you add this in this PR?

test/unit/messages/test-messages.cpp
```diff
@@ -0,0 +1,303 @@
...
+  });
+
+  ASSERT_EQ(q.try_post_message(5), true);
+  ASSERT_EQ(q.try_post_message(7), false);
+
+  ASSERT_EQ(test, 0);
+} 
+
+TEST(message_handler, try_post_message_overwrite) {
+  // TODO: implement
```
artv3: Another here

test/unit/messages/test-messages.cpp
bechols97: These were originally test cases for another queue policy. For the purposes of this PR, this won't be added back. If desired later, these extra policies can be added in a future PR without an API change. I will remove these.

artv3: Sounds good, thanks. Maybe future work can be captured in a GitHub issue

include/RAJA/policy/msg_queue/mpsc_queue.hpp
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+private:
+  std::size_t m_id;
+  Container* m_container;
+};
+
+// TODO: turning off for now
+// need to relook at logic
+#if 0
```
artv3, line 119: I saw this in a couple of places, it might be worth also documenting why this didn't get used and how you intend to use it. Also adding a date.

artv3 (approved):
Just need to get the CI passing, but super excited for this contribution!

docs/sphinx/user_guide/feature/messages.rst
```diff
@@ -0,0 +1,184 @@
...
+.. ## SPDX-License-Identifier: (BSD-3-Clause)
+.. ##
+
+.. _feat-message-label:
+
+===============================
+Messages
+===============================
+
+RAJA provides a portable interface and type-safe way to store function arguments that
```
rhornung67:
```suggestion
RAJA provides a portable interface and type-safe way to store function arguments that are
```

docs/sphinx/user_guide/feature/messages.rst
```diff
@@ -0,0 +1,184 @@
...
+passed to a function at a later time. For example, from a GPU, arguments can be stored
+and passed to function that prints to a file on the CPU. 
+
+.. warning:: This capability is new and should be considered experimental.
```
rhornung67: Since we decided to not put it in the experimental namespace, I would suggest changing this comment to something along the lines of `This capability is new. We would like users to try it out and give us feedback to improve it.`

docs/sphinx/user_guide/feature/messages.rst
```diff
@@ -0,0 +1,184 @@
...
+
+To create the ``message_manager``: 
+
+.. literalinclude:: ../../../../examples/messages-forall.cpp
+    :start-after: _raja_msg_manager_start
+    :end-before: _raja_msg_manager_end
+    :language: C++
+
+``buf_sz`` is the size of the buffer that stores messages. ``res_host`` is the resource of the 
+execution policy, which determines what type of memory to store messages with. For GPU resources,
```
rhornung67:
```suggestion
execution policy, which determines the memory space in which messages will be stored. For GPU resources, for example,
```

docs/sphinx/user_guide/feature/messages.rst
```diff
@@ -0,0 +1,184 @@
...
+Here is how the ``msg_queue`` can be used to publish messages:
+
+.. literalinclude:: ../../../../examples/messages-forall.cpp
+    :start-after: _raja_msg_k2_start
+    :end-before: _raja_msg_k2_end
+    :language: C++
+
+Handling messages
+^^^^^^^^^^^^^^^^^
+Lastly, there needs to be a way to "handle" messages. In others, direct the messages to the corresponding callback(s).
```
rhornung67:
```suggestion
Lastly, there needs to be a way to direct messages to the corresponding callback(s).
```
I am confused by this. Is this what you mean?

docs/sphinx/user_guide/feature/messages.rst
```diff
@@ -0,0 +1,184 @@
...
+  
+This example will show how callbacks can be subscribed to various types of messages as
+well as how to publish messages on multiple different platforms. For the purposes of 
+this example, output for messages will be printed using ``std::cout``.
+
+--------------------------
+Application considerations
+--------------------------
+
+Things to consider when using ``RAJA::messages`` in an application.
```
rhornung67:
```suggestion
There are several things to consider when using ``RAJA::messages`` in an application.
```

docs/sphinx/user_guide/feature/messages.rst
bechols97: Yes, that was what I meant. I had some typos and missing words in there that made it confusing.

examples/messages-forall.cpp
```diff
@@ -0,0 +1,390 @@
...
+ *
+ *  The purpose of this example to show how RAJA::messages can be used.
+ *  For this example, the kernels will be launched using RAJA::forall 
+ *  along with a vector addition calculation. This will show how messages
+ *  can be used:
+ *  - to store messages on various execution policies (serial, OpenMP, GPU)
+ *  - to print to a file on the GPU using the RAJA::messages
+ *  - to interact with multiple GPU streams
+ *  - to create custom types that can be stored (note: these not to be
+ *    trivially destructible and copyable.
```
rhornung67:
```suggestion
 *  - to create custom types that can be stored (note: these not to be
 *    trivially destructible and copyable)
```
I don't understand `these not to be`. What do you mean

examples/messages-forall.cpp
```diff
@@ -0,0 +1,390 @@
...
+ *
+ */
+
+// This is a simplified example fixed string to show how
```
rhornung67:
```suggestion
// This is a simplified example fixed length string to show how
```

rhornung67 (changes requested):
Seems like a solid first draft. 

I have several small requests for documentation changes.

examples/messages-forall.cpp
bechols97: Looks like typo in the comment. I meant to say that "these types should be trivially destructible and trivially copyable". This should now be updated.

rhornung67 (approved):

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,422 @@
...
+    using msg_callback_t = msg_callback<Callable, R(Args...)>;
+
+    m_callback_map[id].emplace_back(
+        std::make_unique<msg_callback_t>(std::move(c)));
+
+    return m_bus.template get_queue<Policy, std::decay_t<Args>...>(id);
+  }
+
+  msg_bus m_bus;
+  std::vector<msg_fn_list_t> m_callback_map;
```
johnbowen42: @bechols97 and I discussed offline that we should replace this with either (a) `std::unordered_map<int, std::unordered_map<decltype(std::hash<Callable>), std::vector<msg_callback_t>>>` or (b) implement a hash combiner in a common header and `std::unordered_map<CombinedHashT, std::vector<msg_callback_t>>`

johnbowen42: Even though `msg_id` is dense initially (assigned in ascending order 1,2,3), I think that if a single manager is used across many kernel launches this could lead to a memory leak, and we should probably `erase` the `msg_id` storage entry after its used

bechols97: Updated to something similar to option (a) with the message header storing an additional `std::size_t` for the hash of the `Callable`.

Also, yes that style (assuming there is a new message type for each kernel launch) would slowly consume more memory and would essentially be a memory leak. I updated the `unsubscribe_all(msg_id)` to `erase` the entry instead of just removing the callback list. However, this still requires users/applications to call either `unsubscribe_all(msg_id)` or `unsubscribe_all()` directly to remove an entry or all entries. This could likely be addressed in a future PR with an additional queue policy that deletes the entry once the queue goes out of scope.

MrBurmark:
Regarding the comment about moving to something more like the reducer interface, we do have users of RAJA who capture the same object in the lambda but then select a policy, for example seq, omp, or cuda/hip, at runtime. I want to circle back on how this supports that use case.

bechols97:
> Regarding the comment about moving to something more like the reducer interface, we do have users of RAJA who capture the same object in the lambda but then select a policy, for example seq, omp, or cuda/hip, at runtime. I want to circle back on how this supports that use case.

I should clarify that the reducer interface comment was about being more similar to the new reducer interface. 

That being said the `mpsc_queue` policy supports storing messages on either the host or device using `RAJA::auto_atomic`. This should perform the correct atomic operation for the host or device. The one exception is if RAJA is built with omp support and the use is selecting a seq policy, then this would perform an extra atomic operation in the seq case. (Note: if both a cuda/hip kernel is used to run in the background while a seq/omp kernel is running at the same time using the same buffer, we would need to use the system atomic operations instead. This would be just a simple addition with another queue policy if this use case is needed). 

Regarding what type of memory is allocated for the buffer and any synchronization before reading the messages, this depends on the Resource provided during the creation of `msg_manager`. For the use case you mentioned, users would need to provide the resource for the behavior that is needed on that application (i.e., Resource::Host for seq, Resource::Cuda for cuda, etc). If these policies needed to be mixed, the buffer allocates pinned memory for the cuda/hip resources and would be accessible by the host though has the chance of having an extra synchronization.

MrBurmark:
Right, I suppose we can use the spsc policy to avoid using host atomics if the only option is a sequential policy?
How does it know to allocate pinned vs another memory? Does it support general allocators?

bechols97:
Yes, the `spsc_queue` policy can be used to always avoid atomics. If OpenMP support is not configured, then the `spsc_queue` and `mpsc_queue` on the host will do the same thing.

The message manager will always try the allocate pinned memory from whichever reousrce is passed in. For Cuda/Hip resources, this will be pinned memory. For the Host resource, the Camp resource fallbacks to just `malloc`. This currently relies on the resource to make allocations, so there isn't any support for general allocators.

MrBurmark:
Hmm, for more flexibility it would be good to have allocator support like `std::vector` or `RAJA::WorkGroup`. Have we discussed that?

bechols97:
Hmm, as much as I can remember, we may have been briefly discuss using an allocator. This would be for the msg_bus (buffer to store messages), correct? Also, did you have a preference in terms of having a default allocator, like `std::allocator`, or always have the type be provided by the user?

MrBurmark:
The current behavior seems reasonable as a default, so I wouldn't be opposed to something like that. We do have a `ResourceAllocator` in our test includes for the WorkGroup tests. You could pull that into our public util directory so people could use that. If you do that it would probably make sense to add a constructor from a resource as it always default constructs one now, in fact the default behavior should probably to get the default resource. Also while pinned makes sense in many contexts it would be nice to let people choose what MemoryAccess they want in addition to the resource type.

docs/sphinx/user_guide/feature/messages.rst
```diff
@@ -0,0 +1,184 @@
...
+ ======================= ============================
+ Message queue Policies  Brief description
+ ======================= ============================
+ spsc                    Supports a single producer,
+                         single consumer; i.e., no
+                         atomic operations 
+ mpsc                    Supports multiple producers,
+                         a single consumer; i.e., 
+                         requires atomic operations. 
+                         Automiatically determines
```
MrBurmark:
```suggestion
                         Automatically determines 
```

examples/messages-forall.cpp
```diff
@@ -0,0 +1,392 @@
...
+#include "RAJA/RAJA.hpp"
+#include "RAJA/util/resource.hpp"
+
+/*
+ *  RAJA::messages example
+ *
+ *  The purpose of this example to show how RAJA::messages can be used.
+ *  For this example, the kernels will be launched using RAJA::forall 
+ *  along with a vector addition calculation. This will show how messages
+ *  can be used:
```
MrBurmark:
```suggestion
 *  The purpose of this example to show how RAJA::messages can be used.
 *  For this example, the kernels will be launched using RAJA::forall 
 *  along with a vector addition calculation. Messages can be used:
```
Would something like this make more sense?

include/RAJA/pattern/messages/msg_callback.hpp
```diff
@@ -0,0 +1,86 @@
...
+public:
+  using return_t = Ret;
+
+  template<typename Object>
+  explicit msg_callback(const Object& obj) : m_callable {obj}
+  {}
+
+  template<typename Object>
+  explicit msg_callback(Object&& obj) : m_callable {std::move(obj)}
+  {}
```
MrBurmark: Did you mean to use Callable, or did you want to template on Object?

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,444 @@
...
+  using queue_allocator = typename resource_deleter::allocator_type;
+
+  message_bus() : message_bus(camp::resources::Host()) {}
+
+  template<typename Resource>
+  message_bus(Resource res, Allocator alloc = Allocator {})
+      : m_res {res},
+        m_alloc {alloc},
+        m_bus {new(queue_allocator(alloc).allocate(1)) queue {},
+               resource_deleter {m_res, alloc}}
```
MrBurmark: The deleter also gets a copy of the resource and allocator?

bechols97: The resource is needed to synchronize before the deleter can deallocate the memory buffer for the queue, and a custom allocator from a user is not guaranteed to synchronize.

include/RAJA/pattern/messages/msg_callback.hpp
bechols97: Yeah, I meant Callable here. I will update this, thanks

include/RAJA/pattern/messages/msg_manager.hpp
MrBurmark: I also don't see anything about checking whether or not the allocator supports propagation on copy. See https://en.cppreference.com/cpp/named_req/Allocator where it talks about things like what equality means for allocators and how things propagate.

MrBurmark: Sure, but if those exist in the resource deleter does it make sense to use those instead of having two copies?

include/RAJA/pattern/messages/msg_manager.hpp
MrBurmark: Generally we have been using CamelCase for classes.

bechols97: Updated to use CamelCase

examples/messages-forall.cpp
bechols97: I updated this description to make more sense

include/RAJA/pattern/messages/msg_callback.hpp
```diff
@@ -0,0 +1,83 @@
...
+{
+public:
+  using return_t = Ret;
+
+  explicit MsgCallback(const Callable& callable) : m_callable {callable} {}
+
+  explicit MsgCallback(Callable&& callable) : m_callable {std::move(callable)}
+  {}
+
+  std::type_index get_type() const final override { return typeid(Callable); }
```
MrBurmark: final implies override
```suggestion
  std::type_index get_type() const final { return typeid(Callable); }
```

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,483 @@
...
+    bool operator==(const MsgIterator& other) const
+    {
+      return (cur_ptr == other.cur_ptr);
+    }
+
+    bool operator!=(const MsgIterator& other) const
+    {
+      return !(*this == other);
+    }
+
```
MrBurmark: Not needed in c++20
```suggestion
```

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,483 @@
...
+  using allocator_type = Allocator;
+  // Allocator for queue struct
+  using queue_allocator = typename ResourceDeleter::allocator_type;
+
+  MessageBus() : MessageBus(camp::resources::Host()) {}
+
+  template<typename Resource>
+  MessageBus(Resource res, Allocator alloc = Allocator {})
+      : m_bus {new(queue_allocator(alloc).allocate(1)) Queue {},
+               ResourceDeleter {res, alloc}}
```
MrBurmark: Its best to copy/move the allocator, then use the version that's being stored long term to allocate the bus.

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,483 @@
...
+    return get_queue_impl<Policy>(
+        id, RAJA::MsgCallback {std::forward<Callable>(c)});
+  }
+
+  template<typename Callable>
+  void subscribe(msg_id id, Callable&& c)
+  {
+    RAJA::MsgCallback callback {std::forward<Callable>(c)};
+    auto& fn_list = m_callback_map.at(id);
+    auto it = std::find_if(fn_list.begin(), fn_list.end(), [](const auto& fn) {
```
MrBurmark: ranges version?

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,483 @@
...
+   * callbacks. Once messages are handled, then container is cleared.
+   */
+  template<typename Container>
+  void handle_all(Container& messages)
+  {
+    if (!m_callback_map.empty())
+    {
+      for (const auto& msg : messages)
+      {
+        msg_id id = std::make_pair(msg->type, msg->hash);
```
MrBurmark, line 435: Is this a message id or a callback id?

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,483 @@
...
+          (*callback)(msg->args);
+        }
+        msg->~MsgHeader();
+      }
+      messages.clear();
+    }
+    clear();
+  }
+
+  void wait_all()
```
MrBurmark, line 457: Should this call handle_all?

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,483 @@
...
+
+  void unsubscribe_all() { m_callback_map.clear(); }
+
+  void clear() { m_bus.clear_messages(); }
+
+  bool test_any() { return m_bus.has_pending_messages(); }
+
+  auto get_messages()
+  {
+    std::vector<const MsgHeader*> messages;
```
MrBurmark, line 411: Could this generate the values on the fly instead of storing them?

include/RAJA/policy/msg_queue/mpsc_queue.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+#include "RAJA/pattern/messages/msg_header.hpp"
+#include "RAJA/policy/msg_queue/policy.hpp"
+
+namespace RAJA
+{
+namespace messages
+{
+
+template<typename Container, typename... Args>
```
MrBurmark: Could this be more descriptive than Container?

include/RAJA/policy/msg_queue/mpsc_queue.hpp
```diff
@@ -0,0 +1,117 @@
...
+  using args_type = camp::tuple<Args...>;
+  using size_type = typename Container::size_type;
+
+  Queue(std::pair<std::size_t, std::size_t> id, Container& container)
+      : m_type {id.first},
+        m_hash {id.second},
+        m_container {&container}
+  {}
+
+  Queue(std::pair<std::size_t, std::size_t> id, Container* container)
```
MrBurmark: Can container be nullptr?

include/RAJA/policy/msg_queue/mpsc_queue.hpp
```diff
@@ -0,0 +1,117 @@
...
+      // operation:
+      // (*address + msg_sz <= capacity) ?  (*address + msg_sz) : *address;
+      do
+      {
+        size_type new_sz = old_sz + msg_sz;
+        local_sz         = old_sz;  // offset to start of message
+        // Checks if fits in queue
+        if (new_sz <= m_container->m_capacity)
+        {
+          old_sz = RAJA::atomicCAS<auto_atomic>(&(m_container->m_end), local_sz,
```
MrBurmark: We are in the internals of RAJA, woud it be more clean to use the internal CAS loop implementation that takes lambdas?

include/RAJA/policy/msg_queue/mpsc_queue.hpp
```diff
@@ -0,0 +1,117 @@
...
+                                                new_sz);
+        }
+      } while (local_sz != old_sz);
+
+      if (m_container->m_data != nullptr &&
+          local_sz + msg_sz <= m_container->m_capacity)
+      {
+        char* buf = m_container->m_data + local_sz;
+        new (buf) MsgHeader {args_sz, m_type, m_hash, buf + header_sz};
```
MrBurmark, line 88: Does it make more sense to store the whole id type in the header?

include/RAJA/util/HashCombiner.hpp
```diff
@@ -0,0 +1,68 @@
...
+ * @brief Combines hashes of all types together.
+ * @param values Objects to combine hashes for
+ * @return Returns hash combined of objects
+ *
+ * This will combine the hash values of various objects. The
+ * hash value of each type uses `std::hash`.
+ *
+ * Note: this is a simple hash combiner that can generate
+ * collisons with being useful for combining hashes for
+ * maps and sets.
```
MrBurmark: Is this all you really meant to say?
```suggestion
 * Note: this is a simple hash combiner that can generate
 * collisons.
```

MrBurmark: Do we want to make a generic util given the note above?

include/RAJA/util/ResourceAllocator.hpp
```diff
@@ -0,0 +1,118 @@
...
+#include "RAJA/util/resource.hpp"
+#include "RAJA/util/macros.hpp"
+
+namespace RAJA
+{
+
+namespace detail
+{
+template<typename Resource>
+struct ResourceAllocator
```
MrBurmark: Should camp provide this class.

include/RAJA/util/ResourceAllocator.hpp
```diff
@@ -0,0 +1,118 @@
...
+    {
+      return lhs.get_resource() == rhs.get_resource();
+    }
+
+    template<typename U>
+    friend inline bool operator!=(allocator const& lhs, allocator<U> const& rhs)
+    {
+      return !(lhs == rhs);
+    }
+
```
MrBurmark: Handled by c++20.
```suggestion
```

include/RAJA/util/ResourceAllocator.hpp
```diff
@@ -0,0 +1,118 @@
...
+      return ptr;
+    }
+
+    void deallocate(value_type* ptr, std::size_t) noexcept
+    {
+      m_res.deallocate(ptr, m_mem_type);
+    }
+
+    Resource const& get_resource() const { return m_res; }
```
MrBurmark: How about a non-const accessor as well?

test/unit/CMakeLists.txt
```diff
@@ -19,3 +19,4 @@ add_subdirectory(view-layout)
 add_subdirectory(algorithm)
 add_subdirectory(workgroup)
 add_subdirectory(indexing)
+add_subdirectory(messages)
```
MrBurmark, line 22: How about functional tests with forall to ensure that this works correctly with each backend?

include/RAJA/policy/msg_queue/mpsc_queue.hpp
bechols97: At the moment, if the container is a `nullptr` when the queue is created, then the message bus has gotten into a bad state.

include/RAJA/policy/msg_queue/mpsc_queue.hpp
bechols97: If I understand this comment correctly, this is asking if storing m_type and m_hash as a std::pair in the message header as well would make more sense, correct? The reason I didn't do this is because it require `constexpr` functions to used in `__host__ __device__` functions. This is possible since `std::pair` is `constexpr` but would force users to use `--expt-relaxed-constexpr` with NVCC, which I don't believe is a requirement yet.

include/RAJA/util/HashCombiner.hpp
bechols97: Should I move this to a `detail` or internal namespace then?

include/RAJA/pattern/messages/msg_manager.hpp
bechols97: The purpose of `get_messages` storing messages was to return something that could:
  - Apply some operation to the message list, such as sorting
  - Avoid object lifetime issues when iterating through messages while destroying the headers

It is possible to generate messages on the fly and avoid object lifetime issues with having more specific loops in `wait_all`. However, I couldn't think of a good way to support sorting though I am open to suggestions.

include/RAJA/pattern/messages/msg_manager.hpp
bechols97: This is a message id. The mapping structure is message id -> callback list. To provide a unique hash, a pair of `type_index(Callable)` and the old incrementing hash is created.

include/RAJA/pattern/messages/msg_manager.hpp
bechols97: Good point, this has been updated

include/RAJA/pattern/messages/msg_manager.hpp
bechols97: Updated to use ranges version

include/RAJA/pattern/messages/msg_manager.hpp
bechols97: At one point `wait_all` and `handle_all` did slightly different things. However, yes, this has been updated to use handle_all to deduplicate code.

include/RAJA/policy/msg_queue/mpsc_queue.hpp
bechols97: Updated the name to be more descriptive than Container. Let me know if the new name works better.

include/RAJA/pattern/messages/msg_manager.hpp
MrBurmark: Ok, we'll think about it more if needed.

include/RAJA/policy/msg_queue/mpsc_queue.hpp
MrBurmark: Oh, right, annoying.

include/RAJA/util/HashCombiner.hpp
MrBurmark: Ya anything that is an implementation detail and shouldn't be called by users should be in a namespace like detail.

include/RAJA/util/ResourceAllocator.hpp
bechols97: For traceability, the PR in camp for this functionality here: https://github.com/llnl/camp/pull/207. Once this is merged, I can update the camp submodule and use the camp version instead.

test/unit/CMakeLists.txt
bechols97: Add some functional tests

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,479 @@
...
+{
+  // Internal classes
+public:
+  // Queue is public due to limitation with extended lambdas
+  // in nvcc
+  struct Queue
+  {
+    using value_type     = char;
+    using size_type      = unsigned long long;
+    using pointer        = value_type*;
```
artv3, line 65: what is the rule of thumb for ending the alias with _type?

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,479 @@
...
+    RAJA::MsgCallback callback {std::forward<Callable>(c)};
+    auto& fn_list = m_callback_map.at(id);
+    auto it       = std::ranges::find_if(fn_list, [](const auto& fn) {
+      return std::type_index {typeid(Callable)} == fn->get_type();
+    });
+
+    using msg_callback_t = decltype(callback);
+    if (it != fn_list.end())
+    {
+      // TODO: would it be better to throw or just replace old one?
```
artv3: Do we want to sort this out before merging? or an update?

include/RAJA/pattern/messages/msg_manager.hpp
bechols97: These are just following the naming convention from the C++ standard library.

include/RAJA/pattern/messages/msg_manager.hpp
bechols97: Personally, I am in favor of just replacing old callbacks, which is the current behavior. But, I am open to changing this if there are other opinions.

include/RAJA/pattern/messages/msg_manager.hpp
artv3: ah okay thanks!

include/RAJA/pattern/messages/msg_manager.hpp
artv3: I'm good with that, I suggest removing the TODO if we stick in that direction

rhornung67:
@bechols97 is this ready to merge? IIRC you were going to pull camp main into this when your camp PR https://github.com/llnl/camp/pull/207 is merged, correct?

bechols97:
Yes, once https://github.com/llnl/camp/pull/207 is merged into camp and I will update the camp submodule in this branch, then this branch will be ready to merge. I am adding a few more tests to that camp PR that were requested.

bechols97:
I updated the camp submodule in this PR (along with some fixes due to a missing macro and `wait_for(event)` no longer needing pointers). Also, one of the functional tests for the CUDA build fails and is dependent on the fix in https://github.com/llnl/RAJA/pull/2054 prior to this merging.

include/RAJA/util/ResourceAllocator.hpp
bechols97: Updated this to use the camp `ResourceAllocator`

include/RAJA/pattern/messages/msg_manager.hpp
bechols97: Removed the TODO comment

docs/sphinx/user_guide/feature/messages.rst
```diff
@@ -0,0 +1,195 @@
...
+.. ## to contribute to RAJA.
+.. ##
+.. ## SPDX-License-Identifier: (BSD-3-Clause)
+.. ##
+
+.. _feat-message-label:
+
+===============================
+Messages
+===============================
```
artv3, line 14: It might help to be more direct in the intro here, the opening paragraph mentions messages at the end but I think it could be better to start with the headlining capability.

bechols97: I updated the intro a bit to provide better context for the capability. Please feel free to let me know if the intro or other sections need to be updated.

bechols97 (commented):
@lc-hubcast approve

bechols97 (commented):
@lc-hubcast approve

cmelone:
@bechols97 are you added as a collaborator on the RAJA Gitlab repo? Hubcast can't mirror without those permissions. you'll need to have at least the "developer" role.

(we're working on better error messages in Hubcast so you'll get immediate feedback in the future)

bechols97 (commented):
@lc-hubcast approve

bechols97 (commented):
@lc-hubcast approve

include/RAJA/pattern/messages/msg_manager.hpp
```diff
@@ -0,0 +1,521 @@
...
+template<concepts::Resource Resource, typename Allocator>
+auto make_message_manager(std::size_t bus_sz, Resource r, Allocator alloc)
+{
+  return RAJA::MessageManager<Allocator>(bus_sz, r, alloc);
+}
+
+template<concepts::Resource Resource>
+auto make_message_manager(std::size_t bus_sz, Resource r)
+{
+  using Allocator = RAJA::ResourceAllocator<char, Resource>;
```
MrBurmark: I'm not sure we want to default this behavior if we want people to use something like umpire instead.

bechols97: Hmm, maybe I am missing something and can certainly remove the default if preferred. But from our earlier discussion, I thought the reason for adding the `ResourceAllocator` was to have a default allocator with users still having an option to use a more specialized allocator (like an umpire allocator) if need be?

bechols97: Removed default allocator

docs/sphinx/user_guide/feature/messages.rst
```diff
@@ -0,0 +1,201 @@
...
+  is ``false``. 
+
+Allocators
+^^^^^^^^^^
+During the creation of the ``MessageManager``, a custom
+allocator needs to be provided. This allocator should follow the C++ standard
+requirements for an allocator and will be used to allocate/deallocate
+the message bus. The message bus will always read messages on the host
+to forward the message to the respect callback(s). However, depending on the build
+and where messages are store, messages may be written on either the host
```
MrBurmark:
```suggestion
and where messages are stored, messages may be written on either the host
```

MrBurmark (approved):

PR #1836, opened 2025-04-30
Enable optional omp 5.1 atomic min/max

artv3:
This PR enables a workaround for issue https://github.com/LLNL/RAJA/issues/1812.

### DIFF
--- a/include/RAJA/policy/openmp/atomic.hpp
+++ b/include/RAJA/policy/openmp/atomic.hpp
@@ -88,7 +88,7 @@ RAJA_SUPPRESS_HD_WARN
 template<typename T>
 RAJA_HOST_DEVICE RAJA_INLINE T atomicMin(omp_atomic, T* acc, T value)
 {
-#if _OPENMP >= 202011
+#if _OPENMP >= 202011 && defined(RAJA_OPENMP_5_1_ATOMICS)
   T old;
 #pragma omp atomic capture compare
   {
@@ -109,7 +109,7 @@ RAJA_SUPPRESS_HD_WARN
 template<typename T>
 RAJA_HOST_DEVICE RAJA_INLINE T atomicMax(omp_atomic, T* acc, T value)
 {
-#if _OPENMP >= 202011
+#if _OPENMP >= 202011 && defined(RAJA_OPENMP_5_1_ATOMICS)
   T old;
 #pragma omp atomic capture compare
   {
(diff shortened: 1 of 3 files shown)


include/RAJA/policy/openmp/atomic.hpp
```diff
@@ -88,7 +88,7 @@ RAJA_SUPPRESS_HD_WARN
 template<typename T>
 RAJA_HOST_DEVICE RAJA_INLINE T atomicMin(omp_atomic, T* acc, T value)
 {
-#if _OPENMP >= 202011
+#if _OPENMP >= 202011 && defined(RAJA_OPENMP_5_1_ATOMICS)
```
adayton1, line 91: Could you just use `RAJA_ENABLE_OPENMP_5_1_ATOMICS` directly here instead of adding a secondary macro `RAJA_OPENMP_5_1_ATOMICS`?

artv3: Yes, I thought the same thing. But I see examples where we do this in a few places ,for example: 

```
#if defined(RAJA_ENABLE_BOUNDS_CHECK)
#define RAJA_BOUNDS_CHECK_INTERNAL
#define RAJA_BOUNDS_CHECK_constexpr
#else
#define RAJA_BOUNDS_CHECK_constexpr constexpr
#endif
```

```
#if defined(RAJA_ENABLE_OPENMP_TASK)
#define RAJA_ENABLE_OPENMP_TASK_INTERNAL
#endif
```

I'm not sure why we do this, but if we want to keep it consistent maybe I add to  the secondary macro _INTERNAL. 

What do others think?

adayton1: Maybe add the check for _OPENMP being defined and >= 202011 when defining the internal macro (maybe named RAJA_USE_OPENMP_5_1_ATOMICS?), and then the actual atomic implementation could simply use the internal macro instead of checking the openmp version.

adayton1: Or maybe the configuration variable should be RAJA_ALLOW_OPENMP_5_1_ATOMICS and the internal define should be RAJA_ENABLE_OPENMP_5_1_ATOMICS. Just brainstorming.

artv3: Sounds good to me!

adayton1 (approved):

adayton1:
Thanks, @artv3 !

PR #1868, opened 2025-07-18
Updates to IndexLayout

gberg617:
# Summary

- This PR is a feature that adds a few minor things to IndexLayout.
  - Adds size methods that call base Layout implementation.  These could be used to check the size of the layout or to determine whether the size is non-zero. All the size methods directly call the base Layout implementation.
  - Adds host/device decorators to `make_index_tuple` and `make_index_layout`


include/RAJA/util/IndexLayout.hpp
```diff
@@ -126,6 +126,28 @@ struct IndexLayout_impl<camp::idx_seq<RangeInts...>, IdxLin, IndexTypes...>
     return sum<IdxLin>(
         (base_.strides[RangeInts] * camp::get<RangeInts>(tuple)(indices))...);
   }
+
+  RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexLinear size() const
```
gberg617, line 130: Does these need unit tests?

gberg617: OK I added a few conditions to check in the existing unit tests.

rhornung67 (approved):

PR #1884, opened 2025-08-18
cpp 17: Simplify CUDA, HIP, and OpenMP forall implementations with if constexpr

johnbowen42:
This PR simplifies forall implementations by using if constexpr.  Specifically, SFINAE overloads for empty/nonempty forall param packs are removed from CUDA and HIP, because we can use `if constexpr` to conditionally init/combine/resolve/ param packs.  For OpenMP, a few other SFINAE overloads are simplified as well


trws (approved):
I love this kind of PR, good stuff.

I notice some of the OpenMP stuff is still duplicated, if the compilers support it those could be collapsed out with metadirective too but this is great. 👍

PR #1912, opened 2025-09-15
cpp 17: Simplify launch implementations if constexpr

johnbowen42:


johnbowen42:
> Is there anything to do for the OpenMP backend?

I couldn't identify an easy simplification for the OpenMP backend

adayton1 (approved):

include/RAJA/pattern/params/forall.hpp
```diff
@@ -238,6 +238,36 @@ struct ForallParamPack
...
+template<typename First, typename... Rest>
+struct is_ForallParamPack_empty<ForallParamPack<First, Rest...>>
+    : std::false_type
+{};
+
+template<>
+struct is_ForallParamPack_empty<ForallParamPack<>> : std::true_type
+{};
+}  // namespace type_traits
+
```
artv3: maybe we should move this to a "params utils" file?

artv3 (approved):
Look great! Just 1 comment on where to place helpers.

include/RAJA/pattern/params/forall.hpp
johnbowen42: I did something similar--I moved all our type trait stuff into its own header

johnbowen42:
@artv3  I thought a bit more about our type traits and reorganized the headers a bit.  I remembered I made a type traits header for kernel-specific work.  I renamed that to be more consistent, and made another TypeTraits.hpp for reduction work.  Let me know what you think

PR #1914, opened 2025-09-19
Make comparator operator() call const

smithsg84:
# Summary

- This PR is a minor refactoring adding const to the the operator() for  the tuple first member comparator.


MrBurmark:
We should get this in the next patch release.

adayton1 (approved):

PR #1915, opened 2025-09-19
Add SubView class

gberg617:
# Summary

- This WIP PR adds the SubView class.  This class stores a view-type object and a tuple of slices that indicate how to access the view's underlying data.  The following slice types are supported:
  - range slice: Indexes a subset of the original dimension's extents.  Does not reduce dimensionality.
  - fixed slice: Indexes along a fixed value of a given dimension.  Reduces dimensionality.
  - no slice: Indexes according to the original view's layout along the given dimension.  Does not reduce dimensionality.

A few simple examples are provided below:
```
    View<Index_type, Layout<2>> view(&my_data[0][0], Layout<2>(3,3));

    // "sv1 = View[1:3,:]"
    auto sv1 = SubView(view, RangeSlice{1,3}, NoSlice{});

    // "sv2 = View[1:3,1]"
    auto sv2 = SubView(view, RangeSlice{1,3}, FixedSlice{1});
```

Here, `sv1` and `sv2` are subviews of dimensions `(2,3)` and `(2)` respectively.

Future tasks:

- Add Slices whose extents are determined at compile time
- Add Slice types that use non-affine mappings (e.g. index lists)
- Add ability to collapse nested sub-layouts into a single sub-layout for slices that have affine mappings
- Add ability to distinguish between projected dimension and an empty slice (e.g. RangeSlice{2,2})


gberg617:
This PR is still a WIP!

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include "camp/tuple.hpp"
+#include "camp/array.hpp"
+
+namespace RAJA
+{
+
+// Slice descriptors
+struct RangeSlice { long start, end; };
+struct FixedSlice { long idx; };
+struct NoSlice { };
```
MrBurmark, line 32: We will probably want to generalize the integer types used in the slices. We will probably want to add strongly typed slices like the strongly typed indices.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+template <typename T, size_t N, RAJA::Index_type... Is>
+RAJA_INLINE RAJA_HOST_DEVICE constexpr auto array_to_tuple_impl(const camp::array<T, N>& arr, camp::idx_seq<Is...>) {
+    return camp::make_tuple(arr[Is]...);
+}
+
+template <typename T, size_t N>
+RAJA_INLINE RAJA_HOST_DEVICE constexpr auto array_to_tuple(const camp::array<T, N>& arr) {
+    return array_to_tuple_impl(arr, camp::make_idx_seq_t<N>{});
+}
```
MrBurmark, line 49: Do we not have this somewhere?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,96 @@
...
+                parent_indices[Is] = [&] {
+                    const auto& s = camp::get<Is>(slices_);
+                    if constexpr (std::is_same_v<std::decay_t<decltype(s)>, RangeSlice>) {
+                        return s.start + idxs[idx++];
+                    } else if constexpr (std::is_same_v<std::decay_t<decltype(s)>, FixedSlice>) {
+                        return s.idx;
+                    } else {
+                        return idxs[idx++];
+                    }
+                }()
```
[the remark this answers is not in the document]
gberg617: I moved it to a member function.

test/unit/view-layout/test-subview.cpp
```diff
@@ -0,0 +1,120 @@
...
+    EXPECT_EQ(sv(0,1), 5);
+    EXPECT_EQ(sv(0,2), 6);
+
+    EXPECT_EQ(sv(1,0), 7);
+    EXPECT_EQ(sv(1,1), 8);
+    EXPECT_EQ(sv(1,2), 9);
+
+}
+
+// void test_subviewGPU() {
```
gberg617, line 101: To Do: Fix GPU tests.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType map_index(IndexType&) const {
+        return idx;
+    }
+};
+
+struct NoSlice { 
+    static constexpr bool reduces_dimension = false;
+
+    template<typename IndexType = Index_type>
```
MrBurmark, line 56: Would it make more sense to template all of these classes on IndexType so they are consistent?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,120 @@
...
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType map_index(IndexType& idx) const {
+        return idx;
+    }
+};
+
+// Helper to count non-fixed dimensions at compile time
+template <typename... Slices>
+RAJA_INLINE RAJA_HOST_DEVICE constexpr size_t count_nonfixed_dims() {
+    return (!Slices::reduces_dimension + ...);
```
MrBurmark, line 65: Are we adding booleans here? How about a ternary with 1 and 0?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,120 @@
...
+class SubView {
+    ViewType view_;
+    camp::tuple<Slices...> slices_;
+    std::array<IndexType, sizeof...(Slices)> map_;
+
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr void make_subview_index_map() {
+        size_t sub_idx = 0;
+        size_t i = 0;
+        ((map_[i++] = (Slices::reduces_dimension ? -1 : sub_idx++)), ...);
+    }
```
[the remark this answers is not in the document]
MrBurmark: It would probably be more clear if make_subview_index_map returned an array. If possible it would be nice to generate this map in a way that looks as much like a normal for loop doing a scan as possible. I think there is a RAJA util like for_each_type that might be useful here.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,120 @@
...
+RAJA_INLINE RAJA_HOST_DEVICE constexpr auto array_to_tuple_impl(const camp::array<T, N>& arr, camp::idx_seq<Is...>) {
+    return camp::make_tuple(arr[Is]...);
+}
+
+template <typename T, size_t N>
+RAJA_INLINE RAJA_HOST_DEVICE constexpr auto array_to_tuple(const camp::array<T, N>& arr) {
+    return array_to_tuple_impl(arr, camp::make_idx_seq_t<N>{});
+}
+
+template <typename ViewType, typename IndexType = Index_type, typename... Slices>
```
MrBurmark: Think about doing using a camp::list<slices...> to avoid having a parameter pack in the class template.

test/unit/view-layout/test-subview.cpp
```diff
@@ -0,0 +1,120 @@
...
+    EXPECT_EQ(sv(0,0), 4);
+    EXPECT_EQ(sv(0,1), 5);
+    EXPECT_EQ(sv(0,2), 6);
+
+    EXPECT_EQ(sv(1,0), 7);
+    EXPECT_EQ(sv(1,1), 8);
+    EXPECT_EQ(sv(1,2), 9);
+
+}
+
```
rcarson3, line 343: @gberg617 how would you do something like we have in the original impl of SubViews: https://github.com/LLNL/SNLS/blob/develop/test/SNLS_forall_subviews.cxx#L69-L75 where we allow for a sliding window with the SubView. Additionally, how can we get the underlying data pointer at the SubView's (0..) index like here: https://github.com/LLNL/SNLS/blob/develop/test/SNLS_forall_subviews.cxx#L150 ?

include/RAJA/util/SubView.hpp
gberg617: Fixed.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+public:
+
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr SubView(ViewType view, Slices... slices)
+        : view_(view), slices_(slices...) { make_subview_index_map(); }
+
+    template <typename... Idxs>
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType operator()(Idxs... idxs) const {
+        constexpr size_t nidx = count_nonfixed_dims<Slices...>();
```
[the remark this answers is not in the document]
gberg617: I simplified this a bit

include/RAJA/util/SubView.hpp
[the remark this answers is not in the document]
gberg617: Fixed.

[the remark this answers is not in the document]
gberg617: I found having a `for_each` that calls a func with the tuple and index seemed to be the cleanest.  Let me know your thoughts.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr void make_subview_index_map() {
+        size_t sub_idx = 0;
+        size_t i = 0;
+        ((map_[i++] = (Slices::reduces_dimension ? -1 : sub_idx++)), ...);
+    }
+
+    template<IndexType I>
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr auto map_subview_idx_to_parent(IndexType* idxs) const {
+        return camp::get<I>(slices_).map_index(idxs[map_[I]]);
```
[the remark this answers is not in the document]
gberg617: I simplified this a bit.  Let me know your thoughts.

include/RAJA/util/SubView.hpp
gberg617: I tried doing something like that.  I'm not quite sure I had the right approach.

test/unit/view-layout/test-subview.cpp
gberg617: Added `set_slice` to allow things like "sliding windows"

gberg617:
I'm considering separating the SubView holding the view from the slices that are used to access the view,  somewhat analogous to the approach of having layouts and Views be separate classes.  Any thoughts on this?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,144 @@
...
+}
+
+template <typename ViewType, typename SliceTypes, typename IndexType = Index_type>
+class SubView;
+
+template <typename ViewType, typename IndexType, typename... Slices>
+class SubView<ViewType, camp::list<Slices...>, IndexType> {
+    ViewType view_;
+    camp::tuple<Slices...> slices_;
+    std::array<IndexType, sizeof...(Slices)> map_;
```
MrBurmark: Still need to make this static inline constexpr

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr auto make_subview_index_map() {
+        size_t sub_idx = 0;
+        std::array<IndexType, sizeof...(Slices)> map;
+
+        for_each_tuple_index( slices_,
```
MrBurmark: This looks great. Unfortunately it uses a non-static member to do something static.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    template<IndexType Index>
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr auto& get_slice() {
+        return camp::get<Index>(slices_);
+    }
+
+    template <typename... Idxs>
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType operator()(Idxs... idxs) const {
+        constexpr size_t nidx = ((Slices::reduces_dimension == false ? 1 : 0) + ...);
```
MrBurmark: This should be a static member variable of the class.

MrBurmark: It would be nice to have something like num_slices as well.

MrBurmark: It would also be nice to have a bit more descriptive names.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,144 @@
...
+        static_assert(sizeof...(idxs) == nidx, "Wrong number of indices for subview");
+
+        camp::array<Index_type, nidx> arr{idxs...};
+        camp::array<Index_type, sizeof...(Slices)> parent_indices;
+
+        for_each_tuple_index( slices_,
+            [&](auto slice, auto index) { 
+                parent_indices[index] = slice.map_index(arr[map_[index]]);
```
MrBurmark: This causes out of bounds accesses for slices that reduce dimensions.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+        camp::array<Index_type, nidx> arr{idxs...};
+        camp::array<Index_type, sizeof...(Slices)> parent_indices;
+
+        for_each_tuple_index( slices_,
+            [&](auto slice, auto index) { 
+                parent_indices[index] = slice.map_index(arr[map_[index]]); 
+            });
+
+        return camp::apply(view_, array_to_tuple(parent_indices));
```
MrBurmark: Why not just use a tuple to start with?

include/RAJA/util/for_each.hpp
```diff
@@ -79,8 +96,31 @@ RAJA_HOST_DEVICE RAJA_INLINE UnaryFunc for_each_tuple(Tuple&& t,
...
+// compile time expansion applying func to a each type in the tuple in order
+RAJA_SUPPRESS_HD_WARN
+template<typename Tuple, typename BinaryFunc, camp::idx_t... Is>
+RAJA_HOST_DEVICE RAJA_INLINE BinaryFunc for_each_tuple_index(Tuple&& t,
+                                                             BinaryFunc func,
+                                                             camp::idx_seq<Is...>)
+{
+  using camp::get;
+  // braced init lists are evaluated in order
+  int seq_unused_array[] = {0, (func(get<Is>(std::forward<Tuple>(t)), Is), 0)...};
```
MrBurmark:
```suggestion
  int seq_unused_array[] = {0, (func(get<Is>(std::forward<Tuple>(t)), std::integral_constant<std::size_t, Is>{}), 0)...};
```

include/RAJA/util/for_each.hpp
```diff
@@ -36,6 +39,20 @@ namespace RAJA
 namespace detail
 {
 
+// compile time expansion applying func to each of the indices
+RAJA_SUPPRESS_HD_WARN
+template<typename UnaryFunc, typename IndexType, IndexType... Is>
+RAJA_HOST_DEVICE RAJA_INLINE UnaryFunc for_each_index(camp::idx_seq<Is...>,
```
MrBurmark:
```suggestion
template<typename UnaryFunc, camp::idx_t... Is>
RAJA_HOST_DEVICE RAJA_INLINE UnaryFunc for_each_index(camp::idx_seq<Is...>,
```

MrBurmark: It should all be size_t or camp::idx_t.

MrBurmark:
> I'm considering separating the SubView holding the view from the slices that are used to access the view, somewhat analogous to the approach of having layouts and Views be separate classes. Any thoughts on this?

You mean something more like making a SubLayout that you can use anywhere a layout can be used?

gberg617:
> > I'm considering separating the SubView holding the view from the slices that are used to access the view, somewhat analogous to the approach of having layouts and Views be separate classes. Any thoughts on this?
> 
> You mean something more like making a SubLayout that you can use anywhere a layout can be used?

Yes, but the intent is that these "sub-layouts" are only used with SubViews.  I will try to prototype something here soon.

MrBurmark:
> > You mean something more like making a SubLayout that you can use anywhere a layout can be used?
> 
> Yes, but the intent is that these "sub-layouts" are only used with SubViews. I will try to prototype something here soon.

We'd have to think about it. It seems simpler to me if you can use a sublayout like a layout, that way a subview would just be a view using a sublayout.

include/RAJA/util/TypedViewBase.hpp
```diff
@@ -669,16 +669,31 @@ class ViewBase
   constexpr layout_type const& get_layout() const { return m_layout; }
 
   RAJA_HOST_DEVICE
-
   RAJA_INLINE
   constexpr linear_index_type size() const { return m_layout.size(); }
 
+  RAJA_HOST_DEVICE
```
gberg617, line 669: changes to View and MultiView are still a WIP.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+    static constexpr bool reduces_dimension = false;
+
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType map_index(const IndexType& idx) const {
+        return start_ + idx;
+    }
+
+    template<IndexType DIM, typename LayoutType>
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType size(const LayoutType&) const {
+        return (end_ - start_ + 1);
+    }
```
MrBurmark: Why is DIM here? If its unused we should use the unused arg macro so we don't get a warning.

MrBurmark: Also why the +1, this is including end?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType stride() const {
+        return 1;
+    }
+};
+
+template<typename IndexType = Index_type>
+struct StridedSlice { 
+    IndexType start_, end_, stride_;
```
MrBurmark: Does this support negative strides? I'm pretty sure that the strided range supports negative strides.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType stride() const {
+        return stride_;
+    }
+};
+
+template<typename IndexType, typename... Slices>
+RAJA_INLINE RAJA_HOST_DEVICE constexpr auto make_slice_to_parent_index_map() {
+    IndexType sub_idx = 0;
+    IndexType i = 0;
+    camp::array<IndexType, sizeof...(Slices)> map = {};
```
MrBurmark: Why not do something like this?
```suggestion
    camp::array<IndexType, sizeof...(Slices)> map{};
```

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+    ((map[i++] = (Slices::reduces_dimension ? -1 : sub_idx++)), ...);
```
MrBurmark: Or even do something like this?
```suggestion
    camp::array<IndexType, sizeof...(Slices)> map{{(Slices::reduces_dimension ? -1 : sub_idx++)...}};
```

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+    return map;
+}
+
+template<typename IndexType, size_t n_parent_dims, typename... Slices>
+RAJA_INLINE RAJA_HOST_DEVICE constexpr auto make_parent_to_slice_index_map() {
```
MrBurmark: Does it make more sense to take n_parent_dims as an argument or calculate it inside?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+/* SubView is a semantic alias for a SubRegion whose parent is a view */
+template <typename LayoutType, typename SliceTypes, typename IndexType = Index_type>
+using SubView = SubRegion<LayoutType, SliceTypes, IndexType>;
+
+template <typename LayoutType, typename IndexType, typename... Slices>
+struct SubRegion<LayoutType, camp::list<Slices...>, IndexType> {
+
+    using IndexLinear = IndexType;
+
+    const LayoutType& parent_;
```
MrBurmark: Instead of having a copy of the parent this has a reference? What if you make the SubRegion on the host and then copy it to the device?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+    camp::tuple<Slices...> slices_;
+
+    static inline constexpr size_t num_slices_ = sizeof...(Slices);
```
MrBurmark: Static data members and functions should be above non-static data members and functions.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+
+    static inline constexpr 
+    IndexType n_dims = ((Slices::reduces_dimension == false ? 1 : 0) + ...);
```
MrBurmark: How about using `Slices::reduces_dimension` as a boolean if its a boolean?

MrBurmark: I like to have all the uses of something look similar if possible. I think this is often used like this `!Slices::reduces_dimension` so it would be simpler to understand if this also looked the same.

include/RAJA/util/SubView.hpp
MrBurmark: Also what do you think about using the member variables are named with m_ and static members are named with s_ convention?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr auto& get_slices() const {
+        return slices_;
+    }
+
+    template<IndexType Index>
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr auto& get_slice() const {
+        return camp::get<Index>(slices_);
+    }
```
MrBurmark: Since there are const they should probably return const references.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr auto size() const {
+
+        IndexType prod_dims = 1;
+        for_each_tuple_index( slices_,
+            [&](auto slice, auto index) {
+                const IndexType dim_size = slice.template size<index>(parent_);
+                prod_dims *= (dim_size == 0) ? 1 : dim_size;
```
MrBurmark: If there really is a dimension that is completely projected out, shouldn't the size be 0?

MrBurmark: Also do we have to consider the 0 dimension case?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+                prod_dims *= slice.template size<index>(parent_);
+            });
+
+        return prod_dims;
+    }
+
+    template<IndexType DIM> 
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr auto get_dim_size() const {
+        constexpr auto SliceDim = parent_to_slice_map_[DIM];
+        return camp::get<SliceDim>(slices_).template size<DIM>(parent_);
```
MrBurmark: We should be able to static_assert that DIM is in bounds.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+    template <typename... Idxs>
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr auto operator()(Idxs... idxs) const {
+        static_assert(sizeof...(idxs) == n_dims, "Wrong number of indices");
+
+        camp::array<IndexType, n_dims> arr{idxs...};
+        camp::array<IndexType, num_slices_> parent_indices;
+
+        for_each_tuple_index( slices_,
+            [&](auto slice, auto index) {
+                if (slice_to_parent_map_[index] >= 0) {
```
MrBurmark: should this be if constexpr?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,268 @@
...
+                    parent_indices[index] = slice.map_index(arr[slice_to_parent_map_[index]]); 
+                } else {
+                    // map_index will not need index values for dimension-reducing slices
+                    // so we pass a "dummy" value.
+                    constexpr IndexType dummy_value = -1;
+                    parent_indices[index] = slice.map_index(dummy_value);
```
MrBurmark: Should dimension reducing slices not take a value at all?

include/RAJA/util/SubView.hpp
rhornung67: It does support negative strides. For example, specifying stride = -1 is how we reverse the iteration order of a loop.

include/RAJA/util/SubView.hpp
gberg617: Yup.  I added a new map_index() function that takes no arguments for dimension reducing slices.

include/RAJA/util/SubView.hpp
gberg617: Yes, fixed.

include/RAJA/util/SubView.hpp
gberg617: Added.

include/RAJA/util/SubView.hpp
gberg617: I'm not sure what you mean here.  The intent was to keep the behavior consistent with what's in Layout: size() will replace zero-sized dimensions with 1 when computing the product of sizes.

include/RAJA/util/SubView.hpp
gberg617: Fixed.

include/RAJA/util/SubView.hpp
gberg617: Do you mean computing n_dims like this: `n_dims = (!Slices::reduces_dimension + ...)`?

include/RAJA/util/SubView.hpp
gberg617: Fixed.

include/RAJA/util/SubView.hpp
gberg617: Fixed, except for `n_dims` which had to stay as it was.

include/RAJA/util/SubView.hpp
gberg617: Fixed.

include/RAJA/util/SubView.hpp
gberg617: Fixed.

include/RAJA/util/SubView.hpp
gberg617: Fixed.

include/RAJA/util/SubView.hpp
gberg617: RangeSlice includes the start and end elements, which is why we need the +1.  So `RangeSlice<>{1,3}` will have a size of 3.  Would it make more sense to have the range be `[start, end)`? Perhaps that's the more standard approach.

include/RAJA/util/SubView.hpp
gberg617: I added negative stride support in size().  It's implicitly assumed here that start/end ordering are consistent with stride sign.  So start < end when stride > 0, and start > end when stride < 0.  I can add run-time checks for this if needed.

include/RAJA/util/SubView.hpp
rhornung67: We've discussed this and I thought we had an issue for it. But I can't find it. Let's discuss at the project meeting tomorrow.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,272 @@
...
+
+    static constexpr bool reduces_dimension = false;
+
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType map_index(const IndexType& idx) const {
+        return start_ + idx;
+    }
+
+    template<IndexType RAJA_UNUSED_ARG(DIM), typename LayoutType>
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType size(const LayoutType&) const {
+        return (end_ - start_ + 1);
```
gberg617: Should we consider changing RangeSlice to the half-open interval `[start_, end_)` instead of the closed interval `[start_, end_]`?

MrBurmark: That would be more consistent with how most things work in C++.

include/RAJA/util/SubView.hpp
MrBurmark: no I mean
```
((!Slices::reduces_dimension ? 1 : 0) + ...)
```

include/RAJA/util/SubView.hpp
MrBurmark: Didn't we agreed that Layout is wrong and that if a dimension has zero items then the product should be 0. I guess its ok to be consistent but I thought we were going to fix that in Layout in the next major release.

MrBurmark: I suppose that completely projected out just means that a single value was chosen? So the size is one in that direction?

include/RAJA/util/SubView.hpp
gberg617: I ended up adding more checks to size() for StridedSlice.  Would it make sense at all to generally consider supporting negative indexing as well?

include/RAJA/util/SubView.hpp
gberg617: Fixed.

tpl/camp
artv3, line 1: does camp need to be updated?

include/RAJA/util/SubView.hpp
gberg617: Fixed.

include/RAJA/util/SubView.hpp
gberg617: Fixed.

include/RAJA/util/SubView.hpp
gberg617: Fixed.

include/RAJA/util/SubView.hpp
gberg617: I think this might be outdated?

include/RAJA/util/SubView.hpp
gberg617: Fixed.

include/RAJA/util/SubView.hpp
gberg617: I think this is also outdated.  We moved to using camp::array.

include/RAJA/util/for_each.hpp
gberg617: Fixed.

tpl/camp
gberg617: Yes, it does.  Thanks for catching that.  See https://github.com/llnl/camp/pull/194.

include/RAJA/util/View.hpp
```diff
@@ -165,6 +165,8 @@ struct MultiView
   using NonConstView =
       MultiView<nc_value_type, layout_type, P2Pidx, nc_pointer_type>;
 
+  static constexpr size_t n_dims = layout_type::n_dims + 1;
```
gberg617, line 168: @rchen20 Does this look right for MultiViews?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+template <typename T, size_t N>
+RAJA_INLINE RAJA_HOST_DEVICE constexpr auto array_to_tuple(const camp::array<T, N>& arr) {
+    return array_to_tuple_impl(arr, camp::make_idx_seq_t<N>{});
+}
+
+template <typename ViewType, typename IndexType = Index_type, typename... Slices>
+class SubView {
+    ViewType view_;
+    camp::tuple<Slices...> slices_;
```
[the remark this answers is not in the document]
gberg617: This compile-time check is a good idea.  The best way I could think of doing this is adding `n_dims` to each view type to have a consistent way of checking for the number of dimensions in the view. See this commit for those changes: https://github.com/llnl/RAJA/pull/1915/changes/0b198afb679758a08e5856af92d85ba48baf638d.

include/RAJA/util/for_each.hpp
gberg617: Fixed.

include/RAJA/util/SubView.hpp
gberg617: Fixed.

include/RAJA/util/SubView.hpp
gberg617: This function wasn't eventually needed.  That said, I can still add something like this to camp if you think it could be useful.

include/RAJA/util/SubView.hpp
```diff
@@ -263,15 +243,28 @@ struct SubRegion<LayoutType, camp::list<Slices...>, IndexType> {
...
+            [&](auto dim_c) constexpr {
+                if constexpr (decltype(dim_c)::value > DIM) {
+                    stride *= this->template get_dim_size<decltype(dim_c)::value>();
+                }
+            });
+        return stride;
+    }
+
+    template<size_t DIM>
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr auto get_parent_dim_stride() const {
```
gberg617, line 266: I'm not sure how useful `get_parent_dim_stride()` is.

test/unit/view-layout/test-subview.cpp
gberg617: Perhaps remove GPU tests?

gberg617:
@MrBurmark Feel free to go through this PR when you have the chance.  The implementation is mostly complete.  I added a few notes of possible extensions that we can eventually implement in the description above.  Let me know if I should also include documentation in this PR as well.

rhornung67:
@gberg617 yes, please add some documentation. Also, if you configure the code with clang 19 on LC and run clang-format, you will get past the style check which prevents all the other GH Actions jobs from running.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,298 @@
...
+        constexpr auto SliceDim = s_parent_to_slice_map[DIM];
+        return m_parent.template get_dim_stride<SliceDim>() *
+                camp::get<SliceDim>(m_slices).stride();
+    }
+
+    template <typename... Idxs>
+    RAJA_INLINE RAJA_HOST_DEVICE constexpr auto operator()(Idxs... idxs) const {
+        static_assert(sizeof...(idxs) == n_dims, "Wrong number of indices");
+
+        std::array<IndexType, n_dims> arr{idxs...};
```
gberg617: @MrBurmark Did we decide to go with `std::array`, or switch back to `camp::array` here?

MrBurmark: Let's use camp array, we haven't asked users to add the cuda flag that would enable this yet.

gberg617: OK, I added back camp::array.  This means that we first need to push the following camp PR:
https://github.com/llnl/camp/pull/194

rhornung67:
@gberg617 do you want this included in the July release?

include/RAJA/util/for_each.hpp
```diff
@@ -26,6 +26,9 @@
 #include <type_traits>
```
gberg617, line 27: I need to fix this rebased file.

test/unit/view-layout/test-subview.cpp
```diff
@@ -1,5 +1,5 @@
 //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
-// Copyright (c) 2016-25, Lawrence Livermore National Security, LLC
+// Copyright (c) 2016-26, Lawrence Livermore National Security, LLC
```
rhornung67, line 2: @gberg617 we changed the copyright header in all RAJA files so when we update each year we only need to touch two files rather than the whole repo. Here is what you should have on all files you add or modify: 
https://github.com/llnl/RAJA/blob/develop/test/unit/view-layout/test-indexlayout.cpp#L1

rhornung67 (changes requested):
See comment about copyright header.

gberg617:
@MrBurmark @adayton1 I assume that SubViews can be 0D in certain cases where the slice(s) reduce the dimension to a scalar. This leads to compiler errors on Windows (see failing CI) because camp array stores a C-style array of size 0. My thoughts are to either disallow 0D SubViews or add a specialization of camp array for a 0-sized array. My preference is the latter because std::array can be zero-sized, so this addition would make camp::array be more consistent with std::array. Any thoughts?

MrBurmark:
> @MrBurmark @adayton1 I assume that SubViews can be 0D in certain cases where the slice(s) reduce the dimension to a scalar. This leads to compiler errors on Windows (see failing CI) because camp array stores a C-style array of size 0. My thoughts are to either disallow 0D SubViews or add a specialization of camp array for a 0-sized array. My preference is the latter because std::array can be zero-sized, so this addition would make camp::array be more consistent with std::array. Any thoughts?

Adding a camp array specialization for 0-size makes a lot of sense.

adayton1:
> > @MrBurmark @adayton1 I assume that SubViews can be 0D in certain cases where the slice(s) reduce the dimension to a scalar. This leads to compiler errors on Windows (see failing CI) because camp array stores a C-style array of size 0. My thoughts are to either disallow 0D SubViews or add a specialization of camp array for a 0-sized array. My preference is the latter because std::array can be zero-sized, so this addition would make camp::array be more consistent with std::array. Any thoughts?
> 
> Adding a camp array specialization for 0-size makes a lot of sense.

I can't remember for sure, but I think that I disallowed the 0 size array because nvcc couldn't handle it (or maybe it was some other compiler). So just do plenty of testing.

test/unit/view-layout/test-subview.cpp
gberg617: Fixed! Sorry for the delay on this.

gberg617:
@rhornung67 @MrBurmark This PR is ready for review

docs/sphinx/user_guide/feature/view.rst
```diff
@@ -138,21 +138,91 @@ the integer indicating which array to access can be reversed.
...
+
+  #include "RAJA/util/SubView.hpp"
+
+  using ParentView = RAJA::View<double, RAJA::Layout<2>>;
+  ParentView A(data, N_r, N_c);
+
+  using Slices = camp::list<RAJA::RangeSlice<>, RAJA::RangeSlice<>>;
+  RAJA::SubView<ParentView, Slices> sub(A,
+                                        RAJA::RangeSlice<>{1, 3},
+                                        RAJA::RangeSlice<>{1, 4});
```
MrBurmark: Do we also have deduction guides to make it possible to avoid naming the SubView type?

docs/sphinx/user_guide/feature/view.rst
```diff
@@ -138,21 +138,91 @@ the integer indicating which array to access can be reversed.
...
+  ParentLayout parent_layout(N_r, N_c);
+
+  using SubLayout = RAJA::SubLayout<
+      ParentLayout, camp::list<RAJA::RangeSlice<>, RAJA::RangeSlice<>>>;
+  SubLayout sub_layout(parent_layout,
+                       RAJA::RangeSlice<>{1, 3},
+                       RAJA::RangeSlice<>{1, 4});
+
+  RAJA::View<double, SubLayout> sub(data, sub_layout);
+  sub(i, j) = 0.0;
```
MrBurmark, line 238: This is the documentation, I thought the intention was to make it easy to make a view with a sublayout. Is there a less verbose way to do that? Are there helper functions to do these things?

include/RAJA/util/for_each.hpp
```diff
@@ -90,7 +137,8 @@ for_each_tuple(Tuple&& t, UnaryFunc func, camp::idx_seq<Is...>)
 RAJA_SUPPRESS_HD_WARN
 template<typename Container, typename UnaryFunc>
 constexpr RAJA_HOST_DEVICE RAJA_INLINE
-    concepts::enable_if_t<UnaryFunc, type_traits::is_range<Container>>
+    camp::concepts::enable_if_t<UnaryFunc,
+                                camp::type_traits::is_range<Container>>
     for_each(Container&& c, UnaryFunc func)
```
MrBurmark, line 142: Did we want to use concepts now that we're c++20?

MrBurmark: I suppose the other utils here are not c++20-ified so we can do this whole file later.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,345 @@
...
+#include "RAJA/util/for_each.hpp"
+#include "RAJA/util/macros.hpp"
+#include "RAJA/util/types.hpp"
+#include "camp/tuple.hpp"
+#include "camp/array.hpp"
+
+namespace RAJA
+{
+
+template<typename IndexType = Index_type>
```
MrBurmark, line 42: Do we want to constrain the indextype at all?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,345 @@
...
+struct RangeSlice
+{
+  IndexType m_start, m_end;
+
+  static constexpr bool reduces_dimension = false;
+
+  RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType map_index(
+      const IndexType& idx) const
```
MrBurmark: We might want to take things that are integers or integer wrappers by value. See #2087.

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,345 @@
...
+  {
+    return m_start + m_stride * idx;
+  }
+
+  template<IndexType DIM, typename LayoutType>
+  RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType size(const LayoutType&) const
+  {
+    if (m_stride == 0)
```
MrBurmark: Should we consider this UB?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,345 @@
...
+
+  auto process_slice = [&](auto slice_type) constexpr {
+    if constexpr (!decltype(slice_type)::reduces_dimension)
+    {
+      map[sub_idx++] = i++;
+    }
+    else
+    {
+      i++;
+    }
```
MrBurmark:
```suggestion
    if constexpr (!decltype(slice_type)::reduces_dimension)
    {
      map[sub_idx++] = i;
    }
    i++;
```
Is this simpler?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,345 @@
...
+      ((!Slices::reduces_dimension ? 1 : 0) + ...);
+
+  static inline constexpr camp::array<size_t, s_num_slices>
+      s_slice_to_parent_map = make_slice_to_parent_index_map<Slices...>();
+
+  static inline constexpr camp::array<size_t, n_dims> s_parent_to_slice_map =
+      make_parent_to_slice_index_map<Slices...>();
+
+  const LayoutType m_parent;
+  camp::tuple<Slices...> m_slices;
```
MrBurmark: Should this be private? Also should the maps be private?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,345 @@
...
+                       : slice.template size<index>(m_parent);
+    });
+
+    return prod_dims;
+  }
+
+  template<size_t DIM>
+  RAJA_INLINE RAJA_HOST_DEVICE constexpr auto get_dim_size() const
+  {
+    static_assert(DIM < n_dims, "DIM out of bounds");
```
MrBurmark: Make this a requires?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,345 @@
...
+  {
+    return m_parent;
+  }
+
+  RAJA_INLINE RAJA_HOST_DEVICE constexpr const auto& get_slices() const
+  {
+    return m_slices;
+  }
+
+  template<size_t Index>
```
MrBurmark: Add a requires that Index is in-bounds?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,345 @@
...
+    return camp::get<SliceDim>(m_slices).template size<SliceDim>(m_parent);
+  }
+
+  template<size_t DIM>
+  RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexLinear get_dim_stride() const
+  {
+    static_assert(DIM < n_dims, "DIM out of bounds");
+    IndexLinear stride = 1;
+    for_each_index<n_dims>([&](auto dim_c) constexpr {
+      if constexpr (decltype(dim_c)::value > DIM)
```
MrBurmark:
```suggestion
      if constexpr (dim_c > DIM)
```
Does this not work?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,345 @@
...
+  }
+
+  template<size_t DIM>
+  RAJA_INLINE RAJA_HOST_DEVICE constexpr auto get_parent_dim_stride() const
+  {
+    static_assert(DIM < n_dims, "DIM out of bounds");
+    constexpr auto SliceDim = s_parent_to_slice_map[DIM];
+    return m_parent.template get_dim_stride<SliceDim>() *
+           camp::get<SliceDim>(m_slices).stride();
+  }
```
MrBurmark: Is this indexing using the parent's dimensions, or the subregion's dimensions?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,345 @@
...
+
+  template<typename... Idxs>
+  RAJA_INLINE RAJA_HOST_DEVICE constexpr auto operator()(Idxs... idxs) const
+  {
+    static_assert(sizeof...(idxs) == n_dims, "Wrong number of indices");
```
MrBurmark, line 377: requires?

include/RAJA/util/SubView.hpp
MrBurmark: I remember having to do stuff like this for nvcc, was that the issue?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,345 @@
...
+  }
+
+  RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexType stride() const
+  {
+    return m_stride;
+  }
+};
+
+template<typename... Slices>
+RAJA_INLINE RAJA_HOST_DEVICE constexpr auto make_slice_to_parent_index_map()
```
MrBurmark: I'm a bit confused about the naming of slice vs parent. Aren't the slices and the parent dimensions the same thing. Isn't this really mapping from the parent dimensions to the sub-dimensions of this subregion where some dimensions are reduced out?

include/RAJA/util/TypedViewBase.hpp
```diff
@@ -674,12 +676,33 @@ class ViewBase
...
+
+  RAJA_INLINE
+  constexpr linear_index_type size_noproj() const
+  {
+    return m_layout.size_noproj();
+  }
+
+  template<camp::idx_t DIM>
+  RAJA_INLINE RAJA_HOST_DEVICE constexpr linear_index_type get_dim_stride()
```
MrBurmark, line 689: When you say "dim" here and in other places you mean the logically rectangular index space?

include/RAJA/util/View.hpp
MrBurmark: @rchen20 Does the layout include the dimension with pointers to data, or only the dimensions of each of the data pointers in that dimension?

MrBurmark: From what I can tell, the layout only handles the "inner" dimensions of a single data pointer, so adding 1 seems correct.

include/RAJA/util/View.hpp
```diff
@@ -235,6 +238,38 @@ struct MultiView
     return layout;
   }
 
+  RAJA_HOST_DEVICE
+
+  RAJA_INLINE
```
MrBurmark: spacing

include/RAJA/util/View.hpp
```diff
@@ -235,6 +238,38 @@ struct MultiView
...
+  constexpr linear_index_type size() const { return layout.size(); }
```
MrBurmark: We probably need to multiply by the size of the pointer dimension.

test/unit/view-layout/test-subview.cpp
```diff
@@ -0,0 +1,497 @@
...
+      view.get_data(), SubLayoutType(view.get_layout(), slices...));
+}
+
+struct UseViewWithSubLayout
+{
+
+  template<typename ViewType, typename... Slices>
+  auto operator()(ViewType& view, Slices... slices) const
+  {
+    return make_view_with_sublayout(view, slices...);
```
MrBurmark: If this is preferred, can we just call this one `make_subview`?

test/unit/view-layout/test-subview.cpp
```diff
@@ -0,0 +1,497 @@
...
+  }
+};
+
+struct UseSubViewWithLayout
+{
+
+  template<typename ViewType, typename... Slices>
+  auto operator()(ViewType& view, Slices... slices) const
+  {
+    return make_subview_with_layout(view, slices...);
```
MrBurmark: If this is not preferred, perhaps we don't have a helper and just make people compose things more like this `make_view(make_sublayout(...))`?

test/unit/view-layout/test-subview.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+  template<typename ViewType>
+  static auto& get_subregion(ViewType& sv)
+  {
+    return sv;
+  }
+};
+
+template<typename Factory>
+class SubViewTest : public ::testing::Test
```
MrBurmark, line 64: Do we want to have subview of multiview tests? Perhaps we could split the subview of multiviews into a later MR?

docs/sphinx/user_guide/feature/view.rst
MrBurmark: I saw the deduction guide, should we use that here?

include/RAJA/util/SubView.hpp
MrBurmark: worry about this in a separate concepts pass.

include/RAJA/util/SubView.hpp
MrBurmark: done

include/RAJA/util/SubView.hpp
MrBurmark: defer to concepts pass.

include/RAJA/util/SubView.hpp
MrBurmark: defer to concepts pass.

include/RAJA/util/SubView.hpp
MrBurmark: defer to concepts pass.

include/RAJA/util/SubView.hpp
MrBurmark: done.

include/RAJA/util/SubView.hpp
MrBurmark: done

include/RAJA/util/SubView.hpp
MrBurmark: renamed to parent and subregion for clarity.

include/RAJA/util/TypedViewBase.hpp
MrBurmark: I removed get_parent_dim_stride and changed get_dim_stride to use the linear index stride in the original view's region instead of the stride in the subregion.

include/RAJA/util/SubView.hpp
MrBurmark: removed

test/unit/view-layout/test-subview.cpp
MrBurmark: split subviews of multiviews into a separate PR.

include/RAJA/util/SubView.hpp
MrBurmark: treating as UB, we can circle back on this later and add a check when bounds checking is enabled or similar.

include/RAJA/util/View.hpp
MrBurmark: removing these functions for now.

docs/sphinx/user_guide/feature/view.rst
MrBurmark: added a helper function.

test/unit/view-layout/test-subview.cpp
MrBurmark: done.

test/unit/view-layout/test-subview.cpp
MrBurmark: done.

MrBurmark:
@gberg617 I made a few changes based on my comments. Hopefully this is still what you had in mind. This requires a fix in camp that is coming soon so expect test failures until that comes in.

MrBurmark:
@gberg617 Take a look, I updated camp so you should be able to write to data using a subview now.

MrBurmark:
It looks like tests are failing because they are using camp 2026.7.0 instead of the updated camp.

rhornung67:
@smithsg84 it looks like we are in the state where we need to test against a camp update that requires the Spack package hack/magic. Do you have notes on how to do this?

include/RAJA/util/SubView.hpp
```diff
@@ -0,0 +1,443 @@
...
+
+  template<size_t SubregionDim>
+  RAJA_INLINE RAJA_HOST_DEVICE constexpr IndexLinear get_dim_begin() const
+  {
+    static_assert(SubregionDim < n_dims, "Dimension out of bounds");
+    return IndexLinear(0);
+  }
+
+  /*!
+   * \brief Access the parent using indices mapped from the subregion index
```
gberg617: Maybe we should rename "subregion" variables to something else now?

MrBurmark: changed to adapter.

include/RAJA/util/SubView.hpp
MrBurmark: You mean like in offset view?

rhornung67:
@johnbowen42 is the idea here that the RADIUSS spack configs stuff will be removed in another PR after this is merged?

johnbowen42:
> @johnbowen42 is the idea here that the RADIUSS spack configs stuff will be removed in another PR after this is merged?

All changes to `scripts/radiuss-spack-configs` and the uberenv json should be rolled back now.  The only CI related change in this PR should now be in `.gitlab-ci.yml`.  Let me know if I have missed anything

rhornung67:
> > @johnbowen42 is the idea here that the RADIUSS spack configs stuff will be removed in another PR after this is merged?
> 
> All changes to `scripts/radiuss-spack-configs` and the uberenv json should be rolled back now. The only CI related change in this PR should now be in `.gitlab-ci.yml`. Let me know if I have missed anything

Yup. Looks good. I was looking at an old link to this PR.

MrBurmark:
@johnbowen42 I updated camp again on this branch, what decides which camp is used in CI now?

johnbowen42:
> @johnbowen42 I updated camp again on this branch, what decides which camp is used in CI now?
@MrBurmark :
https://github.com/llnl/RAJA/blob/03c871fc78fc6b06ac172efc050e64f47aa29220/.gitlab-ci.yml#L46

MrBurmark:
> > @johnbowen42 I updated camp again on this branch, what decides which camp is used in CI now?
> > @MrBurmark :
> > https://github.com/llnl/RAJA/blob/03c871fc78fc6b06ac172efc050e64f47aa29220/.gitlab-ci.yml#L46

Ok, so we have to update that when we update the submodule? It can't query the submodule?

johnbowen42:
> > > @johnbowen42 I updated camp again on this branch, what decides which camp is used in CI now?
> > > @MrBurmark :
> > > https://github.com/llnl/RAJA/blob/03c871fc78fc6b06ac172efc050e64f47aa29220/.gitlab-ci.yml#L46
> 
> Ok, so we have to update that when we update the submodule? It can't query the submodule?

There isn't a way that I know of

johnbowen42:
> > > > @johnbowen42 I updated camp again on this branch, what decides which camp is used in CI now?
> > > > @MrBurmark :
> > > > https://github.com/llnl/RAJA/blob/03c871fc78fc6b06ac172efc050e64f47aa29220/.gitlab-ci.yml#L46
> > 
> > 
> > Ok, so we have to update that when we update the submodule? It can't query the submodule?
> 
> There isn't a way that I know of

Looking into a way of doing it in the `build_and_test.sh` job script

rhornung67:
@johnbowen42 @MrBurmark if I correctly understand how this works, the Camp SHA value that the `CAMP_REFERENCE` variable is set to in the `RAJA/.gitlab-ci.yml` file is what is used in GitLab CI builds if it is set. Otherwise, it will use the Camp version specified in the RAJA Spack package in the Spack packages repo on GitHub.

See here https://github.com/llnl/RAJA/blob/develop/scripts/gitlab/build_and_test.sh#L52 and here https://github.com/llnl/RAJA/blob/develop/scripts/gitlab/build_and_test.sh#L282

So I believe that all we need to do to make sure a RAJA PR branch is being tested with the proper Camp version is to set the camp reference to what we want in the RAJA branch. 

We should also make sure that this is consistent with the camp submodule version of the branch so that local testing done using one of our build scripts also uses the right camp version.

@MrBurmark to answer your question above, updating the RAJA camp submodule is insufficient.

MrBurmark:
Maybe there should be a check in the script that the CI camp version and the submodule camp version are the same or "compatible" for some definition of compatible if we want to allow different versions.

rhornung67:
I think the issue with that is that the process that Spack uses to configure a RAJA build does not pull the submodule(s) when it clones RAJA https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/raja/package.py#L24 So there is no way to check the camp submodule SHA.

After we get the releases out, we will disentangle this.

PR #1917, opened 2025-09-21
Matrix Transpose Tutorial Cleanup

### LINKED ISSUE #1916: Matrix Transpose Tutorial Cleanup
I found a couple things while looking at the transpose tutorial.

First, the launch and kernel solutions could use block_unchecked policies. This will also allow the kernel implementation to skip the second sync threads call.

Second, it doesn't look like the solutions don't actually use shared memory as intended. It looks like the same thread that reads a value writes that value. The intention of shared memory is to let different threads read and write so memory accesses to both matrices are coalesced. This will require the launch solution to have a teamSync call, which it is currently lacking.

MrBurmark:
# Summary

Fix the cuda and hip matrix tutorial. Fix spacing, add proper synchronization, map threads properly in teams implementation.

- This PR is a bugfix
  - Fixes #1916

### DIFF
--- a/exercises/launch-matrix-transpose-local-array_solution.cpp
+++ b/exercises/launch-matrix-transpose-local-array_solution.cpp
@@ -308,6 +308,8 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
             });
           });
 
+          ctx.teamSync();
+
           RAJA::loop_icount<cuda_threads_x>(ctx, col_tile, [&] (int col, int tx) {
             RAJA::loop_icount<cuda_threads_y>(ctx, row_tile, [&] (int row, int ty) {
 
@@ -381,6 +383,8 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
 
             });
           });
+
+          ctx.teamSync();
 
           RAJA::loop_icount<hip_threads_x>(ctx, col_tile, [&] (int col, int tx) {
             RAJA::loop_icount<hip_threads_y>(ctx, row_tile, [&] (int row, int ty) {
--- a/exercises/launch-matrix-transpose-local-array_solution.cpp
+++ b/exercises/launch-matrix-transpose-local-array_solution.cpp
@@ -278,8 +278,8 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
   const int n_blocks_c = RAJA_DIVIDE_CEILING_INT(N_c, c_block_sz);
   const int n_blocks_r = RAJA_DIVIDE_CEILING_INT(N_r, r_block_sz);
 
-  using cuda_teams_y = RAJA::LoopPolicy<RAJA::cuda_block_y_direct>;
-  using cuda_teams_x = RAJA::LoopPolicy<RAJA::cuda_block_x_direct>;
+  using cuda_teams_y = RAJA::LoopPolicy<RAJA::cuda_block_y_direct_unchecked>;
+  using cuda_teams_x = RAJA::LoopPolicy<RAJA::cuda_block_x_direct_unchecked>;
 
   using cuda_threads_y = RAJA::LoopPolicy<RAJA::cuda_thread_y_direct>;
   using cuda_threads_x = RAJA::LoopPolicy<RAJA::cuda_thread_x_direct>;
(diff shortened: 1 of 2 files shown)


exercises/launch-matrix-transpose-local-array_solution.cpp
```diff
@@ -302,16 +308,18 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
             });
           });
 
-         RAJA::loop_icount<cuda_threads_x>(ctx, col_tile, [&] (int col, int tx) {
-           RAJA::loop_icount<cuda_threads_y>(ctx, row_tile, [&] (int row, int ty) {
+          ctx.teamSync();
 
-             Atview(col, row) = Tile_Array[ty][tx];
+          RAJA::loop_icount<cuda_threads_y>(ctx, col_tile, [&] (int col, int tx) {
+            RAJA::loop_icount<cuda_threads_x>(ctx, row_tile, [&] (int row, int ty) {
```
artv3, line 314: The magic is here!

artv3:
Thanks for getting this started @MrBurmark. I took a follow up pass; but I think I could use a second set of eyes on the Kernel implementation. Would anyone in @LLNL/raja-core  be able to take a look?

exercises/launch-matrix-transpose-local-array_solution.cpp
```diff
@@ -374,20 +384,23 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
...
 
-          RAJA::loop_icount<hip_threads_x>(ctx, col_tile, [&] (int col, int tx) {
-           RAJA::loop_icount<hip_threads_y>(ctx, row_tile, [&] (int row, int ty) {
+          ctx.teamSync();
+
+          RAJA::loop_icount<hip_threads_y>(ctx, col_tile, [&] (int row, int ty) {
+            RAJA::loop_icount<hip_threads_x>(ctx, row_tile, [&] (int col, int tx) {
 
-             d_Atview(col, row) = Tile_Array[ty][tx];
+              d_Atview(row, col) = Tile_Array[tx][ty];
```
artv3, line 392: @MrBurmark , I switched it around so its clear that the x and y threads have been transposed in shared memory. I'm not too sure how to express that in Kernel.

rhornung67:
@artv3 I think  you've taken over this PR. There are come compilation issues that are causing the CI checks to fail. Can you get to them or should I try to find time to fix them? I assume we want this in the release.

artv3:
> @artv3 I think you've taken over this PR. There are come compilation issues that are causing the CI checks to fail. Can you get to them or should I try to find time to fix them? I assume we want this in the release.

I can take a look, but I don't think this should hold up the release.

rhornung67:
@MrBurmark @artv3 I fixed the compilation errors in this, but the results are wrong.....

rhornung67 (changes requested):
The code generates the wrong results. Either the code is wrong or the check result method is wrong. I haven't dug into this.

There are also other changes that could improve this and remove some confusion. For example, the printResult() routine assumes that the matrix passed in is `At` but the first call in the code (that can be commented out) is `A`. I would print out which matrix is being printed in a statement before each printResult() call and make the print method not name the matrix.

exercises/launch-matrix-transpose-local-array_solution.cpp
rhornung67: using row_t and col_t was the source of the compilation error. I switched it back so code compiles. Please feel free to undo my change and fix the compile differently.

artv3:
I think we also have some inconsistent expression for the CPU RAJA::kernel examples. That is one thing that I needed to revisit.

rhornung67:
If I can find a few spare minutes, I will work on it. But, I can't promise I will get to it.

artv3:
@llnl/raja-core this may need another review.

exercises/launch-matrix-transpose-local-array_solution.cpp
```diff
@@ -248,10 +252,10 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
             });
           });
 
-          RAJA::loop_icount<loop_pol_2>(ctx, col_tile, [&] (int col, int tx) {
-            RAJA::loop_icount<loop_pol_2>(ctx, row_tile, [&] (int row, int ty) {
+          RAJA::loop_icount<loop_pol_2>(ctx, col_tile, [&] (int row_t, int ty) {
```
MrBurmark, line 255: Do you want to add the sync for all of the policy cases?

PR #1925, opened 2025-10-02
cpp 17: policy/openmp/params/forall.hpp simplifications

johnbowen42:
This PR simplifies the ChunkSize overloads inside  policy/openmp/params/forall.hpp  with if constexpr


include/RAJA/policy/openmp/params/forall.hpp
```diff
@@ -287,46 +227,21 @@ RAJA_INLINE void forall_impl(const ::RAJA::policy::omp::Guided<ChunkSize>& p,
 
     using RAJA::internal::thread_privatize;
     auto body = thread_privatize(loop_body);
-
-#pragma omp for schedule(guided) reduction(combine : f_params)
-    for (decltype(distance_it) i = 0; i < distance_it; ++i)
+    if constexpr (ChunkSize <= 0)
```
adayton1, line 230: Can you pick a consistent ordering? In some forall implementations, you do the ChunkSize > 0 case first and in others the ChunkSize <= 0 first.

adayton1 (approved):

include/RAJA/policy/openmp/params/forall.hpp
johnbowen42: yes, good point.  I will do that

artv3:
@johnbowen42 , not sure if this might be out of the scope of this PR. But there may also be a lot of duplicate capability here: 

https://github.com/LLNL/RAJA/blob/develop/include/RAJA/policy/openmp/forall.hpp

PR #1957, opened 2025-12-12
cmake: Move minimum C++ standard to 20

johnbowen42:
This pull request changes the minimum C++ version in CMake and build scripts to C++20.  We probably want to hold off on merging this change until after the upcoming release is cut.


rhornung67 (approved):
@johnbowen42 we cannot force C++20 yet since it that would break some of our WSC user codes.

rhornung67 (commented):
Let's hold off on merging this and reassess after the release.

rhornung67 (changes requested):
Trying to dismiss approval

kab163:
Umpire has a similar PR - we can reassess after the holidays on when we want to start enforcing c++20 minimum requirement

johnbowen42:
@llnl/raja-core can I get a review on this PR now that we are ready to move forward with C++ 20?

MrBurmark (approved):

artv3 (approved):
Ready!

host-configs/lc-builds/toss4/corona_sycl.cmake
```diff
@@ -12,12 +12,8 @@ set(RAJA_COMPILER "RAJA_COMPILER_CLANG" CACHE STRING "")
...
-set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -g -L${SYCL_LIB_PATH} -fsycl -fsycl-unnamed-lambda -fsycl-targets=amdgcn-amd-amdhsa -Xsycl-target-backend --offload-arch=gfx906" CACHE STRING "")
-set(CMAKE_CXX_FLAGS_DEBUG "-O0 -g -L${SYCL_LIB_PATH} -fsycl -fsycl-unnamed-lambda -fsycl-targets=amdgcn-amd-amdhsa -Xsycl-target-backend --offload-arch=gfx906" CACHE STRING "")
-#set(CMAKE_CXX_FLAGS_RELEASE "-O3 -std=c++17 -fsycl -fsycl-unnamed-lambda --gcc-toolchain=/usr/tce/packages/gcc/gcc-7.1.0" CACHE STRING "")
-#set(CMAKE_CXX_FLAGS_RELWITHDEBINFO " -O3 -g  -std=c++17 -fsycl -fsycl-unnamed-lambda --gcc-toolchain=/usr/tce/packages/gcc/gcc-7.1.0" CACHE STRING "")
-#set(CMAKE_CXX_FLAGS_DEBUG " -O0 -g -std=c++17 -fsycl -fsycl-unnamed-lambda --gcc-toolchain=/usr/tce/packages/gcc/gcc-7.1.0" CACHE STRING "")
-#set(CMAKE_CXX_LINK_FLAGS "-fsycl -Wl,-rpath,/usr/tce/packages/oneapi/oneapi-2021.2/compiler/2021.2.0/linux/compiler/lib/intel64_lin/"  CACHE STRING "")
+set(CMAKE_CXX_FLAGS_RELEASE "-O3 -std=c++20 -L${SYCL_LIB_PATH} -fsycl -fsycl-unnamed-lambda -fsycl-targets=amdgcn-amd-amdhsa -Xsycl-target-backend --offload-arch=gfx906" CACHE STRING "")
```
rhornung67: I don't think we need to specify the c++ std in the host-config. It's specified on the CMake line in the invoking script via BLT_CXX_STD.  Putting it in the host-config is inconsistent with  the others, opens the door for conflicting specification, and makes it hard to change when trying different things during development.

CMakeLists.txt
```diff
@@ -75,8 +72,9 @@ if(NOT DEFINED BLT_CXX_STD)
 else() #check BLT_CXX_STD is high enough by disallowing the only invalid option
   if(("${BLT_CXX_STD}" STREQUAL "c++98") OR
      ("${BLT_CXX_STD}" STREQUAL "c++11") OR
-     ("${BLT_CXX_STD}" STREQUAL "c++14"))
-    message(FATAL_ERROR "RAJA requires minimum C++ standard of c++17")
+     ("${BLT_CXX_STD}" STREQUAL "c++14") OR
```
rhornung67, line 75: Is there a way to do this in CMake with a '<=' type syntax?  Meaning a single statement that says the equivalent of std <= 20.

Or, should we move to something like this?
`set(CMAKE_CXX_STANDARD 20)
set(CMAKE_CXX_STANDARD_REQUIRED ON)`

That probably conflicts with what BLT does or may introduce an ordering dependency based on where the standard is set?!?

rhornung67 (changes requested):
Remove the c++ std from corona sycl host-config

CMakeLists.txt
adayton1: I think the CMake way of specifying the minimum standard is the following (this was suggested in the CMake class offered to LLNL last year):

`target_compile_features(RAJA PUBLIC cxx_std_20)`

The PUBLIC means that downstream targets that include RAJA also need to have a minimum c++ standard of c++20.

smithsg84: That could require a bit of cleverness.   98 (1998) will mess up simple extract year and LESS_EQUAL compare.   

This would be a bit shorter than the OR sequence.

set(invalid_standards c++98 c++11 c++14 C++17)
if(BLT_CXX_STD IN_LIST invalid_standards)

rhornung67: Thanks @adayton1 I did not know that was the cmake idiomatic way.

tomstitt (approved):

CMakeLists.txt
trws: FWIW, we originally had it written using the compile features check, but I ended up adding the extra logic because we regularly had issues with cmake claiming a given compiler didn't support a version when we knew it did. If that problem has gone away, then we should go back to doing it the idiomatic way.

smithsg84 (approved):
Looks OK to me.

CMakeLists.txt
johnbowen42: I looked into this, and `target_compile_features(RAJA PUBLIC cxx_std_20)` will always build RAJA with that standard, and will silently ignore any attempts by users to set a different standard either through BLT or through `CMAKE_CXX_STANDARD`.  This risks some confusion, where a user might try to create a build of RAJA with an older standard and end up with 20 anyway.  Despite this, I prefer this approach because it rigidly enforces the standard alongside the sha, whereas before a user needs to know how to map RAJA version number-->acceptable CXX version or get build configuration errors.  What does everybody here think?

host-configs/lc-builds/toss4/corona_sycl.cmake
johnbowen42: fixed

rhornung67 (approved):

CMakeLists.txt
MrBurmark: I don't think we want to force people into a specific c++ version. Especially if they can think they're setting one but actually getting a different one, which sounds like a nightmare for our library build maintainers.

adayton1: I didn't realize it would silently ignore CMAKE_CXX_STANDARD in this case. I don't really like that. If you set BLT_CXX_STD to c++23, does it still use c++20 or respect the higher standard?

johnbowen42: Just to push back on that a little bit--once we introduce C++20 concepts or other C++20 constructs into develop, we are forcing people into a specific C++ version regardless of this specific cmake command.  As soon as we start actively developing C++20 features, the code will not compile with the standard set to anything older.  

Maybe instead of silently using a newer standard we can add a warning if a user specifies an older version, either though `CMAKE_CXX_STANDARD` or `BLT_CXX_STD`.

johnbowen42: > I didn't realize it would silently ignore CMAKE_CXX_STANDARD in this case. I don't really like that. If you set BLT_CXX_STD to c++23, does it still use c++20 or respect the higher standard?

I will investigate

MrBurmark: I think we're fine with it failing to configure if you set a incompatible c++ standard, but we don't want to stop people from setting a higher c++ standard.

MrBurmark: One of the annoying things we ran into when trying to test out new c++ versions is libraries hard coding a specific c++ version and then we have to try to patch their config files instead of being able to just pass an option.

rhornung67: Whatever we can do to make it clear and explicit for users what we require and and when they something inconsistently is best, I think.  Silently changing something that they specified and which will not work is not a good thing to do, IMO.

johnbowen42: You can set the version to 23, either through `CXX_CMAKE_STANDARD` or through `BLT_CXX_STD`, and CMake will respect that.  On the other hand, CMake will set the version to 20 if a user specifies lower, like 17, 14, etc.   My personal preference would be to print a warning and continue with the correct version if a user specifies a lower version, but it sounds like everybody wants the build to fail instead with a too-low CMAKE_CXX_VERSION, is that right?

MrBurmark: I would prefer a config time fail over trying to figure out why the library builds but code using the library does not.

adayton1: I would prefer it to fail if the version is too low.

adayton1: Another thing to be aware is that if we use cxx_std_20 as a PUBLIC target feature, downstream libraries will also be affected - meaning if they set the standard too low on their project, that will also be ignored and silently updated to c++20. So now that we've talked this through, I think it cxx_std_20 should either not be used or set to PRIVATE so that it only affects RAJA and not downstream projects.

smithsg84: Another vote for failing on version too low.   Silent magic is bad.

rhornung67: Ditto for me -- fail when unsupported version is specified.

adrienbernede: Same here, fail when unsupported version is specified, ability to specify via BLT_CXX_STD.
Also, there have been efforts from folks like Daniel Taller to be able to enforce cxx standard via Spack. If the cxx standard specified as a variant in a spack spec actually resulted in another standard being use, that would be calling for troubles.

johnbowen42:
Shoutout @bechols97  for fixing the windows build in CI!

PR #1967, opened 2025-12-23
Bump cmake min required version

rhornung67:
# Summary

- This PR bump CMake min required version to 3.24.
- Newer compilers and language standards, such as C++20, work better with newer CMake versions.
- LC systems support newer versions
- CMake 3.24 is the lowest version supported in images we use for GH Actions testing


artv3 (approved):
sounds good to me!

RELEASE_NOTES.md
```diff
@@ -20,6 +20,21 @@ Notable changes include:
...
+This release contains ...
+
+Notable changes include:
+
+  * New features / API changes:
+
+  * Build changes/improvements:
+    * Bump minimum CMake version required to 3.24.
+
+  * Bug fixes/improvements:
```
artv3, line 36: I'm a big fan of updating the release_notes / change log per PR!

rhornung67: Ditto. If we are disciplined enough to check on  that for each PR review, we can do it.

rhornung67:
Folks, the CI failures here are due to container images still being updated. I will re-trigger CI when it looked like the new versions are ready.

adayton1 (approved):

RELEASE_NOTES.md
smithsg84: Just FYI, a  check can be added to verify the release notes / changlog file has been changed.   Used this on a prior project to remind people to update. Happy to add that if you want.   It was both useful and annoying at times.

rhornung67: Ooooo, that sounds good. Please make a PR or whatever needs to be done to add that.

PR #1968, opened 2025-12-23
Add workflow to check if PRs have updated the release notes

smithsg84:
Add a workflow check to verify that the RELEASE_NOTES.md file has been edited.   Release notes should be updated with a summary of what the PR does.

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -1,0 +1,19 @@
+name: Run PR Checks
+
+on:
+  pull_request:
+    branches:
+      - develop # Branch to run checks on
+
+jobs:
+  # Verify that release notes have been updated.   PRs should include summary of
+  # what has changed in the release notes.
+  check_release_notes:
+    runs-on: ubuntu-latest
+    name: Verify RELEASE_NOTES.md has been updated
+    steps:
+      - name: Verify RELEASE_NOTES.md updated
+        uses: takanuva15/verify-file-updated@v1
+        with:
+          filename_to_check: RELEASE_NOTES.md # Check the RELEASE_NOTES.md for change
+          excused_label: "no release notes check" # Optional: add this label to PRs to skip the check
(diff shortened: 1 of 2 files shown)


rhornung67:
The check failure seems to apply that this works! 😄

smithsg84:
> The check failure seems to apply that this works! 😄

Yes it looks to be working.   Will see if updating release notes works to fix the failure.

rhornung67:
It would be cool if it checked the release notes update against the git commit messages 😜

smithsg84:
That is probably coming, a test or two of having claude summarize a PR was interesting.   It did catch that new input options were available, the technical details but didn't really set a context for why the change would be useful.

rhornung67:
Just having the CI check fail when the release notes have not been updated should be sufficient to make folks aware.

However, this raises an interesting topic for team discussion. IMO, the release notes should only document changes that are important for users to know. Every detail of the commit history leading to a release can be information overload which would likely discourage users from reading the release notes. For example, some projects use the GitHub feature that automatically generates release notes from the git log and cites every commit. I summarize the important changes in each PR merge when I manually write release notes and omit items that are only of interest to developers, such as internal implementation changes.

Does it make sense to allow ignoring the check by having the author of a PR insert a special comment in the release description?

smithsg84:
The github action does support a github Label we can specify to ignore the check.  We can specify the label name used, the excused_label attribute:

```
excused_label: "no release notes check" # Optional: add this label to PRs to skip 
```

So we can label PR's that don't need a Release Notes check.

For what it's worth, I don't like the "dump git log to release notes" method.   That info already exists in the log so why repeat in a file?  

On PF to help avoid the overload issue I split comments into sections like User/Developer.   This seems to helped.

A little discussion sounds good.   Not sure what to do about PR's that really don't need to be in the release notes.   We ran into this same issue on a global security project.   I'd insert/remove an extra blank line but that is not a super satisfactory method.

PR #1969, opened 2025-12-26
Add kahan sum reduce helper class.

MrBurmark:
# Summary

This adds a KahanSum class to help do Kahan sums.
This also adds tests and generalizes the existing tests for reduce helper classes to accommodate slightly different interfaces.

The intention is to potentially use this or something similar for checksums in https://github.com/llnl/RAJAPerf/pull/608.

- This PR is a  feature
- It does the following:
  - Adds KahanSum class at the request of myself


artv3:
Super nice! Can you add something in the RAJA docs about this new feature? And what do folks think about adding a note about this in our `RELEASE_NOTES` in this PR.

MrBurmark:
Actually we should probably not hide the KahanSum class in RAJA::detail if we want to use it in RAJAPerf.

MrBurmark:
@artv3 I'll add an issue to add documentation, but I'd like to get this in so it can be used in https://github.com/llnl/RAJAPerf/pull/608.

include/RAJA/policy/cuda/multi_reduce.hpp
```diff
@@ -285,8 +285,8 @@ struct MultiReduceGridAtomicHostInit_TallyData
   //! get value for bin, assumes synchronization occurred elsewhere
   T get(int bin) const
   {
-    ::RAJA::detail::HighAccuracyReduce<T, typename Combiner::operator_type>
-        reducer(m_identity);
+    ::RAJA::HighAccuracyReduce<T, typename Combiner::operator_type> reducer(
```
artv3, line 288: Why not just call it `KahanReduce`? What if we want to add other types of High Accuracy Reducers

MrBurmark: I called it sum because it is always a sum and can't do generic operators.

test/unit/algorithm/tests/test-algorithm-util-reduce.hpp
```diff
@@ -199,6 +273,11 @@ using HipAccumulateReduceReducers =
               Accumulate<test_hip>
             >;
 
+using HipKahanReduceReducers =
+  camp::list<
+              KahanSum<test_hip>
+            >;
```
artv3, line 279: I misread it, the policy name is KahanSum.

artv3 (approved):

artv3 (approved):

PR #1970, opened 2025-12-30
proteus: enable JIT compilation of RAJA kernels using Proteus

johnbowen42:
This PR is a draft demonstrating integration of JIT compilation with Proteus into RAJA.  Here are a couple of things to note about this PR:

- The CMake changes only support using Proteus via an installation, which must be provided to RAJA at build configuration time
- Depending on the configuration of the install of Proteus, it will be necessary to have LLVM shared libraries visible to the linker
- We can configure Proteus to be an optional submodule of RAJA as well, similar to CHAI/camp/umpire
- The current setup enables JIT at build config time.  This means by default, with Proteus enabled, all RAJA kernel/launch/forall calls will be JIT compiled.  We want to iterate on this by exposing JIT as an option that can be specified by the user


include/RAJA/policy/cuda/kernel/CudaKernel.hpp
```diff
@@ -683,7 +685,7 @@ struct StatementExecutor<
         auto cuda_data = RAJA::cuda::make_launch_body(
             func, launch_dims.dims.blocks, launch_dims.dims.threads, shmem, res,
             data);
-
+        RAJA::register_lambda(func);
```
MrBurmark: Would this be more clear?
```suggestion
        RAJA::JIT::register_lambda(func);
```

johnbowen42: I do like this more.  However, the namespace JIT would need to be visible even when JIT is not enabled through our build system.  Do you think that's acceptable?

rhornung67: What about making a compile-time selectable namespace alias, something like this in the appropriate header file:
`#if defined(RAJA_ENABLE_JIT)`
 `    namespace Foo = RAJA::JIT;`
 `#else`
 `    namespace Foo = RAJA;`
 `#endif`

Then, in the code `Foo::register_lambda(func);`

Obviously, we would want a better name than `Foo`....

include/RAJA/policy/cuda/forall.hpp
```diff
@@ -381,10 +382,11 @@ template<typename EXEC_POL,
                               (IterationGetter::block_size > 0),
                           size_t> BlockSize = IterationGetter::block_size>
 __launch_bounds__(BlockSize, BlocksPerSM) __global__
-    void forallp_cuda_kernel(const RAJA_CUDA_GRID_CONSTANT LOOP_BODY loop_body,
-                             const RAJA_CUDA_GRID_CONSTANT Iterator idx,
-                             const RAJA_CUDA_GRID_CONSTANT IndexType length,
-                             ForallParam f_params)
+    RAJA_JIT_COMPILE_ARGS(3) void forallp_cuda_kernel(
```
rhornung67, line 385: What does `3` mean here?

It's better not to sprinkle magic numbers in the code and use a label in the form of a macro constant or constexpr variable.

Same comment applies to several places below.

include/RAJA/policy/cuda/kernel/CudaKernel.hpp
trws: FWIW, I'd recommend making it part of an internal namespace unless the intent is for users to call that function. Literally `RAJA::internal::<whatever>` or similar, since we already have those and this isn't something we likely want users to touch.

include/RAJA/policy/cuda/forall.hpp
trws: IIRC this is how proteus gets the number of arguments to JIT based on, so it isn't really a "magic number" per se but based on the signature and how it's intended to be used.

johnbowen42: Yes--3 corresponds to the 1-indexed argument to specialize the function's IR against

include/RAJA/policy/cuda/kernel/CudaKernel.hpp
johnbowen42: I like the idea of hiding it inside `internal`, maybe even under the nested `RAJA::internal::JIT` namespace

include/RAJA/policy/cuda/forall.hpp
rhornung67: Got it. Thank you.

artv3:
Can we add some testing and examples? Documentation would also be nice

johnbowen42:
> Can we add some testing and examples? Documentation would also be nice

I've been thinking about this--I can definitely add examples, but for testing we would basically run all the same tests we run with RAJA but run it through Proteus.  I could for example add a pipeline that runs on Tioga and does this.  What do we think about that? (The only drawback I can think of this extra CI resources)

johnbowen42:
I added a CI pipeline that builds RAJA with Proteus and runs all of RAJA's tests through JIT compilation

johnbowen42:
I added https://github.com/llnl/RAJA/pull/1970/changes#diff-71be002f5ad4c8d7298e0492f85f6c40c6002c6203cfb7bb85b3897682d04954 which demonstrates branch elimination and loop scheduling optimizations with JIT

johnbowen42:
I made some modifications to our CI to limit the number of tests run through proteus in order to ensure CI passes.  Previously, proteus was timing out CI due to the overhead of basically recompiling every test while running it without any cache. Currently, we just run every forall, launch and kernel test

include/RAJA/util/macros.hpp
```diff
@@ -224,4 +224,15 @@ RAJA_HOST_DEVICE inline void RAJA_ABORT_OR_THROW(const char* str)
 
 #endif
 
+#if defined RAJA_ENABLE_JIT
+// This macro accepts a list of integers as its arguments, corresponding to the
+// 1-indexed arguments for which to specialize a function's IR.
+#define RAJA_JIT_COMPILE_ARGS(...) __attribute__((annotate("jit", __VA_ARGS__)))
+#define RAJA_JIT_COMPILE           __attribute__((annotate("jit")))
```
MrBurmark, line 231: Is there a reason to prefer this over the c++ attribute syntax?

examples/forall-jit.cpp
```diff
@@ -0,0 +1,116 @@
...
+    }
+  });
+  aot_timer.stop();
+  double t = aot_timer.elapsed();
+  std::cout << "aot total time = " << t << "\n";
+  RAJA::Timer jit_timer;
+  jit_timer.start();
+
+  RAJA::forall<policy>(RAJA::RangeSegment(0, N), [=,
+    a = proteus::jit_variable(a), b = proteus::jit_variable(b)] (int i) {
```
MrBurmark: I notice that there is a version of jit_variable in an internal RAJA namespace. Does it make sense to use the proteus version here or should RAJA expose its version?

examples/forall-jit.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include "RAJA/RAJA.hpp"
+#include "RAJA/util/Timer.hpp"
+#include "proteus/JitInterface.h"
+
+// layout i x a x b
+#define A(i, r, c) A[i * a * b  + b * r + c]
+// layout i x b x a
+#define B(i, r, c) B[i * a * b  + a * r + c]
+// layout i x a x a.  matrix products of A[i]B[i]
+#define C(i, r, c) C[i * a * b  + a * r + c]
```
MrBurmark: what are the dimensions of c?

include/RAJA/util/macros.hpp
johnbowen42: We could leave the annotations in the C++ code and it should be fine, but this way RAJA is unmodified in the case where no JIT is enabled

examples/forall-jit.cpp
johnbowen42: great point, I will change that

examples/forall-jit.cpp
johnbowen42: good catch

examples/forall-jit.cpp
johnbowen42: I kept it `proteus::jit_variable` because it works better for our analysis, and I removed the RAJA namespace version

include/RAJA/util/macros.hpp
trws: I'm pretty sure @MrBurmark meant using `__attribute__` instead of something like `[[proteus::annotate("jit", ...)]]` inside the macro.  I have actually been wondering why this syntax rather than the one that can appertain directly to arguments as well, but I'm not sure proteus currently supports that.

johnbowen42: I see, thanks for clarifying.  The C++ attribute syntax is 1-1 equivalent to the `__attribute__` syntax

ggeorgakoudis: > I'm pretty sure @MrBurmark meant using `__attribute__` instead of something like `[[proteus::annotate("jit", ...)]]` inside the macro. I have actually been wondering why this syntax rather than the one that can appertain directly to arguments as well, but I'm not sure proteus currently supports that.

That's a very good question on annotating parameters. It's in our scope of future work. We analyze annotations at the LLVM IR level and annotating a parameter requires  (expectedly simple) dataflow analysis to match the emitted annotation intrinsics with the parameter and is more sensitive to optimization interference. We are fairly confident we can tackle those.

Despite syntax elegance, do you folks see more functionality/usability supporting parameter annotations? I would like to gather more motivation to put resources to have it done with higher priority.

trws: We should probably talk sometime, I've been mulling on it since Bronis and I presented some of this for the DOE/MEXT meeting.  The main quick one is that it would make it easier to support doing JIT on parameters other than leading ones, or in variadic template functions where the number is unknown at the declaration. It would also make it more obvious to a reader what's going on than having a count from start.  We try to stick to the lambda at the end convention, but for example if you had something like this it would be useful:

```c++
template<typename F, typename... Args>
void fn_with_trailing_dynamic_args([[proteus::jit]] int range, F func, [[proteus::jit]] Args const &args)
```

examples/forall-jit.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include "RAJA/RAJA.hpp"
+#include "RAJA/util/Timer.hpp"
+#include "proteus/JitInterface.h"
+
+// layout i x a x b
+#define A(i, r, c) A[i * a * b  + b * r + c]
+// layout i x b x a
+#define B(i, r, c) B[i * b * a  + a * r + c]
+// layout i x a x a.  matrix products of A[i]B[i]
+#define C(i, r, c) C[i * a * a + a * r + c]
```
artv3: Maybe use the RAJA::Views here? There is helper function `make_permuted_view` that is handy

johnbowen42: done

examples/forall-jit.cpp
```diff
@@ -0,0 +1,121 @@
...
+    b = proteus::jit_variable(b),
+    accum = proteus::jit_variable(accum)
+  ]  (int i) RAJA_JIT_COMPILE {
+    for (int row = 0; row < a; ++row){
+      for (int col = 0; col < b; ++col) {
+        if (accum) {
+          C(i, row, col) = A(i, row, col) * B(i, col, row);
+        }
+        else {
+          C(i, row, col) = A(i, row, col) * B(i, col, row) + C(i, row, col);
```
MrBurmark:
```suggestion
          C(i, row, col) += A(i, row, col) * B(i, col, row);
```
It would be easier on the eye to write it this way. Did you write it the other way to try to get FMAs?

johnbowen42: I wrote it the other way because I was using the macro and not the views with `operator+=` defined

johnbowen42: hopefully HIP's FMA pattern recognition would recognize both of those!

johnbowen42: committed this change

examples/forall-jit.cpp
```diff
@@ -20,15 +20,9 @@
 
 #include "RAJA/RAJA.hpp"
 #include "RAJA/util/Timer.hpp"
+#include "RAJA/util/View.hpp"
```
artv3: Doesn't view and timer get in through RAJA/RAJA.hpp?

include/RAJA/util/macros.hpp
ggeorgakoudis: Yes, this is a good point.

examples/construct_tuple.cpp
```diff
@@ -0,0 +1,22 @@
+#include <camp/tuple.hpp>
+#include <iostream>
```
MrBurmark: Is this example added in the cmake?

MrBurmark: What does this have to do with jit?

johnbowen42: Accidentally added.  Will remove

scripts/gitlab/build_and_test.sh
```diff
@@ -26,12 +26,16 @@ project_dir="$(pwd)"
...
 module_list=${MODULE_LIST:-""}
+cmake_extra_args_raw=${CMAKE_EXTRA_ARGS:-""}
 job_unique_id=${CI_JOB_ID:-""}
 use_dev_shm=${USE_DEV_SHM:-true}
 spack_debug=${SPACK_DEBUG:-false}
 debug_mode=${DEBUG_MODE:-false}
 push_to_registry=${PUSH_TO_REGISTRY:-true}
 
+# Map CPU core allocations
+declare -A core_counts=(["lassen"]=40 ["poodle"]=28 ["dane"]=28 ["matrix"]=28 ["corona"]=32 ["rzansel"]=48 ["tioga"]=32 ["tuolumne"]=48)
```
MrBurmark, line 37: What is this used for?

MrBurmark: I'm pretty sure some of these have more cores than are listed here.

johnbowen42: this already existed in the script, but I moved it up because I needed to access the variable earlier

MrBurmark: Oh, I'm just blind, carry on.

examples/forall-jit.cpp
johnbowen42: `View.hpp` does get included, `Timer.hpp` does not.

johnbowen42: I removed `#include "RAJA/util/View.hpp"`

johnbowen42:
update: added docs of usage and CMake build configuration.  I also simplified the CMake logic

docs/sphinx/user_guide/feature/jit.rst
```diff
@@ -0,0 +1,124 @@
...
+      b = proteus::jit_variable(b),
+      accum = proteus::jit_variable(accum)
+    ] (int i) RAJA_JIT_COMPILE {
+      // kernel body that uses a, b, accum
+    });
+
+When JIT is disabled (``RAJA_ENABLE_JIT=Off``), ``RAJA_JIT_COMPILE`` expands to
+nothing. However, code that calls Proteus APIs (e.g., ``proteus::jit_variable``)
+must still be conditionally compiled or guarded by build configuration if you
+want a non-JIT build to compile without Proteus available.
```
MrBurmark: Where did we end up on providing a RAJA::jit_variable function?

johnbowen42: I removed it a few commits ago.  I discovered a bug where the lambda fails to specialize IR when proteus::jit_variable is called in a different function than the declaration site of the lambda.  I.E., unless we `__attribute__((always_inline))` `RAJA::jit_variable`, we don't get the fully optimized code generation.  I have a fix though and will add back `RAJA::jit_variable` when soon

johnbowen42: We could add it back with the forced inlining, but I don't want Proteus's analysis to assume anything about how users are inlining code.

examples/forall-jit.cpp
```diff
@@ -0,0 +1,120 @@
...
+  int b = std::stoi(argv[2]);
+  int N =  std::stoi(argv[3]);
+  bool accum = std::stoi(argv[4]);
+  #ifdef RAJA_ENABLE_CUDA
+  using policy = RAJA::cuda_exec<256>;
+  #elif defined(RAJA_ENABLE_HIP)
+  using policy = RAJA::hip_exec<256>;
+  #elif defined(RAJA_ENABLE_OPENMP)
+  using policy = RAJA::omp_parallel_for_exec;
+  #endif
```
artv3, line 58: I think we need the `seq_exec` as a fallback since its always available

johnbowen42: done

examples/forall-jit.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include <iostream>
+#include <limits>
+
+#include "RAJA/RAJA.hpp"
+#include "RAJA/util/Timer.hpp"
+#include "proteus/JitInterface.h"
+
+int main (int argc, char** argv) {
+  if (argc < 3) {
+    std::cout << "Expected integer threshold passed to example through command line\n";
```
artv3: It might be good to add a sample run along with this message

artv3 (approved):

artv3:
Overall looks great! Only minor stuff to address for me -- compiler version support (I was not able to build with newer versions of rocm)

johnbowen42:
> Overall looks great! Only minor stuff to address for me -- compiler version support (I was not able to build with newer versions of rocm)

This has been fixed--my build script had a bug where the LLVM version was hard-coded.  All versions of ROCM included on LC should work now

examples/forall-jit.cpp
johnbowen42: done!

cmake/RAJAMacros.cmake
```diff
@@ -50,6 +67,12 @@ macro(raja_add_executable)
     DEPENDS_ON ${arg_DEPENDS_ON}
     OUTPUT_DIR ${_output_dir}
     )
+
+  # If JIT is enabled, link with target with Proteus and identify Proteus
```
rhornung67:
```suggestion
  # If JIT is enabled, link target with Proteus and identify Proteus
```

cmake/RAJAMacros.cmake
```diff
@@ -86,6 +109,11 @@ macro(raja_add_plugin_library)
     SHARED ${arg_SHARED}
     )
 
+  # If JIT is enabled, link with target with Proteus and identify Proteus
```
rhornung67:
```suggestion
  # If JIT is enabled, link target with Proteus and identify Proteus
```

docs/sphinx/user_guide/feature/jit.rst
```diff
@@ -0,0 +1,127 @@
...
+  ``find_package(proteus ...)`` using that prefix.
+* Otherwise, RAJA will attempt to fetch Proteus via CMake ``FetchContent`` at
+  configure time.
+
+LLVM installation requirement
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Unless you provide a Proteus installation that is statically linked with LLVM,
+Proteus support requires an LLVM 18, 19, or 20 installation that you must point
+RAJA/Proteus at it via ``LLVM_INSTALL_DIR``:
```
rhornung67:
```suggestion
RAJA/Proteus at via ``LLVM_INSTALL_DIR``:
```

docs/sphinx/user_guide/feature/jit.rst
```diff
@@ -0,0 +1,127 @@
...
+  cmake -DRAJA_ENABLE_JIT=On -DLLVM_INSTALL_DIR=/path/to/llvm-19 ...
+  
+An example of how to configure a JIT build of RAJA with HIP on LC machines is included
+in ``scripts/toss4_amdclang_proteus.sh``.
+
+--------------------------
+Marking a kernel for JIT
+--------------------------
+
+The user-facing interface shown in ``examples/forall-jit.cpp`` consists of:
```
rhornung67, line 66: I suggest that you link in an actual code section from the file, and refer to that in the text, so it's easy to keep the documentation consistent with the example code.

docs/sphinx/user_guide/feature/jit.rst
```diff
@@ -0,0 +1,127 @@
...
+    [=,
+      a = proteus::jit_variable(a),
+      b = proteus::jit_variable(b),
+      accum = proteus::jit_variable(accum)
+    ] (int i) RAJA_JIT_COMPILE {
+      // kernel body that uses a, b, accum
+    });
+
+When JIT is disabled (``RAJA_ENABLE_JIT=Off``), ``RAJA_JIT_COMPILE`` expands to
+nothing. However, code that calls Proteus APIs (e.g., ``proteus::jit_variable``)
```
rhornung67: You should show the conditional compilation in the example.

docs/sphinx/user_guide/feature/jit.rst
```diff
@@ -0,0 +1,127 @@
...
+must still be conditionally compiled or guarded by build configuration if you
+want a non-JIT build to compile without Proteus available.
+
+Building and running the example
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The example is built from the RAJA source tree when:
+
+* ``ENABLE_EXAMPLES=On`` (default in RAJA)
```
rhornung67: I don't think `ENABLE_EXAMPLES` is defaulted to On in the RAJA CMake. We have a CMake dependent option `RAJA_ENABLE_EXAMPLES` that defaults to Off. Please double check this.

docs/sphinx/user_guide/feature/jit.rst
```diff
@@ -0,0 +1,127 @@
...
+* ``RAJA_ENABLE_JIT=On``
+
+The example takes four command-line arguments:
+
+.. code-block:: bash
+
+  ./bin/forall-jit <a> <b> <N> <accum>
+
+where ``a`` and ``b`` are matrix dimensions, ``N`` is the problem size, and
```
rhornung67, line 101: This needs more explanation. For example, what does the example code do? How are the arguments used? Etc....

docs/sphinx/user_guide/feature/jit.rst
```diff
@@ -0,0 +1,127 @@
...
+``accum`` is the branch condition (0/1) that is specialized with JIT.
+
+Specializing argument indices (advanced)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+RAJA also provides ``RAJA_JIT_COMPILE_ARGS(...)`` to annotate functions and
+specify which 1-indexed arguments should be treated as specialization inputs:
```
rhornung67, line 113: Is the `1-indexed` args convention going to be changed?

scripts/lc-builds/toss4_amdclang_proteus.sh
```diff
@@ -0,0 +1,114 @@
...
+module load cmake/${CMAKE_VER}
+
+# unload rocm to avoid configuration problems where the loaded rocm and COMP_VER
+# are inconsistent causing the rocprim from the module to be used unexpectedly
+module unload rocm
+
+
+cmake \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DROCM_ROOT_DIR="/opt/rocm-${COMP_VER}" \
```
rhornung67, line 79: We really should investigate which of the ROCM/HIP/AMD options are truly necessary at this point. This is kinda crazy.

CMakeLists.txt
```diff
@@ -331,6 +331,42 @@ install(FILES
 # Setup internal RAJA configuration options
 include(cmake/SetupRajaConfig.cmake)
 
+if (RAJA_ENABLE_JIT)
```
rhornung67, line 334: Is there a reason that it is preferable that this section is inserted here instead of in a *.cmake file in the cmake directory and included here?

rhornung67 (changes requested):
Some questions and minor change requests.

docs/sphinx/user_guide/feature/jit.rst
johnbowen42: I'm tagging @ggeorgakoudis because I'm not sure if this is on the roadmap

CMakeLists.txt
johnbowen42: no, we can do it that way

johnbowen42: originally the CMake logic was much simpler, but I think that what you say is a better way to handle it now

docs/sphinx/user_guide/feature/jit.rst
ggeorgakoudis: I don't foresee changing this convention, but we plan to build more ergonomic interfaces (e.g., provide the parameter name or annotate directly the parameter) by expanding to a frontend (Clang) plugin.

johnbowen42: Thanks Giorgis

docs/sphinx/user_guide/feature/jit.rst
johnbowen42: I added a full explanation

docs/sphinx/user_guide/feature/jit.rst
johnbowen42: fixed

docs/sphinx/user_guide/feature/jit.rst
johnbowen42: Instead, I added a macro `RAJA_JIT_VARIABLE` that simply forwards its argument if jit is not enabled, but returns `proteus::jit_variable` otherwise.

johnbowen42:
great feedback, thanks @rhornung67 .  I think I addressed everything, and I added a macro that wraps `proteus::jit_variable`, so no conditional compilation will be necessary in user code.

docs/sphinx/user_guide/feature/jit.rst
rhornung67: OK. I'll refrain from making a Fortran joke ;-)

rhornung67 (approved):

docs/sphinx/user_guide/feature/jit.rst
ggeorgakoudis: > OK. I'll refrain from making a Fortran joke ;-)

Please don't! We'll be stuck in a DO LOOP of suspense.

artv3:
@johnbowen42 we usually delete the branch after merge

PR #1975, opened 2026-01-13
draft: enable openmp with windows

johnbowen42:
We saw some failures in axoms CI as a result of (1) C++17 changes to OpenMP backends, (2) Windows build, (3) openmp enabled, so I am enabling openmp here to troubleshoot and hopefully add coverage


rhornung67:
@johnbowen42 We're going to have problems with this because MSVC does not support some OpenMP standard stuff we use; e.g., it doesn't know about the `combine` directive in a reduction clause, which we use.  There may be others, but that is one reason why the windows build with OpenMP enabled is failing here.

PR #1977, opened 2026-01-15
bugfix: Fix axom's MSVC with OpenMP build

johnbowen42:
Axom found some bugs in RAJA's latest release with two root causes:
https://github.com/llnl/RAJA/pull/1884
https://github.com/llnl/RAJA/pull/1947

The first led to MSVC attempting to compile some OpenMP pragmas it didn't understand.  The second led to undefined symbol names.  Both of these are ultimately strange, because our tests and examples do not even compile with `ENABLE_OpenMP=On` with `msvc`.  Axom's Windows workflow is to build RAJA without any of the tests and exercises enabled, install RAJA, then only use non-reduction versions of RAJA's OpenMP constructs.  This only ever worked because SFINAE and code duplication allowed reduction-free OpenMP code to be compiled by Axom downstream, but RAJA's reduction versions of launch, forall and kernel to go uncompiled.  

tldr; RAJA + OpenMP + msvc doesn't work in most cases, but axom carved out some use cases that do.  This PR retains that functionality.  

We might need to add more #ifdef guards to the launch and forall backends 

@bmhan12


johnbowen42:
this pipeline https://github.com/llnl/RAJA/pull/1975 should catch problems like the one from #1947 in the future, but the error the cpp 17 PR introduced won't be tested by this.  Our only option for testing that would be to add a Windows-only test, because many tests and examples do not compile with Windows and OpenMP

bmhan12:
Can confirm these changes do fix the failures Axom was seeing in LLNL/axom#1769 !

include/RAJA/internal/ThreadUtils_CPU.hpp
```diff
@@ -24,7 +24,10 @@
 #include "RAJA/config.hpp"
 
 #include "RAJA/pattern/thread.hpp"
+#if defined(RAJA_ENABLE_OPENMP)
```
adayton1, line 27: The include of omp.h should be moved into RAJA/policy/openmp/thread.hpp.

include/RAJA/policy/openmp/forall.hpp
```diff
@@ -82,6 +82,9 @@ forall_impl(resources::Host host_res,
   }
   else
   {
+// This branch handles the case of an OpenMP reduction through the RAJA::kernel
+// abstraction. MSVC is not supported in this case.
+#if !defined(RAJA_COMPILER_MSVC)
```
adayton1: There has to be a different way to do this. This approach will just give the wrong answer silently, and we definitely don't want that.

adayton1: I think you might be able to use static assert since this is the else branch of an if constexpr.

```
#if defined(RAJA_COMPILER_MSVC)
static_assert(false, "MSVC does not support an OpenMP reduction through the RAJA::kernel abstraction");
#endif
```

You might have to make use of the template arguments instead of directly using false in the static_assert, though I found in some simple examples that false worked for clang and gcc.

adayton1: https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGEjaSuADJ4DJgAcj4ARpjEEgCspAAOqAqETgwe3r7%2ByanpAiFhkSwxcVyJdpgOGUIETMQEWT5%2BXAFVNQJ1DQRFEdGxCbb1jc05bcM9fSVlCQCUtqhexMjsHOYAzKHI3lgA1CYbbk4KBMSYrIfYJhoAgpvbu5gHRwQAnkmYAPpnTIQKVxu9zuBEwLCSBlBLzc70%2BzDYewAKoC7gA3VB4dB7KgQRF7VFiLyYOYHADsVjueypezwVD2aAYp0wqiSxD2EFO6BAIDwCi%2BClY31RhzciNINMEVxJJnJQOp8r2nO5aC8BGhIvxhMwhwpt3lMoAInLqZhaEoybqFdTTkxHMgvkwFEpGtQxEpxeYzAB5T7EW0ZPYMVBqhReJIpRqYdCeuY641Uw1AxN3IGhNUsP4MCBpvYNYDIcXIBANABUJdzxGAqOlssp1JxG1jG0t9YgGwAdPEmy2qecCMsGHsNHGU6SDRwFrROPFeH4OFpSKhOG5rNZFUsVs9NjxSARNBOFgBrEDxLjtrgANg0AA5SdeAJzXi8XyRmMz6TiSWf7xecXgKEAAj3ecJ1IOBYBgRAUFQcE6FichKDQWD6DiHZDGALh7wmGhaFBYgAIgKIfyiUIGjeTgdxI5hiDeL0om0apgJ3JC2EEL0GFociQNILAoi8YA3DEM0KN4LAMyMcRuFEvBzhqVFMAA7jmWqVU1gXNNTRE/Q8CiP0aI8LAfzOPAWC0%2BTiCiVJMANMF0NoUJQBAhYqAMYAFAANTwTAAHcfUYLT%2BEEEQxHYKQZEERQVHUbjdC4fR0JQVdLG0qIAMgBZUCSRwBEUgBaTlDgNUxLGsLhSV4VBzOITEFPgBYOmyvwIFcMZWkCBh0GmAZyjyNJGtavQUj6jIutKQY4oa2oRiaTwWj0Saumm0bZgm6aBtWqZQn6MbynqjdVgkSdp2/bilw4PZVCfXKX3pAwjD2TDz3bDR2VwQgSAOMwNi4OZeGArQ5iPEAzEkZ6KlJLgNkke8NgvO8NDiqcOC/UhTI2Dt0cxrHsYvUg5wXM7/0A3d9wWcCoKWAgklVBCICQpI4OIcJBU4S6L2uyQaWQVQ9jMDQ3yehco3e6quTiwLhFEcQwolyK1B/WLSG8v0klJj8OBnPGfzOr1VWptVUDpNmOdu9CHvvQX2Q8ZDYk%2B77fpJpyFgQC4sDiCAjuR3g0Yx7G/fR3H8Yqv9bGJ/7QPJiAkCwVE8FWWn6cZ5m2FZq6bvzZAHvidszFEzBY4OjheGFohRb0CXgul6RZaUeWYvm01GIyFwOvW9rOq2mZxt6gpMlm8Ye8a5bu4WhhulGfu2tH8fek77r5rWyeF824p55%2B3dzkwTFAM9zWg9/DgbIL55PJ823jZutD7ses8XogN7S7tuK9mthmUKfh3w8B9WUZ99t/f9lrU6IcAJAVJkDSQ152zXg2FhMwbRIYbHiHedWGwToExDl/T2ucgEYKLo7AGCxzJpGcJIIAA%3D%3D

johnbowen42: that's a good idea Alan, I have added the static_assert statement.  However, we still need the ifdef guard around the OpenMP pragmas.  We have to hide the whole pragma from MSVC to support Axom's particular use case.

adayton1 (approved):

include/RAJA/policy/thread_auto.hpp
```diff
@@ -25,6 +25,7 @@
 #include "RAJA/util/macros.hpp"
 
 #ifdef RAJA_OPENMP_ACTIVE
+#include <omp.h>
```
adayton1: This looks redundant to me. I see omp.h being included in openmp/policy.hpp.

adayton1: It really should be included in the files where it is actually used. If it was missing from somewhere, I'm guessing it is openmp/thread.hpp.

johnbowen42: tracked it down to a ifdef MSVC conditional include of omp.h in policy.hpp

rhornung67: Good detective work @johnbowen42. This is timely since we will be pushing out a patch release soon.

PR #1980, opened 2026-01-23
Change build scripts to accept an optional cmake version argument

rhornung67:
# Summary

- This PR modifies build scripts to accept an optional cmake version argument
- If no cmake version is given, a default based on the RAJA cmake min version is set
- This helps with the fact that some of these scripts can be used on multiple platforms that may have different cmake version defaults or versions that don't exist on others


scripts/lc-builds/corona_sycl.sh
```diff
@@ -9,25 +9,38 @@
...
+
+# Detect optional second positional argument as a CMake version if it looks like N.M or N.M.P
+# Otherwise, treat it as a normal CMake argument.
+if [ -n "$2" ] && [[ "$2" =~ ^[0-9]+(\.[0-9]+)*$ ]]; then
+  CMAKE_VER=$2
+  shift 2
+else
+  CMAKE_VER=$DEFAULT_CMAKE_VER
+  shift 1
+fi
```
MrBurmark, line 35: Would it be more clear if the cmake version was an optional named argument like `script compiler_ver [--cmake ver] [cmake args...]`? Would putting it first be more uniform `script [--cmake ver] compiler_ver [cmake args...]`?

rhornung67: Maybe, but we don't use named args for input to any of the scripts.

I put it last because it's optional and in most cases, it probably won't be needed.

This PR is intended to deal with the specific issue that the CMake version is currently hard-coded in the scripts and we need to be more flexible and it's a pain to modify the scripts for minor changes like that.

johnbowen42: I think in general named arguments are better, and that it would be better if all the build scripts had only named arguments (ideally in Python). However, given that these scripts are used primarily internally, I think it would probably be better to make this as minimal of a change as possible.  

In terms of improving usability of these scripts, I think the best thing would probably be a Python driver that calls the shell scripts automatically, and provides users with a nice `--help` description.

scripts/lc-builds/corona_sycl.sh
```diff
@@ -9,25 +9,38 @@
...
+  echo "If no CMake version is provided, version ${DEFAULT_CMAKE_VER} will be used."
+  exit 1
 fi
 
 SYCL_PATH=$1
-shift 1
+
+# Detect optional second positional argument as a CMake version if it looks like N.M or N.M.P
+# Otherwise, treat it as a normal CMake argument.
+if [ -n "$2" ] && [[ "$2" =~ ^[0-9]+(\.[0-9]+)*$ ]]; then
```
johnbowen42, line 29: Maybe it would be better to have an additional clause like this
```
elif [ -n "$2" ] && [[ "$2" !=~ ^[0-9]+(\.[0-9]+)*$ ]]; then
echo "Unable to parse CMake version number"
```

johnbowen42: Given that this argument won't be used widely I think this help message is not strictly necessary

johnbowen42 (approved):

scripts/lc-builds/corona_sycl.sh
rhornung67: I agree with that. If you feel inclined to pursue that, please do so.  

My intent with this PR was to do a quick, simple change to make the scripts more useful w.r.t. a particular issue.

scripts/lc-builds/corona_sycl.sh
rhornung67: I agree that we can make the scripts much more helpful. However, they are specific to LC systems and used mostly for our own development purposes. Sometimes, I point people to them for examples.

MrBurmark:
We'll have to change the RAJAPerf scripts too.

rhornung67:
> We'll have to change the RAJAPerf scripts too.

Yes. I wanted to run this by folks first.

MrBurmark (approved):

PR #1985, opened 2026-02-04
Use C++ compiler to check for symbols

smithsg84:
# Symbol existence checks changed to use C++ compiler rather than the C compiler

check_symbol_exists use the C compiler for the symbol check, check_cxx_symbol_exists is the C++ variant.   The C check was being used to check for a C++ symbol name (std::aligned_alloc) which was failing.   Other checks were refactored to use C++ checks since use is in C++ code.

- This PR is a bugfix

### DIFF
--- a/cmake/SetupRajaConfig.cmake
+++ b/cmake/SetupRajaConfig.cmake
@@ -49,10 +49,10 @@ else ()
     set(RAJA_USE_CLOCK   OFF CACHE BOOL "Use clock from time.h for timer"    )
 endif ()
 
-include(CheckSymbolExists)
-check_symbol_exists(posix_memalign stdlib.h RAJA_HAVE_POSIX_MEMALIGN)
-check_symbol_exists(std::aligned_alloc stdlib.h RAJA_HAVE_ALIGNED_ALLOC)
-check_symbol_exists(_mm_malloc "" RAJA_HAVE_MM_MALLOC)
+include(CheckCXXSymbolExists)
+check_cxx_symbol_exists(posix_memalign cstdlib RAJA_HAVE_POSIX_MEMALIGN)
+check_cxx_symbol_exists(std::aligned_alloc cstdlib RAJA_HAVE_ALIGNED_ALLOC)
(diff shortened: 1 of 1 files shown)


davidbeckingsale (approved):

davidbeckingsale:
Very nice - I think we've had similar issues in Umpire.

smithsg84:
The first choice for aligned allocation is currently posix_memalign.   Since RAJA now requires C++ 17 should the first selection be std::aligned_alloc which is in C++17?   Remove RAJA wrapper entirely and use standard directly since we have bumped to C++17?   Or perhaps there are other reasons for the selection?

https://github.com/llnl/RAJA/blob/05256aa85b42412abb5c0b9ef7421a9c52ddd8f4/include/RAJA/internal/MemUtils_CPU.hpp#L46

rhornung67:
@MrBurmark @artv3 will you please comment on @smithsg84 question above about removing the RAJA wrapper and going with C++ std library directly?

rhornung67 (approved):
Thanks @smithsg84 As you can see we've accumulated vestigial code over the years as we've evolved through C++ standard versions.

adayton1:
I would be inclined to use std::aligned_alloc and drop the others.

MrBurmark:
I don't know of a performance reason to prefer one aligned allocation function over another. I'm in favor of using the std version.

smithsg84:
This PR is superseded by #1986 to remove the use of platform/compiler aligned allocators and use the C++17 standard allocator.

smithsg84:
This PR is superseded by https://github.com/llnl/RAJA/pull/1986 to remove the use of platform/compiler aligned allocators and use the C++17 standard allocator.

smithsg84:
MSVC does not support the std::aligned_alloc and MACOS std::aligned_alloc use failed CI tests.   So using the std::aligned_alloc has portability issues.

Will use fix to address the CMake issues related to using C compiler for checks and leave allocation changes for future investigation.

PR #1986, opened 2026-02-05
Use C++17 aligned memory allocation

smithsg84:
# Use C++17 aligned memory allocation rather than platform/compiler allocators

With the adoption of C++17 requirement for RAJA the standard aligned memory allocator (std::aligned_alloc) is available and can replace platform / compiler specific options.  This PR removes CMake checks for the platform allocators and simplifies the aligned memory allocator wrappers.

- This PR is a refactoring
  - Refactors the internal aligned memory allocators to use the std allocators.
  
 Discussion on removal is in a prior cmake fix that has been closed #1985.   This PR replaces that change with a more complete change.

### DIFF
--- a/cmake/SetupRajaConfig.cmake
+++ b/cmake/SetupRajaConfig.cmake
@@ -49,10 +49,10 @@ else ()
     set(RAJA_USE_CLOCK   OFF CACHE BOOL "Use clock from time.h for timer"    )
 endif ()
 
-include(CheckSymbolExists)
-check_symbol_exists(posix_memalign stdlib.h RAJA_HAVE_POSIX_MEMALIGN)
-check_symbol_exists(std::aligned_alloc stdlib.h RAJA_HAVE_ALIGNED_ALLOC)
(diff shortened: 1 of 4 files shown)


adayton1 (approved):

smithsg84:
@MrBurmark @artv3 Based on discussion in #1985 I have removed the old checks for aligned memory allocators and replaced using the C++17 aligned allocators.   Simple testing showed no measurable performance difference between the posix_memalign and std::aligned_alloc.

rhornung67 (approved):

smithsg84:
Deep sigh, Windows MSVC compiler does not support std::aligned_alloc so the platform specific calls are needed.

smithsg84:
Closing PR.

Windows MSVC compiler does not support std::aligned_alloc as of 2/5/2026, see [notes section](https://en.cppreference.com/w/cpp/memory/c/aligned_alloc).

MacOS CI tests were failing when using the std::aligned_alloc allocator.

PR #1989, opened 2026-02-13
Update RSCI and adapt to machine-status-override

adrienbernede:
This fixes the issue where a failure status due to machine outage is never overridden when the machine comes back and CI passed.


adrienbernede:
@rhornung67 This is ready.

rhornung67 (approved):

PR #1990, opened 2026-02-19
c++20:  convert SFINAE in forall and sort to use concepts, consolidate concepts around a single header

johnbowen42:
This PR consolidates our concepts into a single concepts header, and replaces SFINAE inside forall and sort with concepts


include/RAJA/pattern/forall.hpp
```diff
@@ -174,36 +170,33 @@ namespace wrap
...
+template<Resource Res,
+         ForallPolicy ExecutionPolicy,
+         Range Container,
          typename LoopBody>
-RAJA_INLINE concepts::enable_if_t<
-    RAJA::resources::EventProxy<Res>,
-    concepts::negate<type_traits::is_indexset_policy<ExecutionPolicy>>,
-    type_traits::is_range<Container>>
-forall(Res r, ExecutionPolicy&& p, Container&& c, LoopBody&& loop_body)
+RAJA_INLINE RAJA::resources::EventProxy<Res> forall(Res r,
```
MrBurmark, line 196:
```suggestion
requires(!IndexSetPolicy<ExecutionPolicy>)
RAJA_INLINE RAJA::resources::EventProxy<Res> forall(Res r,
```

include/RAJA/pattern/forall.hpp
```diff
@@ -467,24 +448,19 @@ forall(ExecutionPolicy&& p, Container&& c, LoopBody&& loop_body)
...
  ******************************************************************************
  */
-template<typename ExecutionPolicy,
-         typename Res,
-         typename Container,
-         typename IndexType,
+template<ForallPolicy ExecutionPolicy,
+         Resource Res,
+         RandomAccessRange Container,
+         Integer IndexType,
```
MrBurmark, line 454: Index type and container appear in other templates above, should they also use Integer and RandomAccessRange?

include/RAJA/pattern/concepts.hpp
```diff
@@ -0,0 +1,115 @@
...
+    ForallParams<T> &&
+    RAJA::expt::type_traits::is_ForallParamPack_empty<std::decay_t<T>>::value;
+
+template<typename T>
+concept NonEmptyForallParams =
+    ForallParams<T> &&
+    !RAJA::expt::type_traits::is_ForallParamPack_empty<std::decay_t<T>>::value;
+
+template<typename T>
+concept MultiPolicyConcept = static_cast<bool>(
```
MrBurmark: Why does this one have a static_cast and the others don't?

include/RAJA/pattern/concepts.hpp
```diff
@@ -0,0 +1,115 @@
...
+    RAJA::type_traits::is_multi_policy<std::decay_t<T>>::value);
+
+/// Iteration mappings inherit from things like StridedLoopBase
+#define MakeExecPolWithIterMappingConcept(ConceptName, BaseName)               \
```
MrBurmark: Defines should always start with the library name, did we not do that before?

include/RAJA/pattern/forall.hpp
```diff
@@ -175,36 +176,32 @@ namespace wrap
...
-template<typename Res,
-         typename ExecutionPolicy,
-         typename Container,
+template<concepts::Resource Res,
+         concepts::ExecutionPolicy ExecutionPolicy,
+         concepts::Range Container,
          typename LoopBody>
-RAJA_INLINE concepts::enable_if_t<
-    RAJA::resources::EventProxy<Res>,
-    concepts::negate<type_traits::is_indexset_policy<ExecutionPolicy>>,
```
MrBurmark, line 205: We should probably be consistent about requiring not index set policy here and above. I requested it be removed, but I don't know why anymore and now these two differ for seemingly no reason.

include/RAJA/pattern/concepts.hpp
johnbowen42: this one is defined from `RAJA::type_traits::SpecializationOf` which inherits from camp's type traits, not stl `true_type`.  For some reason, the underlying type of `camp::false_type` is long (`using false_type = num<false>;`)

johnbowen42: same reason as `IndexSetPolicy` and `IndexSet`

include/RAJA/pattern/concepts.hpp
johnbowen42: We did not--I removed some macros last PR that were this way.  I'll rename this one

include/RAJA/pattern/forall.hpp
johnbowen42: Ok, I added a `requires` here to match the one above

include/RAJA/pattern/concepts.hpp
johnbowen42: I reorganized and added and explanatory comment

MrBurmark (approved):

PR #1994, opened 2026-02-23
ci: move from clang 14 to clang 19 in docker testing

johnbowen42:
clang 14 does not support C++20 fully, so move to 19 in preparation.  Also move to clang-format 19


johnbowen42:
closing this PR as there are stale (clang14) expected checks associated with it from an old commit

PR #1997, opened 2026-02-25
Use project specific variable to specify github status token

adrienbernede:
Fix issue with draft PR filter.

The issue identified is that token cannot be accessed correctly when defined at group level anymore.
Instead, we defined the token at project level, and seize the occasion to give it a name that is more specific.

We turn the draft filter back on.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -81,7 +81,7 @@ include:
     inputs:
       github_project_name: $GITHUB_PROJECT_NAME
       github_project_org: $GITHUB_PROJECT_ORG
-      github_token: $GITHUB_TOKEN
+      github_token: $RSCI_GITHUB_STATUS_TOKEN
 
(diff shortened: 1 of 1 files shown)


rhornung67:
@adrienbernede tioga is back up so I restarted the CI on that machine. it is running now.  hopefully, the check wil be green when it finishes

rhornung67:
@adrienbernede all tests passed, including on tioga. But, that status is not being reported back to GitHub as it still shows tioga being down. We've seen this before. Can this issue be fixed?

adrienbernede:
@rhornung67 This is fixed in https://github.com/llnl/RAJA/pull/1989 which is ready but has a failed status in wait of the present PR.
Please merge this PR first and then the other one will be passing CI and ready to merge.

rhornung67 (approved):

PR #1999, opened 2026-03-03
Update RSC (rocm toolchain, mpich, cce), spack and spack-packages

adrienbernede:
- [X] Update to rocm 6.4.3 and mpich 8.1.33, then with mpich 9.0.1
- [X] Update to spack 1.1.1 (concretizer perf improvements)
- [X] Update to cce 21
- [X] Add one job with rocm 7.2 and mpich 9.1, keep rocm 6.4.3 for most specs (production state).


rhornung67:
@adrienbernede same comment that I posted on the similar RAJA Perf PR....

Where is the cray-mpich version defined? Above, you say move to 8.1.33. That version is old. We should be using version 9.0.1. I think a new version will be arriving in the near future.

adrienbernede:
@rhornung67 OK, I’m going to be using mpich 9.1.0 with rocm 7.2.

adayton1:
I was under the impression that codes should be using rocm 6.4.2 or 6.4.3 with cray-mpich 8.1.33 or 9.0.1 in production and that LC wants us to start testing out rocm 7.2.0 with cray-mpich 9.1.0 to determine if there are any problems. So I think both combinations are important (6.4.3 + 9.0.1 and 7.2.0 + 9.1.0).

Any version of rocm 7 requires at least cray-mpich 9.1.0. I think 9.1.0 might work with earlier versions of rocm, but it sounds like it is still considered kind of a beta.

rhornung67:
Good point @adayton1 Thanks for posting your comment.

adrienbernede:
@rhornung67 would you like to test both
- rocm 6.4.3 + mpich 9.0.1
- rocm 7.2.0 + mpich 9.1.0

or just one of those ?

rhornung67:
Let's do both. The first one is being used by all of our WSC/CP app users. The second will be used by those projects in the near future. It's good to stay ahead of the downstream projects.

adrienbernede:
@rhornung67 This is ready!

adrienbernede:
@rhornung67 this should really be merged: it will update the toolchains and reduce spack concretization times.

rhornung67 (approved):

PR #2005, opened 2026-03-16
Generic atomic operation

adayton1:
# Summary

- This PR is a feature
- It does the following:
  - Adds a generic atomic operation at the request of myself

Use case:
- Computing a signed distance field, and need to do the atomic comparison based on the absolute value but update with the signed value.
- I think I've also seen the need for an atomic multiply in an application.

Design discussion points:
- atomicGeneric seemed like a less overloaded name than atomicOperation (which would confuse vendors)
- To keep the interface and implementation simple, no short-circuiting or custom comparison functions were added in this merge request. They could be added in a future merge request.


include/RAJA/policy/desul/atomic.hpp
```diff
@@ -27,6 +32,50 @@ using raja_default_desul_scope = desul::MemoryScopeDevice;
...
+}
+
+template<typename T, typename Oper>
+RAJA_HOST_DEVICE RAJA_INLINE T desul_atomicCAS_loop(T* acc, Oper&& oper)
+{
+  T old = desul::atomic_load(acc, raja_default_desul_order {},
+                            raja_default_desul_scope {});
+  T expected;
+
+  do
```
MrBurmark: Do we want to add a short circuit between the atomic_load and the atomic_compare_exchange?

include/RAJA/policy/desul/atomic.hpp
```diff
@@ -153,6 +184,34 @@ atomicCAS(AtomicPolicy, T* acc, T compare, T value)
...
+      return expected; // no-op
+    }
+
+    const T old = desul::atomic_compare_exchange(acc, expected, desired,
+                                                 raja_default_desul_order {},
+                                                 raja_default_desul_scope {});
+
+    if (desul_atomicCAS_equal(old, expected)) {
+      return old; // success
+    }
```
MrBurmark, line 209: Can we get away with one conditional per iteration?

adayton1: If we want short-circuiting, I don't think so.

adayton1: I'll think about this some more.

adayton1: Yeah, I don't think there's a way to do short-circuiting without two checks. If we are really worried about performance, we could expose a couple of different overloads of atomicOperation (with a functor like the different cas loop overloads we have for CUDA and HIP), or a single overload with a boolean parameter. There are definitely cases where shortcircuiting helps (like min/max or there's a chance you are multiplying by 1 or adding 0 to a non-builtin type). But there are plenty of cases where it would just be extra overhead.

MrBurmark: That makes sense. To summarize some operators like min/max make sense to short circuit as they will act like the identity function in some cases. Others like plus/mult make less sense to short circuit as they (almost) always return a different value.
Given that we internally have both I would lean toward providing both.

include/RAJA/policy/desul/atomic.hpp
```diff
@@ -27,6 +32,32 @@ using raja_default_desul_scope = desul::MemoryScopeDevice;
 namespace RAJA
 {
 
+namespace detail
+{
+
+template<typename T>
+RAJA_HOST_DEVICE RAJA_INLINE bool desul_atomicCAS_equal(const T& a, const T& b)
```
MrBurmark, line 39: Do we not already have something like this?

include/RAJA/policy/desul/atomic.hpp
```diff
@@ -27,6 +32,32 @@ using raja_default_desul_scope = desul::MemoryScopeDevice;
...
+{
+  return a == b;
```
MrBurmark, line 41: This makes me wonder whether it makes sense to also allow a used-defined comparison function for comparing `old` and `expected` as well, as some may not care about equivalent vs bit-wise equal. We may want to default to using bit-wise equality in user-defined types in addition to floating point and integer types.

include/RAJA/policy/desul/atomic.hpp
MrBurmark: We could have a single implementation with a short circuit and "old_matches_expected" operator where the short circuit functor defaults to one that always returns false and the "old_matches_expected" operator does a bit-wise comparison.

include/RAJA/policy/desul/atomic.hpp
adayton1: Yes, it's duplicated in multiple places. Is there a centralized place you would suggest moving these implementations?

MrBurmark: Somewhere in pattern or maybe internal would be a good place.

include/RAJA/policy/desul/atomic.hpp
adayton1: Are you thinking about this just for the short-circuiting, or for determining whether the atomic CAS happened as well?

adayton1: I'm a little worried about the interface getting too complicated.

include/RAJA/policy/desul/atomic.hpp
adayton1: I'm pulling out the short-circuiting stuff for now. Having the simple generic implementation is valuable, and optimizations or additional overloads could be added at a future date.

include/RAJA/policy/desul/atomic.hpp
adayton1: I added it as a function in util/TypeConvert.hpp, which is where the reinterp_A_as_B function is found.

include/RAJA/policy/desul/atomic.hpp
adayton1: I'm going to leave this out for now to keep the implementation simple and easy to use. Optimizations can be added later.

include/RAJA/util/TypeConvert.hpp
adayton1, line 52: Should we also assert that both types are trivially copyable?

include/RAJA/policy/cuda/atomic.hpp
```diff
@@ -366,11 +345,36 @@ RAJA_INLINE __device__ T cuda_atomicCAS_loop(T* acc,
...
 }
 
+/*!
+ * Generic atomic operation implemented using a CAS loop.
+ * Returns the OLD value that was replaced by the result of this operation.
+ */
+template<typename T, typename Operation>
+RAJA_INLINE __device__ T cuda_atomicGeneric(T* acc, Operation&& operation)
+{
+#if defined(RAJA_ENABLE_CUDA_ATOMIC_REF)
```
MrBurmark: How about adding this in a separate PR as well?

include/RAJA/policy/desul/atomic.hpp
```diff
@@ -153,6 +156,29 @@ atomicCAS(AtomicPolicy, T* acc, T compare, T value)
...
 
+RAJA_SUPPRESS_HD_WARN
+template<typename AtomicPolicy, typename T, typename Operation>
+RAJA_HOST_DEVICE RAJA_INLINE T atomicGeneric(AtomicPolicy,
+                                             T* acc,
+                                             Operation&& operation)
+{
+  static_assert(std::is_trivially_copyable_v<T>);
+
+  T old = desul::atomic_load(acc, raja_default_desul_order {},
```
MrBurmark: This order should always be relaxed or lower.

include/RAJA/policy/cuda/atomic.hpp
adayton1: Done.

include/RAJA/policy/atomic_builtin.hpp
```diff
@@ -1011,6 +986,25 @@ atomicCAS(builtin_atomic, T* acc, T compare, T value)
   return detail::builtin_atomicCAS(acc, compare, value);
 }
 
+template<typename T, typename Operation>
+RAJA_DEVICE_HIP RAJA_INLINE T atomicGeneric(builtin_atomic,
```
MrBurmark, line 990: Should this use `builtin_atomicCAS_loop`?

include/RAJA/policy/desul/atomic.hpp
adayton1: In just the load or the compare_exchange as well?

include/RAJA/util/TypeConvert.hpp
MrBurmark: That would make sense, what requirements does std::bit_cast have?

include/RAJA/policy/desul/atomic.hpp
MrBurmark: Just the load, the CAS will handle the order, we just need a starting value.

adayton1: Done.

include/RAJA/policy/atomic_builtin.hpp
adayton1: Done.

include/RAJA/util/TypeConvert.hpp
MrBurmark: Maybe do that in a separate PR?

adayton1: "This overload participates in overload resolution only if sizeof(To) == sizeof(From) and both To and From are [TriviallyCopyable](https://en.cppreference.com/cpp/named_req/TriviallyCopyable) types."

MrBurmark (approved):

include/RAJA/util/TypeConvert.hpp
adayton1: Done.

adayton1: Oh I missed your comment about doing this in a separate PR. Do you still want me to do that or just leave the change in?

MrBurmark: Let's add it in a subsequent PR. I think we'll want to make this function use concepts/requirements instead of static asserts anyway so we will have to come back to it anyway.

adayton1: OK, I reverted the requirement and added a todo instead.

johnbowen42: We could conceptify this now--the basic infrastructure for concepts is in.  [include/pattern/concepts.hpp ](https://github.com/llnl/RAJA/blob/develop/include/RAJA/pattern/concepts.hpp) has concepts related to execution--policies, Resource, etc.  https://github.com/llnl/RAJA/blob/develop/include/RAJA/util/concepts.hpp would be a good place for `concept TriviallyCopyable = is_trivially_copyable_v<T>;`.  We could also roll the sizeof asserts into the concept (with `decay_t` on the typename parameter)

include/RAJA/policy/desul/atomic.hpp
```diff
@@ -153,6 +156,29 @@ atomicCAS(AtomicPolicy, T* acc, T compare, T value)
...
+                             raja_default_desul_scope {});
+  T expected;
+
+  do
+  {
+    expected = old;
+    old = desul::atomic_compare_exchange(acc, expected, operation(expected),
+                                         raja_default_desul_order {},
+                                         raja_default_desul_scope {});
+  } while (!RAJA::util::bit_equal(old, expected));
```
johnbowen42, line 177: we might get a small performance hit here whenever we perform an operation like `<=`, where early exit is possible upon equality

johnbowen42 (approved):
I have two nits: one where we could replace a static_assert with concepts, one where extra performance could be gained with an early exit condition inside the generic loop.  However I think this is fine as is and don't want to block

include/RAJA/policy/desul/atomic.hpp
adayton1: Yeah, that's likely. I was thinking we could add an overload with short-circuiting in a future PR.

PR #2006, opened 2026-03-25
Python like Range

artv3:
#Summary

Many of my loops start at zero, this gives users a shortcut to generate the desired range segment.


include/RAJA/index/RangeSegment.hpp
```diff
@@ -530,6 +530,33 @@ using RangeSegment = TypedRangeSegment<Index_type>;
 //! Alias for TypedRangeStrideSegment<Index_type>
 using RangeStrideSegment = TypedRangeStrideSegment<Index_type>;
 
+/*!
+ * \brief Function to make a RangeSegment for the interval [0, end)
+ *
+ *  \return a newly constructed RangeSegment over the half-open interval
+ *          starting at zero and ending at @end.
+ */
+RAJA_HOST_DEVICE RAJA_INLINE constexpr RangeSegment ZeroTo(
```
MrBurmark: Can we not just have a single argument constructor? Doesn't python's `range(n)` give you `[0,n)`?

artv3: Partially motivated since I find ``RAJA::RangeSegment`` kind of long to write. `Range()` can work but I liked `ZeroTo` since its explicit.

MrBurmark: Funny, to me `RAJA::RangeSegment(N)` seems obvious. First seeing `RAJA::ZeroTo` was confusing to me as it didn't have the word range in its name and since the first letter is capitalized I assumed it was a different class of some sort. I started thinking that it must be some kind of range that didn't have a begin offset to apply so it was slightly cheaper to use, we've found small performance differences between the base and RAJA variants due to that kind of thing before.

artv3:
Hi @llnl/raja-core I've pushed up the discussed changes.

examples/raja-ranges.cpp
```diff
@@ -0,0 +1,86 @@
...
+{
+  constexpr RAJA::Index_type N = 8;
+
+  int values[N] = {};
+  int typed_values[N] = {};
+  int subrange_values[N] = {};
+  int odd_values[N] = {};
+
+  // Equivalent to Python range(N): [0, N)
+  RAJA::forall<RAJA::seq_exec>(RAJA::Range(N), [&](RAJA::Index_type i) {
```
MrBurmark: If this is a function that returns ranges then it should be `range`.

include/RAJA/index/RangeSegment.hpp
```diff
@@ -530,6 +530,33 @@ using RangeSegment = TypedRangeSegment<Index_type>;
 //! Alias for TypedRangeStrideSegment<Index_type>
 using RangeStrideSegment = TypedRangeStrideSegment<Index_type>;
 
+/*!
+ * \brief Function to make a RangeSegment for the interval [0, end)
+ *
+ *  \return a newly constructed RangeSegment over the half-open interval
+ *          starting at zero and ending at @end.
+ */
+RAJA_HOST_DEVICE RAJA_INLINE constexpr RangeSegment Range(
```
MrBurmark: I would combine these two versions so this happens.
```c++
range(10) -> TypedRangeSegment<int>;
range(long{10}) -> TypedRangeSegment<long>;
range(StrongIndexType(10)) -> TypedRangeSegment<StrongIndexType>;
range<long>(10) -> TypedRangeSegment<long>;
```

include/RAJA/index/RangeSegment.hpp
```diff
@@ -549,6 +576,56 @@ using common_type_t = typename common_type<Ts...>::type;
...
+ */
+template<typename BeginT,
+         typename EndT,
+         typename Common = detail::common_type_t<BeginT, EndT>>
+RAJA_HOST_DEVICE RAJA_INLINE constexpr TypedRangeSegment<Common> Range(
+    BeginT&& begin,
+    EndT&& end) noexcept
+{
+  return {begin, end};
+}
```
MrBurmark: You can probably just take a single type for begin and end so the user can supply a template arg for the type they want, a.k.a. `range<long>(0, 10) -> TypedRangeSegment<long>`. Or possibly have a defaulted first argument that is used for the type if its set or it uses the common type otherwise. Maybe we should look to see how other libraries handle these multi-arg cases.

rhornung67: Yes, that is better. Having two different types for the range bounds doesn't really make sense and could result in compiler warnings resulting from questionable comparisons.

include/RAJA/index/RangeSegment.hpp
```diff
@@ -547,8 +547,101 @@ struct common_type<T>
...
+ *          value_type is equivilent to the common type of
+ *          @begin and @end unless an explicit template argument
+ *          is provided for the segment storage type. If stride is zero,
+ *          execution aborts or throws.
+ */
+template<typename StorageT = void,
+         typename BeginT,
+         typename EndT,
+         typename StrideT,
+         typename Common = detail::range_storage_type_t<StorageT, BeginT, EndT>,
```
MrBurmark: Should `StrideT` be taken into account?

include/RAJA/index/RangeSegment.hpp
```diff
@@ -547,8 +547,101 @@ struct common_type<T>
...
+         typename DiffT = make_signed_t<strip_index_type_t<Common>>>
+RAJA_HOST_DEVICE RAJA_INLINE TypedRangeStrideSegment<Common> range(
+    BeginT&& begin,
+    EndT&& end,
+    StrideT&& stride)
+{
+  static_assert(std::is_signed<strip_index_type_t<StrideT>>::value,
+                "range requires a signed stride type.");
```
MrBurmark: Does it?

include/RAJA/index/RangeSegment.hpp
```diff
@@ -547,8 +547,101 @@ struct common_type<T>
...
+  static_assert(!std::is_floating_point<strip_index_type_t<StrideT>>::value,
```
MrBurmark:
```suggestion
  static_assert(!std::is_floating_point_v<strip_index_type_t<StrideT>>,
```

include/RAJA/index/RangeSegment.hpp
rhornung67: Also, same types for begin, end, and stride.

MrBurmark: I worry about usability here, if you do `range(0, 1u)` then it won't be able to deduce the index type as you gave it 2 different options. I run into this with code like `std::max(working_value, 0)` which I have to change to something like `std::max(working_value, decltype(working_value){0})` to get it to compile.

artv3: yes, I see both sides here.

include/RAJA/index/RangeSegment.hpp
```diff
@@ -533,22 +533,123 @@ using RangeStrideSegment = TypedRangeStrideSegment<Index_type>;
 namespace detail
 {
 
+template<typename T>
+using range_arg_type_t = std::decay_t<T>;
```
MrBurmark: Why not just `decay_t`, its shorter?

include/RAJA/index/RangeSegment.hpp
```diff
@@ -533,22 +533,123 @@ using RangeStrideSegment = TypedRangeStrideSegment<Index_type>;
...
+
 template<typename T, typename... Rest>
 struct common_type
-    : std::common_type<T, typename std::common_type<Rest...>::type>
+    : std::common_type<range_arg_type_t<T>, typename common_type<Rest...>::type>
 {};
```
MrBurmark: Doesn't std::common_type handle the variadic case?

MrBurmark: Also this https://en.cppreference.com/w/cpp/types/common_type.html seems to say that it will do the decay for you.

test/unit/index/test-rangesegment.cpp
```diff
@@ -166,3 +168,144 @@ TYPED_TEST(RangeSegmentUnitTest, Equality)
...
+  auto r = RAJA::range<RangeStrongIndex>(17);
+
+  ASSERT_EQ((RAJA::TypedRangeSegment<RangeStrongIndex>(0, 17)), r);
+  ASSERT_EQ(RangeStrongIndex(0), *r.begin());
+  ASSERT_EQ(RAJA::Index_type(17), r.size());
+}
+
+TEST(RangeSegmentUnitTest, DeducedStrongRangeEnd)
+{
+  auto r = RAJA::range(RangeStrongIndex(17));
```
MrBurmark, line 215: Do we have a test where begin and end are strongly typed?

include/RAJA/index/RangeSegment.hpp
artv3: nice

test/unit/index/test-rangesegment.cpp
```diff
@@ -166,3 +168,158 @@ TYPED_TEST(RangeSegmentUnitTest, Equality)
...
+
+  ASSERT_EQ((RAJA::TypedRangeSegment<int>(3, 17)), r);
+  ASSERT_EQ(14, r.size());
+}
+
+TEST(RangeSegmentUnitTest, StrongRangeBeginEnd)
+{
+  RangeStrongIndex begin(3);
+  RangeStrongIndex end(17);
+  auto r = RAJA::range(begin, end);
```
MrBurmark, line 248: Should we be able to do something like this `range(1, end)` where end is a strongly typed index?

artv3: If we can dream it we can build it.... for we are Team RAJA! 1 .. 2.... 3... RAJA!

docs/sphinx/user_guide/feature/iteration_spaces.rst
```diff
@@ -94,3 +94,68 @@ and two types:
...
+
+The return type depends on the arguments:
+
+* ``RAJA::range(end)`` and ``RAJA::range(begin, end)`` return a
+  ``RAJA::TypedRangeSegment``.
+* ``RAJA::range(begin, end, stride)`` returns a
+  ``RAJA::TypedRangeStrideSegment``.
+* When one of the bounds is a RAJA strong index type, such as a type created
+  with ``RAJA_INDEX_VALUE``, that strong type is preserved for the loop
+  variable when possible.
```
MrBurmark, line 122: When possible? Doesn't it always use the strong type if one is provided?
The only case I'm worried about is something like a strong index type with underlying type int combined with a long choosing long instead of being an error.

docs/sphinx/user_guide/feature/iteration_spaces.rst
```diff
@@ -94,3 +94,68 @@ and two types:
...
+* Providing an explicit template argument, such as
+  ``RAJA::range<MyIndex>(end)``, overrides the deduced storage type.
```
MrBurmark, line 124: It should still be an error to, for example, explicitly ask for int but provide a strong index type as an argument, or explicitly ask for a strong index type and provide a different strong index type as an argument.

MrBurmark: I wonder if we can use c++20's require to test that these cases fail?

trws: You definitely can, you could even make concepts to make it easier most likely.

MrBurmark:
I played with making a range type that could have any of begin, end, and stride be runtime or compile time constants. Is that something that you are interested in @artv3. Right now you construct the range in the lambda so the compiler knows that begin is a constant 0, but if you constructed the range outside of the lambda and captured it then the compiler would not know that begin is a constant. If you had a range type that could have any of the parameters be constant or runtime could fix that.

artv3:
> I played with making a range type that could have any of begin, end, and stride be runtime or compile time constants. Is that something that you are interested in @artv3. Right now you construct the range in the lambda so the compiler knows that begin is a constant 0, but if you constructed the range outside of the lambda and captured it then the compiler would not know that begin is a constant. If you had a range type that could have any of the parameters be constant or runtime could fix that.

Yes absolutely! We are always stronger when we combine our ideas! @rhornung67, @MrBurmark  could this be something we carve out time to discuss at the next RAJA meeting?

MrBurmark:
> Yes absolutely! We are always stronger when we combine our ideas! @rhornung67, @MrBurmark could this be something we carve out time to discuss at the next RAJA meeting?

Here is a basic prototype without support for things like strongly typed indices. https://godbolt.org/z/o4zdcEEdT

MrBurmark:
> > Yes absolutely! We are always stronger when we combine our ideas! @rhornung67, @MrBurmark could this be something we carve out time to discuss at the next RAJA meeting?
> 
> Here is a basic prototype without support for things like strongly typed indices. https://godbolt.org/z/o4zdcEEdT

Regarding making a range where the member variables can be runtime or compile time, that could involve changing the underlying TypedRange class and combining it with the strided version. Maybe we could even unify our integral range types into a single class. Potentially usable in the subviews PR as well #1915. Let's think about doing that as a separate PR.

docs/sphinx/user_guide/feature/iteration_spaces.rst
artv3: that is a good point, I think we should error if so

artv3: I updated the code accordingly

artv3:
I pushed up code that aims to address the comments, my question now though if we are happy with the following: 

```
  - Compile-time rejection for the explicit mismatch cases:
      - RAJA::range<int>(RangeStrongIndex(3), 17) is rejected.
      - RAJA::range<RangeStrongIndex>(AnotherRangeStrongIndex(3), 17) is rejected.
      - The same pattern is enforced for the strided overload.
```

Is this too aggressive or should we allow mixing if the underlying data type is the same? I don't use strongly typed indices so I would be interested in how folks would like to handle these cases

MrBurmark:
> I pushed up code that aims to address the comments, my question now though if we are happy with the following:
> 
> ```
>   - Compile-time rejection for the explicit mismatch cases:
>       - RAJA::range<int>(RangeStrongIndex(3), 17) is rejected.
>       - RAJA::range<RangeStrongIndex>(AnotherRangeStrongIndex(3), 17) is rejected.
>       - The same pattern is enforced for the strided overload.
> ```
> 
> Is this too aggressive or should we allow mixing if the underlying data type is the same? I don't use strongly typed indices so I would be interested in how folks would like to handle these cases

I think these cases should fail, rejecting incorrect index types is the whole point.
Earlier on this PR I asked that mixing one or more non-strongly typed index and one strongly typed index work. I'm not sure is that is a good idea or not, I was trying to let people do this `RAJA::range(1, rsiN)` instead of this `RAJA::range(RangeStrongIndex(1), rsiN)`. It seemed overly verbose, but I'm not sure that is really in the spirit of stringly typed indices or not.

MrBurmark:
Actually thinking about how types are decided, I'm not sure that it makes sense to take the type if the stride into account. If `begin` and `end` are `int` and `stride` is `long` for example we would currently choose `long` for the overall index type. However all of the values would have to be representable as `int`, I suppose the one-past the last value could possibly not be representable, for example with 2-bit ints range(0,3,2) would be problematic as it would contain {0, 2} with 4 being the one-past the end value and not representable. I'm not sure that we want to handle that case by hoping people provide a larger stride type if they plan to have a range like that.

artv3:
> Actually thinking about how types are decided, I'm not sure that it makes sense to take the type if the stride into account. If `begin` and `end` are `int` and `stride` is `long` for example we would currently choose `long` for the overall index type. However all of the values would have to be representable as `int`, I suppose the one-past the last value could possibly not be representable, for example with 2-bit ints range(0,3,2) would be problematic as it would contain {0, 2} with 4 being the one-past the end value and not representable. I'm not sure that we want to handle that case by hoping people provide a larger stride type if they plan to have a range like that.

right, maybe we can bring it up at the next @llnl/raja-core  meeting

MrBurmark:
```cpp
RAJA::range<int>(RangeStrongIndex(3), 17) // is rejected.
RAJA::range<RangeStrongIndex>(AnotherRangeStrongIndex(3), 17) // is rejected.
RAJA::range<RangeStrongIndex>(RangeStrongIndex(3), RangeStrongIndex(17)) // is ok.
RAJA::range<RangeStrongIndex>(RangeStrongIndex(3), 17) // is ok.
RAJA::range<RangeStrongIndex>(3, 17) // is ok.
RAJA::range(RangeStrongIndex(3), RangeStrongIndex(17)) // is ok, chooses RangeStrongIndex.
RAJA::range(RangeStrongIndex(3), 17) // is ok, chooses RangeStrongIndex.
RAJA::range(3, 17) // is ok, chooses int.
RAJA::range(RangeStrongIndex(3), static_cast<long>(17)) // is rejected.
RAJA::range(static_cast<short>(3), 17) // is ok, chooses int.
RAJA::range(3, static_cast<short>(17)) // is ok, chooses int.
```

artv3:
> ```c++
> RAJA::range<int>(RangeStrongIndex(3), 17) rejected.
> RAJA::range<RangeStrongIndex>(AnotherRangeStrongIndex(3), 17) is rejected.
> RAJA::range<RangeStrongIndex>(RangeStrongIndex(3), RangeStrongIndex(17)) is ok.
> RAJA::range<RangeStrongIndex>(RangeStrongIndex(3), 17) is ok?
> RAJA::range<RangeStrongIndex>(3, 17) is ok?
> RAJA::range(RangeStrongIndex(3), RangeStrongIndex(17)) is ok.
> RAJA::range(RangeStrongIndex(3), 17) is ok?
> RAJA::range(3, 17) is ok.
> ```

I'm in favor of supporting those that read "ok" -- let me know if there are objections. The only concern is type narrowing, but I think we should build with `-Werror=narrowing`

rhornung67:
I added warnings as errors to CI improvements issue here: https://github.com/llnl/RAJA/issues/1943

Please add specific items to that list and we will eventually work them in.

MrBurmark:
```cpp
RAJA::range<int>(3, 17, 1) // is ok
RAJA::range<int>(3, 17, static_cast<short>(1))  // is ok
RAJA::range<int>(3, 17, static_cast<long>(1))  // is rejected? or is ok?
RAJA::range<int>(3, 17, RangeStrongIndex(1))  // is rejected
RAJA::range<RangeStrongIndex>(3, 17, 1)  // is ok.
RAJA::range<RangeStrongIndex>(RangeStrongIndex(3), RangeStrongIndex(17), 1)  // is ok.
RAJA::range<RangeStrongIndex>(RangeStrongIndex(3), RangeStrongIndex(17), RangeStrongIndex(1))  // is ok.
RAJA::range(3, 17, 1) // is ok, chooses int.
RAJA::range(3, 17, static_cast<short>(1))  // is ok, chooses int.
RAJA::range(3, 17, static_cast<long>(1))  // is rejected? or is ok, chooses int?
RAJA::range(3, 17, RangeStrongIndex(1))  // is rejected
RAJA::range(RangeStrongIndex(3), RangeStrongIndex(17), 1)  // is ok, chooses RangeStrongIndex.
RAJA::range(RangeStrongIndex(3), 17, 1)  // is ok, chooses RangeStrongIndex.
```

artv3:
> ```c++
> RAJA::range(3, 17, static_cast<long>(1))  // is rejected? or is ok, chooses int?
> ```

this would be type narrowing which I think should throw an error

trws:
> ```c++
> RAJA::range(3, 17, static_cast<long>(1))  // is rejected? or is ok, chooses int?
> ```
> this would be type narrowing which I think should throw an error
> 

I've mostly been just lurking here since it looks like you've hit everything I expected @artv3, also I love seeing this idea happen great stuff, but this surprises me.  Is it not using `std::common_type<Ts...>` to get the type to use so that it will use the largest of the specified integral types?

artv3:
> > ```c++
> > RAJA::range(3, 17, static_cast<long>(1))  // is rejected? or is ok, chooses int?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this would be type narrowing which I think should throw an error
> 
> I've mostly been just lurking here since it looks like you've hit everything I expected @artv3, also I love seeing this idea happen great stuff, but this surprises me. Is it not using `std::common_type<Ts...>` to get the type to use so that it will use the largest of the specified integral types?

oh thats right that would happen, but then is that we we want? to promote (3,17) to longs?

trws:
Speaking just for me it's what I would want. As long as all three are convertible, which I think common_type requires anyway, it makes it much easier to be sure it's correct. I could easily see someone using the range(1, groups::size) construction as self documenting range and if the upper is an int64_t or unsigned or whatever then it just does the right thing. If they aren't losslessly convertible, like the strongly typed indices, then failing makes sense though.

MrBurmark:
@trws We do use common type across all of the arguments to get a type for the range if you don't specify one. My question for `RAJA::range(3, 17, static_cast<long>(1))` is what does it mean if stride is of a larger type than begin and end? In general all of the values in the range have to representable as the common type of begin and end. Stride tells you which of those values to use and in which order they appear. So does it make sense for the type of stride to affect the type used by the range?

trws:
Honestly I think it does, at least in as much as when `iter + stride` is executed we want the result to be of the same type as `iter`.  I suppose it wouldn't _have_ to be, but if not it would have to be convertible to the type of iter and that seems a bit harder to reason about. Also even if we actually do re-order them based on the sign of the stride, in a logical sense I think of it more like:

```c++
for (std::common_type_t<Begin,End,Stride> iter = begin; iter < end; iter += stride)
```

artv3:
```
  RAJA::range<int>(3, 17, 1)                         // OK: storage int
  RAJA::range<int>(3, 17, short{1})                  // OK: storage int
  RAJA::range<int>(3, 17, long{1})                   // OK, not rejected: storage int, diff long
  RAJA::range<int>(3, 17, RangeStrongIndex(1))       // rejected

  RAJA::range<RangeStrongIndex>(3, 17, 1)            // OK
  RAJA::range<RangeStrongIndex>(RSI(3), RSI(17), 1)  // OK
  RAJA::range<RangeStrongIndex>(RSI(3), RSI(17), RSI(1)) // OK

  RAJA::range(3, 17, 1)                              // OK: chooses int
  RAJA::range(3, 17, short{1})                       // OK: chooses int
  RAJA::range(3, 17, long{1})                        // OK: chooses long, not int
  RAJA::range(3, 17, RangeStrongIndex(1))            // rejected

  RAJA::range(RSI(3), RSI(17), 1)                    // OK: chooses RangeStrongIndex
  RAJA::range(RSI(3), 17, 1)                         // OK: chooses RangeStrongIndex
```

My latest push should support this matrix

adrienbernede (commented):
@lc-hubcast approve

include/RAJA/index/RangeSegment.hpp
```diff
@@ -533,22 +534,367 @@ using RangeStrideSegment = TypedRangeStrideSegment<Index_type>;
...
-    : std::common_type<T, typename std::common_type<Rest...>::type>
+template<typename... Ts>
+using common_type_t = std::common_type_t<Ts...>;
+
+struct no_strong_index
 {};
 
 template<typename T>
-struct common_type<T>
+using strong_index_candidate_t =
```
johnbowen42, line 544: can use this type trait instead of `std:is_base_of`.  https://github.com/llnl/RAJA/blob/6e4fe62d810711a0af9020d4e94c6c41c9a6117b/include/RAJA/index/IndexValue.hpp#L306

johnbowen42: at some point we should decide if we want the `std::decay_t` inside of the type_traits, or if we want the user to handle that decision.  I prefer the latter

include/RAJA/index/RangeSegment.hpp
```diff
@@ -533,22 +534,367 @@ using RangeStrideSegment = TypedRangeStrideSegment<Index_type>;
...
+};
+
+template<typename T, typename... Ts>
+struct strong_index_type<T, Ts...>
+{
+  using current = strong_index_candidate_t<T>;
+  using rest    = typename strong_index_type<Ts...>::type;
+
+  static_assert(
+      std::is_same_v<current, no_strong_index> ||
```
johnbowen42, line 562: A more "C++20" approach would be to use https://github.com/llnl/RAJA/blob/6e4fe62d810711a0af9020d4e94c6c41c9a6117b/include/RAJA/index/IndexValue.hpp#L317, but then you'd need to add `std::decay_t` to the type trait defined there

include/RAJA/index/RangeSegment.hpp
```diff
@@ -575,23 +924,64 @@ RAJA_HOST_DEVICE TypedRangeSegment<Common> make_range(BeginT&& begin,
...
-template<typename BeginT,
-         typename EndT,
-         typename StrideT,
-         typename Common = detail::common_type_t<BeginT, EndT>>
-RAJA_HOST_DEVICE TypedRangeStrideSegment<Common> make_strided_range(
-    BeginT&& begin,
-    EndT&& end,
-    StrideT&& stride)
+template<
+    typename BeginT,
```
johnbowen42:
```suggestion
    concepts::IndexValued BeginT,
```

include/RAJA/index/RangeSegment.hpp
```diff
@@ -575,23 +924,64 @@ RAJA_HOST_DEVICE TypedRangeSegment<Common> make_range(BeginT&& begin,
...
+    typename EndT,
+    typename StrideT,
+    typename Common =
+        detail::deduced_range_stride_storage_type_t<BeginT, EndT, StrideT>,
+    typename DiffT = detail::deduced_range_stride_diff_type_t<Common, StrideT>>
+  requires detail::
```
johnbowen42, line 932: IIRC, if we made the concepts change, I think the equivalent `requires` would be just a `std::is_same_v`?

include/RAJA/index/RangeSegment.hpp
```diff
@@ -575,23 +924,64 @@ RAJA_HOST_DEVICE TypedRangeSegment<Common> make_range(BeginT&& begin,
...
-         typename Common = detail::common_type_t<BeginT, EndT>>
-RAJA_HOST_DEVICE TypedRangeStrideSegment<Common> make_strided_range(
-    BeginT&& begin,
-    EndT&& end,
-    StrideT&& stride)
+template<
+    typename BeginT,
+    typename EndT,
+    typename StrideT,
+    typename Common =
```
johnbowen42: I believe this could be eliminated with the concept addition

include/RAJA/index/RangeSegment.hpp
```diff
@@ -575,23 +924,64 @@ RAJA_HOST_DEVICE TypedRangeSegment<Common> make_range(BeginT&& begin,
...
+  return {static_cast<strip_index_type_t<Common>>(stripIndexType(begin)),
+          static_cast<strip_index_type_t<Common>>(stripIndexType(end)),
+          static_cast<DiffT>(stripIndexType(stride))};
 }
 
+namespace concepts
+{
+
+template<typename T, typename U>
+concept RangeConstructible =
```
johnbowen42: Let's move these to https://github.com/llnl/RAJA/blob/develop/include/RAJA/pattern/concepts.hpp, or to https://github.com/llnl/RAJA/blob/develop/include/RAJA/util/concepts.hpp just to keep all the core concepts consolidated

include/RAJA/index/RangeSegment.hpp
artv3: maybe we can do that in a follow up PR

include/RAJA/index/RangeSegment.hpp
artv3: I think that only holds if we are checking strong index types and we want to verify that they have the same underlying type

include/RAJA/index/RangeSegment.hpp
artv3: I don't think I see it, the right hand side of Common looks for a common type while concept checks requirements from the type?

artv3:
@johnbowen42 I took a pass at addressing your comments, let me know if there are still things to discuss.

include/RAJA/index/RangeSegment.hpp
artv3: done

include/RAJA/index/RangeSegment.hpp
johnbowen42: I was saying that adding the `concepts::IndexValued` checks to `BeginT/EndT` and then also have `requires std::is_same_v<BeginT, EndT>` should be equivalent, right?

include/RAJA/index/RangeSegment.hpp
johnbowen42: See my other comment, I clarified what I meant

artv3:
@llnl/raja-core , is this ready to go?

test/unit/index/test-rangesegment.cpp
```diff
@@ -15,18 +15,33 @@
...
+};
+
 template<typename T>
 class RangeSegmentUnitTest : public ::testing::Test {};
 
 TYPED_TEST_SUITE(RangeSegmentUnitTest, UnitIndexTypes);
 
 
-template< typename T, typename std::enable_if<std::is_unsigned<T>::value>::type* = nullptr>
+template<typename T, typename std::enable_if_t<std::is_unsigned_v<T>>* = nullptr>
```
bechols97: Would it make more sense to use concepts here?

```suggestion
template<std::unsigned_integral T>
```

test/unit/index/test-rangesegment.cpp
```diff
@@ -15,18 +15,33 @@
...
 void NegativeRangeSegConstructorsTest()
 {
 }
 
-template< typename T, typename std::enable_if<std::is_signed<T>::value>::type* = nullptr>
+template<typename T, typename std::enable_if_t<std::is_signed_v<T>>* = nullptr>
```
bechols97: Same thing here 

```suggestion
template<std::signed_integral T>
```

test/unit/index/test-rangesegment.cpp
```diff
@@ -76,12 +91,12 @@ TYPED_TEST(RangeSegmentUnitTest, Swaps)
...
 }
 
-template< typename T, typename std::enable_if<std::is_unsigned<T>::value>::type* = nullptr>
+template<typename T, typename std::enable_if_t<std::is_unsigned_v<T>>* = nullptr>
 void NegativeRangeSegIteratorsTest()
 {
 }
 
-template< typename T, typename std::enable_if<std::is_signed<T>::value>::type* = nullptr>
+template<typename T, typename std::enable_if_t<std::is_signed_v<T>>* = nullptr>
```
bechols97:
```suggestion
template<std::signed_integral T>
```

test/unit/index/test-rangesegment.cpp
```diff
@@ -76,12 +91,12 @@ TYPED_TEST(RangeSegmentUnitTest, Swaps)
   ASSERT_EQ(r2, r3);
 }
 
-template< typename T, typename std::enable_if<std::is_unsigned<T>::value>::type* = nullptr>
+template<typename T, typename std::enable_if_t<std::is_unsigned_v<T>>* = nullptr>
```
bechols97:
```suggestion
template<std::unsigned_integral T>
```

artv3: good call, updated across the board.

docs/sphinx/user_guide/feature/iteration_spaces.rst
```diff
@@ -99,3 +99,75 @@ and two types:
...
+  RAJA::range(begin, end)          // [begin, end)
+  RAJA::range(begin, end, stride)  // [begin, end) with stride
+
+The return type depends on the arguments:
+
+* ``RAJA::range(end)`` and ``RAJA::range(begin, end)`` return a
+  ``RAJA::TypedRangeSegment``.
+* ``RAJA::range(begin, end, stride)`` returns a
+  ``RAJA::TypedRangeStrideSegment``.
+* When one of the bounds is a RAJA strong index type, such as a type created
```
adayton1, line 132: Are this point and the next one still completely true?

docs/sphinx/user_guide/feature/iteration_spaces.rst
```diff
@@ -99,3 +99,75 @@ and two types:
...
+  });
+
+  RAJA::forall<RAJA::seq_exec>(RAJA::range(2, 6), [=](int i) {
+    subrange_values[i] = i;
+  });
+
+  RAJA::forall<RAJA::seq_exec>(RAJA::range(CellIndex {1}, CellIndex {N},
+                                            CellIndex {2}),
+                               [=](CellIndex i) {
+                                 strided_values[*i] = *i;
```
adayton1, line 172: Why dereference `i`?

adayton1: If it's some feature about the CellIndex type created by the RAJA_INDEX_VALUE macro, then it might be worth mentioning in the documentation.

test/unit/index/test-rangesegment.cpp
```diff
@@ -166,3 +182,318 @@ TYPED_TEST(RangeSegmentUnitTest, Equality)
...
+  auto r = RAJA::range(RAJA::Index_type(17));
+
+  ASSERT_EQ(RAJA::RangeSegment(RAJA::Index_type(0), RAJA::Index_type(17)), r);
+  ASSERT_EQ(RAJA::Index_type(0), *r.begin());
+  ASSERT_EQ(RAJA::Index_type(17), r.size());
+}
+
+TEST(RangeSegmentUnitTest, LongRangeEnd)
+{
+  auto r = RAJA::range(long {10});
```
adayton1: You could use `10L` instead.

test/unit/index/test-rangesegment.cpp
```diff
@@ -166,3 +182,318 @@ TYPED_TEST(RangeSegmentUnitTest, Equality)
...
+  ASSERT_EQ(0, *r.begin());
+  ASSERT_EQ(10, r.size());
+}
+
+TEST(RangeSegmentUnitTest, RangeBeginEnd)
+{
+  auto r = RAJA::range(3, 17);
+
+  ASSERT_EQ((RAJA::TypedRangeSegment<int>(3, 17)), r);
+  ASSERT_EQ(14, r.size());
```
adayton1, line 261: You should also check what is returned by `*r.begin()`

test/unit/index/test-rangesegment.cpp
```diff
@@ -166,3 +182,318 @@ TYPED_TEST(RangeSegmentUnitTest, Equality)
...
+static_assert(
+    TypedRangeCallable<RangeStrongIndex, RangeStrongIndex, RangeStrongIndex>);
+static_assert(!TypedRangeCallable<int, RangeStrongIndex, int>);
+static_assert(!TypedRangeCallable<RangeStrongIndex, RangeStrongIndex, int>);
+static_assert(
+    !TypedRangeCallable<RangeStrongIndex, AnotherRangeStrongIndex, int>);
+
+TEST(RangeSegmentUnitTest, MixedStrongRangeStrongBegin)
+{
+  static_assert(!RangeCallable<RangeStrongIndex, int>);
```
adayton1: Why are there some tests with only a static_assert while other static_assert calls are outside of all tests?

test/unit/index/test-rangesegment.cpp
```diff
@@ -166,3 +182,318 @@ TYPED_TEST(RangeSegmentUnitTest, Equality)
...
+  ASSERT_EQ(RangeStrongIndex(3), *r.begin());
+  ASSERT_EQ(RAJA::Index_type(14), r.size());
+}
+
+TEST(RangeSegmentUnitTest, MixedStrongRangeBeginEnd)
+{
+  static_assert(!RangeCallable<int, RangeStrongIndex>);
+}
+
+static_assert(RAJA::concepts::RangeConstructible<RangeStrongIndex,
```
adayton1, line 280: Why do we need both RangeStrideConstructible and RangeCallable?

test/unit/index/test-rangesegment.cpp
```diff
@@ -166,3 +182,318 @@ TYPED_TEST(RangeSegmentUnitTest, Equality)
...
+  auto r = RAJA::range(2, 11, 3);
+
+  ASSERT_EQ((RAJA::TypedRangeStrideSegment<int>(2, 11, 3)), r);
+  ASSERT_EQ(3, r.size());
+  ASSERT_EQ(2, *r.begin());
+}
+
+TEST(RangeSegmentUnitTest, MixedStrongRangeBeginEndStrideStrongEnd)
+{
+  auto r = RAJA::range(2, RangeStrongIndex(11), 3);
```
adayton1: It seems that the documentation and some test cases and/or static_asserts say you can't mix types like this in some calls to RAJA::range. Don't we want to be consistent?

test/unit/index/test-rangesegment.cpp
```diff
@@ -166,3 +182,318 @@ TYPED_TEST(RangeSegmentUnitTest, Equality)
...
+  auto r_long  = RAJA::range<int>(3, 17, static_cast<long>(1));
+
+  static_assert(
+      std::is_same_v<decltype(r_int), RAJA::TypedRangeStrideSegment<int>>,
+      "range<int>(begin, end, int stride) should use int storage.");
+  static_assert(
+      std::is_same_v<decltype(r_short), RAJA::TypedRangeStrideSegment<int>>,
+      "range<int>(begin, end, short stride) should use int storage.");
+  static_assert(
+      std::is_same_v<typename decltype(r_long)::value_type, int>,
```
adayton1, line 382: I thought we were taking the common type, which would be long?

adayton1: Also, why does this static_assert differ from the others?

test/unit/index/test-rangesegment.cpp
```diff
@@ -166,3 +182,318 @@ TYPED_TEST(RangeSegmentUnitTest, Equality)
...
+  ASSERT_EQ(3, *r.begin());
+}
+
+TEST(RangeSegmentUnitTest, RangeBeginEndUnsignedStrideWorks)
+{
+  auto r = RAJA::range(-2, 11, 3u);
+
+  static_assert(
+      std::is_same_v<decltype(r), RAJA::TypedRangeStrideSegment<int>>,
+      "range(begin, end, unsigned stride) should preserve signed storage when "
```
adayton1: Should this case just be disallowed?

docs/sphinx/user_guide/feature/iteration_spaces.rst
artv3: Indeed and you are right, I think missing it that part in the RAJA documentation. Any preference to doing:  Use *idx, or RAJA::stripIndexType(idx), ? @rchen20 What is the practice in your application?

include/RAJA/util/concepts.hpp
```diff
@@ -39,6 +40,35 @@ concept BinaryFunction = std::is_invocable_r_v<Return, Function&, Arg1, Arg2>;
...
+     std::is_same_v<std::remove_cvref_t<T>, std::remove_cvref_t<StrongT>>) ||
+    (!IndexValued<T> &&
+     std::is_integral_v<strip_index_type_t<std::remove_cvref_t<T>>>);
+
+template<typename BeginT, typename EndT, typename StrideT>
+concept RangeStrideConstructible =
+    (!IndexValued<BeginT> && !IndexValued<EndT> && !IndexValued<StrideT> &&
+     requires {
+       typename std::common_type_t<std::remove_cvref_t<BeginT>,
+                                   std::remove_cvref_t<EndT>>;
```
adayton1: What about StrideT?

test/unit/index/test-rangesegment.cpp
artv3: I thought it was we take the templated type for begin and end, while we allow a different stride/difference type -- I know we went back and forth on this. And in this case: 

  Without explicit storage:

  RAJA::range(3, 17, 1L)

  then yes, the deduced storage type is long

test/unit/index/test-rangesegment.cpp
artv3: good call, I can see that opening the door to confusing cases.

docs/sphinx/user_guide/feature/iteration_spaces.rst
artv3: ah this is stale now, thanks

test/unit/index/test-rangesegment.cpp
artv3: sounds good

test/unit/index/test-rangesegment.cpp
artv3: yes this needs cleaning up, will push up

test/unit/index/test-rangesegment.cpp
artv3: This might be extra, but RangeStrideConstructible will check validity in combinations while RangeCallable will check if RAJA::range will accept the call

test/unit/index/test-rangesegment.cpp
artv3: yes good call, the combinations made it tricky to keep up with

include/RAJA/util/concepts.hpp
artv3: yes! my favorite, will add

docs/sphinx/user_guide/feature/iteration_spaces.rst
```diff
@@ -99,3 +99,85 @@ and two types:
...
+  RAJA::range(begin, end, stride)  // [begin, end) with stride
+
+The return type depends on the arguments:
+
+* ``RAJA::range(end)`` and ``RAJA::range(begin, end)`` return a
+  ``RAJA::TypedRangeSegment``.
+* ``RAJA::range(begin, end, stride)`` returns a
+  ``RAJA::TypedRangeStrideSegment``.
+* When one of the bounds is a RAJA strong index type, such as a type created
+  with ``RAJA_INDEX_VALUE``, all bounds must use that same strong type. Mixed
```
adayton1, line 133: I think it would be helpful to allow the following mixed case:

`RAJA::range(0, CellIndex{N}, 2)`

If there is no stride, then obviously you could shorten to `RAJA::range(CellIndex{N})`, but when you have a stride and you need a starting value, it would be nice to throw in 0 without having to wrap it since that is probably one of the most common use cases.

docs/sphinx/user_guide/feature/iteration_spaces.rst
```diff
@@ -99,3 +99,85 @@ and two types:
...
+RAJA also provides ``RAJA::range(...)`` helpers that construct the segment
+type for common half-open iteration patterns. These helpers are intended to
+mirror the shape of Python's ``range`` while returning RAJA segment objects
+that can be passed directly to ``RAJA::forall`` and other execution
+interfaces.
+
+The supported forms are::
+
+  RAJA::range(end)                 // [0, end)
+  RAJA::range<IndexT>(end)         // [0, end) with explicit storage type
```
adayton1: The other overloads also allow explicit storage type, right? I think I would organize like so:

```
RAJA::range(end)                 // [0, end)
RAJA::range(begin, end)          // [begin, end)
RAJA::range(begin, end, stride)  // [begin, end) with stride

The storage type can be explicitly specified if desired, and the arguments will be converted to that storage type. If the argument type is not convertible to the storage type, a compilation error will occur.

RAJA::range<IndexT>(end)                 // [0, end)
RAJA::range<IndexT>(begin, end)          // [begin, end)
RAJA::range<IndexT>(begin, end, stride)  // [begin, end) with stride
```

docs/sphinx/user_guide/feature/iteration_spaces.rst
```diff
@@ -99,3 +99,85 @@ and two types:
...
+Python-like Range Helpers
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+RAJA also provides ``RAJA::range(...)`` helpers that construct the segment
+type for common half-open iteration patterns. These helpers are intended to
+mirror the shape of Python's ``range`` while returning RAJA segment objects
+that can be passed directly to ``RAJA::forall`` and other execution
+interfaces.
+
+The supported forms are::
```
adayton1, line 112: Can you remove the extra colon?

rhornung67: The double colon means that the following indented lines will be formatted as code.

docs/sphinx/user_guide/feature/iteration_spaces.rst
adayton1: I think `*i` makes sense for this example now that you've explained it. To me it seems like `stripIndexType` is better suited to template metaprogramming since it could dereference a strong integer type or use a regular integer type directly.

include/RAJA/index/RangeSegment.hpp
```diff
@@ -533,22 +534,378 @@ using RangeStrideSegment = TypedRangeStrideSegment<Index_type>;
...
+        detail::selected_range_storage_t<StorageT, detail::common_type_t<EndT>>,
+    typename StripCommon = strip_index_type_t<Common>>
+  requires detail::range_storage_compatible_v<StorageT, EndT, EndT>
+RAJA_HOST_DEVICE RAJA_INLINE constexpr TypedRangeSegment<Common> range(
+    EndT&& end) noexcept
+{
+  static_assert(!std::is_floating_point_v<StripCommon>,
+                "range requires a non-floating point index type.");
+  return {StripCommon {0}, static_cast<StripCommon>(stripIndexType(end))};
+}
```
adayton1: This implementation seems over-complicated to me and there are so many type traits that are needed to back this all. I think it might actually be simpler to have two overloads (e.g. `range(end)` and `range<StorageT>(end)`).

```
template<concepts::IndexValued EndT>
  requires detail::range_storage_compatible_v<EndT, EndT, EndT>
RAJA_HOST_DEVICE RAJA_INLINE constexpr auto range(EndT&& end) noexcept
{
  return TypedRangeSegment<EndT>{static_cast<EndT>(0), end};
}

template<typename StorageT,
                concepts::IndexValued EndT>
  requires detail::range_storage_compatible_v<StorageT, EndT, EndT>
RAJA_HOST_DEVICE RAJA_INLINE constexpr auto range(EndT&& end) noexcept
{
  using CommonType = std::common_type_t<StorageT, EndT>;
  return TypedRangeSegment<CommonType>{static_cast<CommonType>(0), static_cast<CommonType>(end)};
}
```

include/RAJA/index/RangeSegment.hpp
```diff
@@ -559,11 +916,14 @@ using common_type_t = typename common_type<Ts...>::type;
...
          typename EndT,
-         typename Common = detail::common_type_t<BeginT, EndT>>
+         typename Common = detail::deduced_range_storage_type_t<BeginT, EndT>,
+         typename StripCommon = strip_index_type_t<Common>>
+  requires detail::deduced_range_storage_compatible_v<BeginT, EndT>
 RAJA_HOST_DEVICE TypedRangeSegment<Common> make_range(BeginT&& begin,
                                                       EndT&& end)
 {
-  return {begin, end};
+  return {static_cast<StripCommon>(stripIndexType(begin)),
```
adayton1: Do we really have to strip the index type? Doesn't the TypedRangeSegment constructor accept arguments of the common type?

include/RAJA/index/RangeSegment.hpp
```diff
@@ -533,22 +534,378 @@ using RangeStrideSegment = TypedRangeStrideSegment<Index_type>;
...
+ *
+ *  \return a newly constructed TypedRangeSegment where the
+ *          value_type is equivilent to the common type of
+ *          @begin and @end unless an explicit template argument
+ *          is provided for the segment storage type.
+ */
+template<typename StorageT = void,
+         typename BeginT,
+         typename EndT,
+         typename Common = detail::selected_range_storage_t<
```
adayton1: Instead of making Common and StripCommon template parameters, use auto for the function return type, put these as type aliases in the function body, then explicitly return TypedRangeSegment<Common>. I think that will make this much easier to understand and probably faster to compile.

adayton1: This comment applies in a bunch of places, not just here.

test/unit/index/test-rangesegment.cpp
adayton1: To me it seems like this has the potential for testing the wrong thing or artificially passing tests. But I'm probably missing something.

docs/sphinx/user_guide/feature/iteration_spaces.rst
artv3: Here it makes it a code block -- nay on the code block?

include/RAJA/index/RangeSegment.hpp
adayton1: But this shouldn't hold up this PR.

docs/sphinx/user_guide/feature/iteration_spaces.rst
adayton1: Didn't realize that made a code block.

include/RAJA/index/RangeSegment.hpp
artv3: This mostly for preserving the strong-index behavior and rejecting invalid conversions consistently across the range(end), range(begin, end), and strided forms

docs/sphinx/user_guide/feature/iteration_spaces.rst
rhornung67: Yup. There are multiple ways. This is the least verbose for a small snippet.

include/RAJA/index/RangeSegment.hpp
artv3: For a strong index storage type, the constructor expects the underlying integral value, not the
  strong wrapper. stripIndexType(...) is what makes that work uniformly for both plain integral and
  strong-index arguments.

include/RAJA/index/RangeSegment.hpp
artv3: deal!

docs/sphinx/user_guide/feature/iteration_spaces.rst
artv3: Discussed in meeting -- we expect strongly typed loop bounds to be of the same strong type

include/RAJA/index/RangeSegment.hpp
adayton1: I see you moved StripCommon and DiffT which is already a lot better. You should be able to make Common an alias in the function body as well.

adayton1: Thanks! This is a lot more readable!

adayton1 (approved):

artv3:
thanks @adayton1  -- onto merging!

docs/sphinx/user_guide/feature/iteration_spaces.rst
```diff
@@ -99,3 +99,92 @@ and two types:
...
+chosen storage type, compilation fails. The explicit-type forms are::
+
+  RAJA::range<IndexT>(end)                 // [0, end)
+  RAJA::range<IndexT>(begin, end)          // [begin, end)
+  RAJA::range<IndexT>(begin, end, stride)  // [begin, end) with stride
+
+The return type depends on the arguments:
+
+* ``RAJA::range(end)`` and ``RAJA::range(begin, end)`` return a
+  ``RAJA::TypedRangeSegment``.
```
rhornung67:
```suggestion
  ``RAJA::TypedRangeSegment<IndexT>``.
```

Better to be complete and explicit here (and below)?

docs/sphinx/user_guide/feature/iteration_spaces.rst
```diff
@@ -99,3 +99,92 @@ and two types:
...
+                                            CellIndex {2}),
+                               [=](CellIndex i) {
+                                 strided_values[*i] = *i;
+                               });
+
+Strided ranges follow the same half-open interval convention as
+``RAJA::TypedRangeStrideSegment``. The stride argument must have a signed
+integral type. Positive strides move forward, and negative strides move
+backward. For example, ``RAJA::range(N - 1, -1, -2)`` visits ``N - 1,
+N - 3, ...`` down to the first value that remains greater than ``-1``. A zero
```
rhornung67:
```suggestion
N - 3, ...`` down to the last value in the sequence that is greater than ``-1``. A zero
```

rhornung67 (changes requested):
Some comments in the docs

adayton1:
Source code comments with the purpose of each new concept and type trait would be nice.

docs/sphinx/user_guide/feature/iteration_spaces.rst
```diff
@@ -126,7 +126,7 @@ chosen storage type, compilation fails. The explicit-type forms are::
 The return type depends on the arguments:
 
 * ``RAJA::range(end)`` and ``RAJA::range(begin, end)`` return a
-  ``RAJA::TypedRangeSegment``.
+  ``RAJA::TypedRangeSegment<IndexT>``.
 * ``RAJA::range(begin, end, stride)`` returns a
   ``RAJA::TypedRangeStrideSegment``.
```
rhornung67: Line 131 should also have the template type specialization. I wasn't explicit about each of them in my earlier comment.

rhornung67 (changes requested):
One more. Sorry.

docs/sphinx/user_guide/feature/iteration_spaces.rst
artv3: thanks, I think I got them all now

rhornung67 (approved):

artv3:
@lc-hubcast approve

PR #2008, opened 2026-04-01
alloc optimization

adrienbernede:
This PR switches off shared allocation for CI jobs. Each job has its own allocation and time limit.
This adds more jobs to the CI queue, but simplifies the interpretation of failures.

For Dane and Matrix, we use `--exclusive=user` allocation, which may lead to oversubscription between jobs of the same user, which is good in CI context (except for performance analysis). We hope to improve the throughput with this option, however, it may require finer tuning of allocations to specify that they require less than a full node.

Finally, this PR turns tests back on for the matrix job.

### DIFF
--- a/.gitlab/custom-variables.yml
+++ b/.gitlab/custom-variables.yml
@@ -47,7 +47,7 @@ variables:
 
   # Tioga (flux) allocation settings
   TIOGA_SHARED_ALLOC: "OFF"
-  TIOGA_JOB_ALLOC: "--queue=pci --nodes=1 --timelimit=30m --begin-time=+5s"
+  TIOGA_JOB_ALLOC: "--queue=pci --nodes=1 --time-limit=30m --begin-time=+5s"
   # Project specific variants for tioga
   PROJECT_TIOGA_VARIANTS: "~shared +openmp +vectorization +tests"
   # Project specific deps for tioga
--- a/.gitlab/custom-variables.yml
+++ b/.gitlab/custom-variables.yml
@@ -54,8 +54,8 @@ variables:
   PROJECT_TIOGA_DEPS: ""
 
   # Tuolumne (flux) allocation settings
-  TUOLUMNE_SHARED_ALLOC: "--exclusive --queue=pci --time-limit=90m --nodes=1 -o per-resource.count=2"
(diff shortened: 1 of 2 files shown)


rhornung67 (approved):
@adrienbernede we frequently run into timeouts for the amdclang/HIP jobs on tuolumne. Not sure why this happens since the other jobs on tuo almost never timeout. Can any changes be made to that to make timeouts less frequent?

adrienbernede:
@rhornung67 Noted:
I adapted the matrix allocation to actual usage.
I turned off allocation sharing on Tuolumne due to suspicion that this is causing longer wait times to get in.

To answer your questions: within a shared allocation, it is possible that the HIP jobs are scheduled last and so appear to timeout more often. That means we likely needed a larger shared allocation (due to execution time variability across the different jobs).

By not using the shared allocation, we will have each job run independently on Tuolumne. If a job times out, we’ll be able to add more time for that job specifically.

rhornung67:
When jobs time out, I can usually see others running in the CI queues. Is the 3hour time limit waiting in the queue an LC/GitLab policy. It would be better, if the jobs stayed in the queue until they ran.

adrienbernede:
Several things:
- The 3 hours time out is a GitLab runner timeout. If the shared alloc job times out, the build and test jobs will not run.
- Once the shared alloc is granted, it may happen that the build and test jobs take too much time, and the last in queue may fail because the shared alloc expire.
  - An inner job that already started will be killed.
  - An inner job that did not start yet will try to get an allocation on its own. But in that particular case, the requested allocation will not be in the CI queue because flux does not allow me to specify the ci queue for an inner job (the shared alloc itself is in the CI queue). I’ve tried to address it without success, but that’s a corner cases: all inner jobs should fit in the shared alloc.
 
CI passed without using the shared alloc on Tuolumne. Now it’s on the CI queue to handle the load.
Would you like me to also stop using the shared alloc on other machine to simplify the configuration and the interpretation of the failures ?

rhornung67:
Yes, let's try that. Removing the use of the shared alloc may solve part of the wait time disparity we observe between Axom CI and RAJA CI (RAJA is almost a factor of 10 slower between submit and job start on average). It also gets rid of the issue where other jobs in a shared alloc are killed if one timesout.

PR #2009, opened 2026-04-02
Add RAJA::mask for RAJA::launch

artv3:
# Summary
 RAJA::mask<Policy>(ctx, body) is the per-team single-thread helper intended for setup work before ctx.teamSync()
  - Backend behavior is aligned so the body runs only when the relevant thread-local indices are zero
  
  Supported across all launch backends


examples/raja-launch.cpp
```diff
@@ -162,9 +162,12 @@ int main(int RAJA_UNUSED_ARG(argc), char **RAJA_UNUSED_ARG(argv[]))
            // Array shared within threads of the same team
            RAJA_TEAM_SHARED int s_A[1];
 
-           RAJA::loop<threads_x>(ctx, RAJA::RangeSegment(0, 1), [&](int c) {
+           // __once_loop_start
+           // Use a single logical thread per team for shared initialization.
+           RAJA::loop<threads_x>(ctx, RAJA::once(), [&](int c) {
```
MrBurmark, line 167: Does it make more sense to have a once policy?

artv3: and have loop not take an iteration space?

MrBurmark: Sure, is that consistent with the use case?

MrBurmark: Or have a non-loop function like loop?
```c++
RAJA::non_loop(ctx, [&]() {...});
```

trws:
As a stylistic note, we do this in OpenMP with `single` if the other threads should wait, or with `masked` or the deprecated `master` if they shouldn't, while `once` is used by C++, C, and posix to mean "run exactly one time, no matter how many threads encounter this, and block all encountering threads until that one run is done".

It might be worth using an alternate name to disambiguate it from that "once" behavior, possibly also to indicate the blocking or non-blocking behavior.

examples/raja-launch.cpp
MrBurmark: I missed that you only want one thread doing this so you really do need the proper policy. Do you need index c?

artv3: thats a good point, the unchecked policies would not work in this case .

MrBurmark: Well they would work in the same cases where the unchecked policies would work anyway, but that's besides the point.
I was confused and thought you just wanted a dummy loop that didn't mask off any threads. But you do want a real loop that masks off all of the threads but one. So my question now is should this really be a loop or a more specialized construct, hence do you actually need `c`?

artv3: Right we can drop c, and yes basically its for the case that I want to mask out threads and just have 1 do something. Take a look at this block: 
https://github.com/llnl/RAJAPerf/blob/develop/src/apps/MASS3DEA-Cuda.cpp#L76-L88 

This comes up in my application all the time. We have been able to do it via `RAJA::RangeSegment(0,1)` but a separate construct would work too. Happy to get a discussion going, maybe separate loop construct would be clear.

artv3: maybe `masked_loop` could work? points for creative names if anyone wants to jump in

MrBurmark: The less to maintain the better, with the range PR you could just have the range be `range(1)` which is pretty short.

artv3: Whats holding me back is that range(1) is kind of a trick, something more explicit would be nice. @tomstitt  what do you think?

trws: The `range(1)` on a worksharing loop would give you the equivalent of `single`, one thread runs it but all other threads wait until it's done.  If that's what this is then that would be a reasonable way to implement it or have people do it, but if you want it _not_ to wait, then I don't think that would get you there, at least not without an appropriate policy or something to request a non-waiting behavior.  It's also only sort-of a loop right? Doing `masked` might be reasonable, have it default to a mask of `0` but allow for an argument that would let you either say how many or which threads should execute it?

tomstitt: I like having a wrapper around `range(1)`, we do want something that we can `range(1)` for any (combination) of threads in `x`, `y`, `z`, not sure if that helps or hurts ideas. The main use cases I can think of are "once across x,y,z" and "once in z" where we nest some x & y work under that, including a "once across x,y"

artv3: I like the wrapper too, in reading the code it makes it easier to identify what is going on.

artv3:
@llnl/raja-core , building on the comments and suggestions, the latest push explores the following function to be used in RAJA::launch : 

```
 165            // __lead_loop_start                                                                                                                  
 166            // Use a single logical thread per team for shared initialization.                                                                    
 167            RAJA::lead<threads_x>(ctx, [&] {                                                                                                      
 168               s_A[0] = r;                                                                                                                        
 169            });                                                                                                                                   
 170            // __lead_loop_end  
 ```

What do you think @tomstitt and @llnl/raja-core ?

**Comment by artv3** [comment]
@llnl/raja-core given that mask is used within RAJA::kernel I thought I would rename the function ``RAJA::lead``, let me know what y'all think.

**Comment by artv3** [comment]
Quick poll @llnl/raja-core , do we want to call it RAJA::mask (🚀 ) or RAJA::lead (🎉 ) ? Only 1 vote per person please. Please respond by Thursday June 25th.

**Comment by artv3** [comment]
> Quick poll @llnl/raja-core , do we want to call it RAJA::mask (🚀 ) or RAJA::lead (🎉 ) ? Only 1 vote per person please. Please respond by Thursday June 25th.

RAJA::mask it is! Will update shortly...

**Comment by adayton1** [comment]
Could you update the title of this MR?

File: docs/sphinx/user_guide/feature/iteration_spaces.rst [hunk 2]
```diff
@@ -56,6 +56,11 @@ execute as a unit for a kernel. RAJA provides the following Segment types:
    * ``RAJA::TypedRangeStrideSegment`` represents a (non-unit) stride range
    * ``RAJA::TypedListSegment`` represents an arbitrary set of indices
 
+RAJA also provides a convenience helper ``RAJA::mask<Policy>(ctx, body)`` for
+launch kernels when one logical thread should execute setup work. It is mainly
```
adayton1: launch -> launching? Or is this referring to RAJA::launch?

include/RAJA/pattern/launch/launch_core.hpp
```diff
@@ -573,6 +601,13 @@ RAJA_HOST_DEVICE RAJA_INLINE void loop_icount(CONTEXT const& ctx,
       ctx, segment0, segment1, segment2, body);
 }
 
+RAJA_SUPPRESS_HD_WARN
+template<typename POLICY_LIST, typename CONTEXT, typename BODY>
+RAJA_HOST_DEVICE RAJA_INLINE void mask(CONTEXT const& ctx, BODY const& body)
```
adayton1, line 604: Does this method need to be implemented both inside and outside the expt namespace?

docs/sphinx/user_guide/feature/iteration_spaces.rst
artv3: yes, RAJA::launch , I'll add the namespace

docs/sphinx/user_guide/tutorial/launch_basic.rst
```diff
@@ -98,3 +98,12 @@ equivalent of the kernel body using the policy shown above is:
    :start-after: // _device_loop_start
    :end-before: // _device_loop_end
    :language: C++
+
+When only one logical thread should execute a piece of work inside a launch
```
artv3:
```suggestion
When only one logical thread should execute a piece of work inside a RAJA::launch
```

docs/sphinx/user_guide/feature/iteration_spaces.rst
artv3:
```suggestion
RAJA::launch kernels when one logical thread should execute setup work. It is mainly
```

docs/sphinx/user_guide/tutorial/launch_basic.rst
```diff
@@ -98,3 +98,12 @@ equivalent of the kernel body using the policy shown above is:
    :start-after: // _device_loop_start
    :end-before: // _device_loop_end
    :language: C++
+
+When only one logical thread should execute a piece of work inside a RAJA::launch
+loop, use ``RAJA::mask<threads_x>(ctx, ...)``. This keeps the intent explicit
```
artv3:
```suggestion
kernel execution space, use ``RAJA::mask<threads_x>(ctx, ...)``. This keeps the intent explicit
```

include/RAJA/pattern/launch/launch_core.hpp
artv3: No, I should actually remove the experimental namespace in launch, those are for methods that I was playing around with but will not keep.

adayton1 (approved):

artv3:
@llnl/raja-core , any additional comments? or are we ready rock and or roll?

rhornung67 (approved):

include/RAJA/pattern/launch/launch_core.hpp
```diff
@@ -476,6 +488,15 @@ using loop_policy = typename POLICY_LIST::device_policy_t;
...
 #endif
 
+template<typename POLICY_LIST>
+#if defined(RAJA_GPU_DEVICE_COMPILE_PASS_ACTIVE)
+using mask_policy = typename POLICY_LIST::device_policy_t;
+#elif defined(RAJA_ENABLE_OPENMP)
+using mask_policy = RAJA::omp_thread;
+#else
+using mask_policy = typename POLICY_LIST::host_policy_t;
+#endif
```
MrBurmark: This always uses omp on the host if its enabled?

include/RAJA/policy/sycl/launch.hpp
```diff
@@ -339,6 +339,64 @@ using sycl_flatten_group_local_201_loop =
 using sycl_flatten_group_local_210_loop =
     sycl_flatten_group_local_loop<2, 1, 0>;
 
+template<int DIM0, int DIM1>
+struct MaskExecute<sycl_flatten_group_local_direct<DIM0, DIM1>>
```
MrBurmark: Could this use a pack of ints like the cuda/hip impl?

test/functional/launch/run-time-switch/tests/test-launch-Mask.hpp
```diff
@@ -0,0 +1,87 @@
...
+      RAJA::LaunchParams(RAJA::Teams(num_teams), RAJA::Threads(threads_per_team)),
+      [=] RAJA_HOST_DEVICE(RAJA::LaunchContext ctx) {
+        RAJA::loop<TEAM_POLICY>(ctx, RAJA::RangeSegment(0, num_teams), [&](int team) {
+          RAJA::mask<THREAD_POLICY>(ctx, [&] {
+            working_array[team] = team + 1;
+          });
+        });
+      });
+
+  working_res.memcpy(check_array, working_array, sizeof(int) * num_teams);
```
MrBurmark, line 51: what about wait?

MrBurmark (approved):

artv3:
@adrienbernede my PR keeps getting stuck in `gitlab-ciExpected — Waiting for status to be reported` could I be missing an account setup?

adrienbernede:
@lc-hubcast approve

adrienbernede (commented):
@lc-hubcast approve

adrienbernede:
@artv3 it looks like we are missing something for your PR to be synced. Have you linked your GitHub account on IDM ?

rhornung67:
@llnl/raja-core I just updated this PR with latest develop. Let's merge it next.

PR #2012, opened 2026-04-14
CPP20Hackathon: Add more concepts and use them in pattern/forall

MrBurmark:
I tried to conceptify the remaining types in the pattern forall interface.

I notice that the `RAJA::Resource` concept is close to overlapping with `camp::resources::Resource`.
The `RAJA::Resource` concept seems to be more of a typed resource concept and does not seem to include `camp::resources::Resource`. We may want to consider renaming for clarity.

Some of the concepts that I added, like `PotentiallyInvokable` or `NonResource`, are vague but may still be useful to document the types that we are intended to be passed through some functions.

### DIFF
--- a/include/RAJA/pattern/forall-concepts.hpp
+++ b/include/RAJA/pattern/forall-concepts.hpp
@@ -42,6 +42,71 @@ template<typename T>
 concept ForallParameterPack =
     RAJA::expt::type_traits::is_ForallParamPack<camp::decay<T>>::value;
 
+
+
+// Concept for strongly typed index types.
+template < typename T >
+concept StrongIndexType = !std::integral<T> && requires (T val)
+    {
+      { stripIndexType(val) } -> std::integral;
+    };
+
+// Concept index types or strongly typed index types.
+// TODO: improve the name
+template < typename T >
+concept IndexTypeConcept = std::integral<T> || StrongIndexType<T>;
+
+// Concept for types that may be callable.
+// Use std::invokable when the full set of arguments is known.
+template < typename B >
+concept PotentialInvokable =
+    std::is_class_v<std::remove_cvref_t<B>> ||
+    std::is_union_v<std::remove_cvref_t<B>> ||
+    std::is_function_v<std::remove_cvref_t<B>> ||
+    ( std::is_pointer_v<std::remove_cvref_t<B>> &&
+      std::is_function_v<std::remove_cv_t<std::remove_pointer_t<std::remove_cvref_t<B>>>> );
+
+consteval bool testPotentialInvokable() noexcept
+{
+  using I = int;
+  using F = int();
+  struct S{};
+  union U{};
+  using L = decltype([]{});
+  static_assert(!PotentialInvokable<I>);
+  static_assert(!PotentialInvokable<I&>);
+  static_assert(!PotentialInvokable<I*>);
+  static_assert(!PotentialInvokable<I[]>);
+  static_assert(PotentialInvokable<F>);
+  static_assert(PotentialInvokable<F&>);
+  static_assert(PotentialInvokable<F*>);
+  static_assert(PotentialInvokable<F*&>);
+  static_assert(!PotentialInvokable<F**>);
+  static_assert(!PotentialInvokable<F*[]>);
+  static_assert(PotentialInvokable<S>);
+  static_assert(PotentialInvokable<S&>);
+  static_assert(!PotentialInvokable<S*>);
+  static_assert(!PotentialInvokable<S[]>);
+  static_assert(PotentialInvokable<U>);
+  static_assert(PotentialInvokable<U&>);
+  static_assert(!PotentialInvokable<U*>);
+  static_assert(!PotentialInvokable<U[]>);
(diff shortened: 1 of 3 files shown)


include/RAJA/pattern/forall-concepts.hpp
```diff
@@ -42,6 +42,71 @@ template<typename T>
...
+
+consteval bool testPotentialInvokable() noexcept
+{
+  using I = int;
+  using F = int();
+  struct S{};
+  union U{};
+  using L = decltype([]{});
+  static_assert(!PotentialInvokable<I>);
+  static_assert(!PotentialInvokable<I&>);
```
johnbowen42, line 77: Could you make a variadic template that tries to substitute the pack into std::is_invocable?

include/RAJA/pattern/forall.hpp
```diff
@@ -94,16 +94,16 @@ namespace RAJA
 namespace detail
 {
 /// Adapter to replace specific implementations for the icount variants
-template<typename Range, typename Body, typename IndexT>
+template<Range Container, PotentialInvokable Body, IndexTypeConcept IndexT>
```
johnbowen42, line 97: We might want to review what concretely we care about in term of `Body`.  I think we might something conceptually much narrower than Invocable

include/RAJA/pattern/forall.hpp
```diff
@@ -557,19 +557,19 @@ requires(!IndexSetPolicy<ExecutionPolicy> &&
 
 template<ForallPolicy ExecutionPolicy,
          Range Container,
-         typename LoopBody,
-         typename Res = typename resources::get_resource<ExecutionPolicy>::type>
+         PotentialInvokable Body,
```
johnbowen42, line 560: Maybe we want to create a `concept` for lambdas for external, user facing function templates, and then another `Privatizer` based concept for internal calls

include/RAJA/pattern/forall.hpp
```diff
@@ -94,16 +94,16 @@ namespace RAJA
...
-template<typename Range, typename Body, typename IndexT>
+template<Range Container, PotentialInvokable Body, IndexTypeConcept IndexT>
 struct icount_adapter
 {
   using index_type = typename std::decay<IndexT>::type;
   typename std::decay<Body>::type body;
-  using container_type = typename std::decay<Range>::type;
+  using container_type = typename std::decay<Container>::type;
   typename container_type::iterator begin_it;
   Index_type icount;
```
artv3, line 104:
```suggestion
  index_type icount;
```
outside of this PR, but line 100 defines `  using index_type = typename std::decay<IndexT>::type;` are we using the right one?

include/RAJA/pattern/forall.hpp
```diff
@@ -94,16 +94,16 @@ namespace RAJA
...
 {
 /// Adapter to replace specific implementations for the icount variants
-template<typename Range, typename Body, typename IndexT>
+template<Range Container, PotentialInvokable Body, IndexTypeConcept IndexT>
 struct icount_adapter
 {
   using index_type = typename std::decay<IndexT>::type;
   typename std::decay<Body>::type body;
-  using container_type = typename std::decay<Range>::type;
+  using container_type = typename std::decay<Container>::type;
```
artv3, line 102:
```suggestion
using container_type = std::decay_t<Container>;
```
More compact

include/RAJA/pattern/forall.hpp
```diff
@@ -94,16 +94,16 @@ namespace RAJA
 namespace detail
 {
 /// Adapter to replace specific implementations for the icount variants
-template<typename Range, typename Body, typename IndexT>
+template<Range Container, PotentialInvokable Body, IndexTypeConcept IndexT>
 struct icount_adapter
 {
   using index_type = typename std::decay<IndexT>::type;
```
artv3, line 100:
```suggestion
using index_type = std::decay_t<IndexT>;
```

include/RAJA/pattern/forall-concepts.hpp
```diff
@@ -42,6 +42,71 @@ template<typename T>
...
+// TODO: improve the name
+template < typename T >
+concept IndexTypeConcept = std::integral<T> || StrongIndexType<T>;
+
+// Concept for types that may be callable.
+// Use std::invokable when the full set of arguments is known.
+template < typename B >
+concept PotentialInvokable =
+    std::is_class_v<std::remove_cvref_t<B>> ||
+    std::is_union_v<std::remove_cvref_t<B>> ||
```
artv3, line 64: do we have callable unions in our applications?

include/RAJA/pattern/forall.hpp
rhornung67: @artv3 I think this sort of thing is now your favorite type of PR review comment 😆

include/RAJA/pattern/forall.hpp
rhornung67: I agree with that. Also, it seems like the `PotentialInvokable` name is somewhat confusing (vs. Invokable which may be a requirement we want to enforce). If the `Body` arg is NOT invokable, what does that mean?

PR #2017, opened 2026-04-16
Support forall params with icount and update clang-format in build scripts

MrBurmark:
# Summary

Support forall params with `forall_Icount` by adding support to pass an arbitrary number of extra parameters through the `icount_adapter`.
Also update the version of clang-format in the build scripts.

- This PR is a bugfix
  - Fixes something that obviously noone is using.

### DIFF
--- a/include/RAJA/pattern/forall.hpp
+++ b/include/RAJA/pattern/forall.hpp
@@ -119,7 +119,8 @@ struct icount_adapter
   template<typename T, typename... Params>
   RAJA_HOST_DEVICE void operator()(T const& i, Params&&... params) const
   {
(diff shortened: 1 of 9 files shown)


adayton1 (approved):

artv3 (approved):
nitpicking: the tile should also include "updates to builds scripts" I didn't expect all the changes. Or have the update to the scripts be a different PR.

MrBurmark:
> nitpicking: the tile should also include "updates to builds scripts" I didn't expect all the changes. Or have the update to the scripts be a different PR.

It originally was just the `forall_Icount` stuff, but then it failed the style test and I realized that our scripts were out of date.

rhornung67 (approved):

PR #2018, opened 2026-04-16
Support nonarithmetic types, generic comparison operators, and non-pointer iterators in gpu sort

### LINKED ISSUE #1810: Support non-arithmetic types in GPU sort
**Describe the bug**

A clear and concise description of what the bug is:

RAJA does not currently support non-arithmetic types in sort with cuda and hip.

**To Reproduce**

Try to sort a non-arithmetic type with the cuda or hip backends.

**Expected behavior**

A clear and concise description of what you expected to happen:

 - Compiler & version: [e.g. GCC 4.9.3]: nvcc 11.6+ and rocm5+ have cub and rocprim that support a comparison sort like mergesort
 - CUDA version (if applicable):

**Additional context**

CARE already supports this.
https://github.com/LLNL/CARE/blob/111d94b74a959d251c28c69028e4230ed67fa394/src/care/KeyValueSorter_impl.h#L158

MrBurmark:
# Summary

Adds support non-arithmetic types, generic comparison operators, and non-pointer iterators in the Cuda and Hip `sort` and `sort_pairs` backends.
This uses the cub and rocprim merge sort implementations that were previously unavailable.

- This PR is a bugfix
- It does the following:
  - Fixes #1810

### DIFF
--- a/docs/sphinx/user_guide/feature/sort.rst
+++ b/docs/sphinx/user_guide/feature/sort.rst
@@ -30,13 +30,17 @@ A few important notes:
 
 Also:
 
-.. note:: For sorts using the CUDA or HIP back-end, RAJA implementation uses
-          the NVIDIA CUB library or AMD rocPRIM library, respectively.
-          Typically, the CMake variable ``CUB_DIR`` or ``ROCPRIM_DIR`` will
-          be automatically set to the location of the CUB or rocPRIM library
-          for the CUDA or rocPRIM installation specified when either back-end
-          is enabled. More details for configuring the CUB or rocPRIM library
-          for a RAJA build can be found :ref:`getting_started_depend-label`.
+.. note:: * For sorts using the CUDA or HIP back-end, RAJA implementation uses
+            the NVIDIA CUB library or AMD rocPRIM library, respectively.
+            Typically, the CMake variable ``CUB_DIR`` or ``ROCPRIM_DIR`` will
+            be automatically set to the location of the CUB or rocPRIM library
+            for the CUDA or rocPRIM installation specified when either back-end
+            is enabled. More details for configuring the CUB or rocPRIM library
+            for a RAJA build can be found :ref:`getting_started_depend-label`.
+          * The RAJA CUDA and HIP sort back-ends support user provided operators,
+            but special case the RAJA operators 'less' and 'greater' when used
+            with raw pointers to arithmetic types for performance as a more
+            performant radix sort implementation is available in those cases.
 
 Please see the following tutorial sections for detailed examples that use
 RAJA scan operations:
--- a/docs/sphinx/user_guide/feature/sort.rst
+++ b/docs/sphinx/user_guide/feature/sort.rst
@@ -30,7 +30,7 @@ A few important notes:
 
 Also:
 
-.. note:: * For sorts using the CUDA or HIP back-end, RAJA implementation uses
+.. note:: * For sorts using the CUDA or HIP back-end, RAJA uses
             the NVIDIA CUB library or AMD rocPRIM library, respectively.
             Typically, the CMake variable ``CUB_DIR`` or ``ROCPRIM_DIR`` will
             be automatically set to the location of the CUB or rocPRIM library
--- a/include/RAJA/policy/cuda/sort.hpp
+++ b/include/RAJA/policy/cuda/sort.hpp
@@ -84,7 +84,8 @@ resources::EventProxy<resources::Cuda> sort(
   void* d_temp_storage      = nullptr;
   size_t temp_storage_bytes = 0;
 
-  auto call_impl = [&, tmp_begin = static_cast<R*>(nullptr)](auto phase) mutable {
+  auto call_impl = [&,
+                    tmp_begin = static_cast<R*>(nullptr)](auto phase) mutable {
     RAJA_UNUSED_VAR(phase);
     RAJA_UNUSED_VAR(tmp_begin);
 
@@ -213,7 +214,8 @@ resources::EventProxy<resources::Cuda> sort_pairs(
   size_t temp_storage_bytes = 0;
 
   auto call_impl = [&, tmp_keys_begin = static_cast<K*>(nullptr),
-                    tmp_vals_begin = static_cast<V*>(nullptr)](auto phase) mutable {
+                    tmp_vals_begin =
+                        static_cast<V*>(nullptr)](auto phase) mutable {
     RAJA_UNUSED_VAR(phase);
     RAJA_UNUSED_VAR(tmp_keys_begin);
     RAJA_UNUSED_VAR(tmp_vals_begin);
@@ -230,8 +232,10 @@ resources::EventProxy<resources::Cuda> sort_pairs(
       // Setup temporary storage for the output arrays
       if constexpr (phase == 0)
       {
-        tmp_keys_begin = cuda::device_mempool_type::getInstance().malloc<K>(len);
-        tmp_vals_begin = cuda::device_mempool_type::getInstance().malloc<V>(len);
+        tmp_keys_begin =
+            cuda::device_mempool_type::getInstance().malloc<K>(len);
+        tmp_vals_begin =
+            cuda::device_mempool_type::getInstance().malloc<V>(len);
       }
 
       // Use DoubleBuffers to reduce temporary memory requirements
@@ -259,7 +263,8 @@ resources::EventProxy<resources::Cuda> sort_pairs(
       if constexpr (phase == 1)
       {
         // copy keys and values back if necessary
-        if (d_keys.Current() == tmp_keys_begin && d_vals.Current() == tmp_vals_begin)
+        if (d_keys.Current() == tmp_keys_begin &&
+            d_vals.Current() == tmp_vals_begin)
         {
           // Copy keys and values back via kernel for performance
           forall_impl(
--- a/include/RAJA/policy/cuda/sort.hpp
+++ b/include/RAJA/policy/cuda/sort.hpp
@@ -268,7 +268,7 @@ resources::EventProxy<resources::Cuda> sort_pairs(
                   IterationMapping, IterationGetter, Concretizer, BLOCKS_PER_SM,
                   true> {},
               TypedRangeSegment<IndexType>(static_cast<IndexType>(0), len),
-              ::RAJA::detail::Copy2Functor {keys_begin, d_keys_out, vals_begin,
+              ::RAJA::detail::TwoCopiesFunctor {keys_begin, d_keys_out, vals_begin,
                                             d_vals_out},
               expt::get_empty_forall_param_pack());
         }
--- a/include/RAJA/policy/cuda/sort.hpp
+++ b/include/RAJA/policy/cuda/sort.hpp
@@ -399,7 +399,7 @@ resources::EventProxy<resources::Cuda> stable_pairs(
 }
 
 /*!
-        \brief stable sort given range of pairs in order of keys
+        \brief unstable sort given range of pairs in order of keys
 */
 template<typename IterationMapping,
          typename IterationGetter,
--- a/include/RAJA/policy/cuda/sort.hpp
+++ b/include/RAJA/policy/cuda/sort.hpp
@@ -75,99 +75,95 @@ resources::EventProxy<resources::Cuda> sort(
 
   using R = RAJA::detail::IterVal<Iter>;
 
-  int len       = std::distance(begin, end);
-  int begin_bit = 0;
-  int end_bit   = sizeof(R) * CHAR_BIT;
+  using IndexType = camp::decay<decltype(std::distance(begin, end))>;
 
-  // Allocate temporary storage for the output array
-  R* d_out = cuda::device_mempool_type::getInstance().malloc<R>(len);
+  const IndexType len     = std::distance(begin, end);
+  constexpr int begin_bit = 0;
+  constexpr int end_bit   = sizeof(R) * CHAR_BIT;
 
-  // use cub double buffer to reduce temporary memory requirements
-  // by allowing cub to write to the begin buffer
-  cub::DoubleBuffer<R> d_keys(begin, d_out);
-
-  // Determine temporary device storage requirements
   void* d_temp_storage      = nullptr;
   size_t temp_storage_bytes = 0;
-  if constexpr (std::is_arithmetic_v<R> && std::is_pointer_v<Iter> &&
-                std::is_same_v<std::decay_t<Compare>, operators::less<R>>)
+
+  auto call_impl = [&, d_out = static_cast<R*>(nullptr)](auto phase) mutable
   {
-    CAMP_CUDA_API_INVOKE_AND_CHECK(::cub::DeviceRadixSort::SortKeys,
-                                   d_temp_storage, temp_storage_bytes, d_keys,
-                                   len, begin_bit, end_bit, stream);
-  }
-  else if constexpr (std::is_arithmetic_v<R> && std::is_pointer_v<Iter> &&
-                     std::is_same_v<std::decay_t<Compare>,
-                                    operators::greater<R>>)
-  {
-    CAMP_CUDA_API_INVOKE_AND_CHECK(::cub::DeviceRadixSort::SortKeysDescending,
-                                   d_temp_storage, temp_storage_bytes, d_keys,
-                                   len, begin_bit, end_bit, stream);
-  }
-  else
-  {
-    if (Stable)
+    RAJA_UNUSED_VAR(phase);
+    RAJA_UNUSED_VAR(d_out);
+
+    if constexpr (std::is_arithmetic_v<R> && std::is_pointer_v<Iter> &&
+                  ( std::is_same_v<std::decay_t<Compare>, operators::less<R>> ||
+                    std::is_same_v<std::decay_t<Compare>, operators::greater<R>> ))
     {
-      CAMP_CUDA_API_INVOKE_AND_CHECK(::cub::DeviceMergeSort::StableSortKeys,
-                                     d_temp_storage, temp_storage_bytes,
-                                     d_keys.Current(), len, comp, stream);
+      // The begin iterator is a pointer in this constexpr conditional
+      // so we can use DoubleBuffer
+
+      // Setup temporary storage for the output array
+      if constexpr (phase == 0) {
+        d_out = cuda::device_mempool_type::getInstance().malloc<R>(len);
+      }
+
+      // Use DoubleBuffers to reduce temporary memory requirements
+      // by allowing cub to write to the begin buffer
+      cub::DoubleBuffer<R> d_keys(begin, d_out);
+
+      if constexpr (std::is_same_v<std::decay_t<Compare>, operators::less<R>>)
+      {
+        CAMP_CUDA_API_INVOKE_AND_CHECK(::cub::DeviceRadixSort::SortKeys,
+                                       d_temp_storage, temp_storage_bytes, d_keys,
+                                       len, begin_bit, end_bit, stream);
+      }
+      else if constexpr (std::is_same_v<std::decay_t<Compare>, operators::greater<R>>)
+      {
+        CAMP_CUDA_API_INVOKE_AND_CHECK(::cub::DeviceRadixSort::SortKeysDescending,
+                                       d_temp_storage, temp_storage_bytes, d_keys,
+                                       len, begin_bit, end_bit, stream);
+      }
+
+      // Tear-down temporary storage for the output array
+      if constexpr (phase == 1)
+      {
+        // Copy result back if necessary
+        if (d_keys.Current() == d_out)
+        {
+          CAMP_CUDA_API_INVOKE_AND_CHECK(cudaMemcpyAsync, begin, d_out,
+                                         len * sizeof(R), cudaMemcpyDefault, stream);
+        }
+
+        // Free temporary output array
+        cuda::device_mempool_type::getInstance().free(d_out);
+      }
     }
     else
     {
-      CAMP_CUDA_API_INVOKE_AND_CHECK(::cub::DeviceMergeSort::SortKeys,
-                                     d_temp_storage, temp_storage_bytes,
-                                     d_keys.Current(), len, comp, stream);
+      if (Stable)
+      {
+        CAMP_CUDA_API_INVOKE_AND_CHECK(::cub::DeviceMergeSort::StableSortKeys,
+                                       d_temp_storage, temp_storage_bytes,
+                                       begin, len, comp, stream);
+      }
+      else
+      {
+        CAMP_CUDA_API_INVOKE_AND_CHECK(::cub::DeviceMergeSort::SortKeys,
+                                       d_temp_storage, temp_storage_bytes,
+                                       begin, len, comp, stream);
+      }
     }
-  }
+  };
+
+  // Determine temporary device storage requirements
+  call_impl(std::integral_constant<int, 0>{});
+
   // Allocate temporary storage
   d_temp_storage =
       cuda::device_mempool_type::getInstance().malloc<unsigned char>(
           temp_storage_bytes);
 
   // Run
-  if constexpr (std::is_arithmetic_v<R> && std::is_pointer_v<Iter> &&
-                std::is_same_v<std::decay_t<Compare>, operators::less<R>>)
-  {
-    CAMP_CUDA_API_INVOKE_AND_CHECK(::cub::DeviceRadixSort::SortKeys,
-                                   d_temp_storage, temp_storage_bytes, d_keys,
-                                   len, begin_bit, end_bit, stream);
-  }
-  else if constexpr (std::is_arithmetic_v<R> && std::is_pointer_v<Iter> &&
-                     std::is_same_v<std::decay_t<Compare>,
-                                    operators::greater<R>>)
-  {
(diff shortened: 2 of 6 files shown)


include/RAJA/policy/cuda/sort.hpp
```diff
@@ -128,88 +89,72 @@ stable(resources::Cuda cuda_res,
...
+                                   d_temp_storage, temp_storage_bytes, d_keys,
+                                   len, begin_bit, end_bit, stream);
+  }
+  else
+  {
+    if (Stable)
+    {
+      CAMP_CUDA_API_INVOKE_AND_CHECK(::cub::DeviceMergeSort::StableSortKeys,
+                                     d_temp_storage, temp_storage_bytes, d_keys.Current(),
+                                     len, comp, stream);
```
MrBurmark: For cub merge sort there is an option to use an in-place or out-of-place sort implementation. I'm not sure which is better or uses fewer resources. Currently when we use the merge sort implementation with cub we allocate but don't use an output array, so it would probably be best to at least not allocate the output array to reduce memory usage.

include/RAJA/policy/cuda/sort.hpp
```diff
@@ -460,198 +318,112 @@ stable_pairs(resources::Cuda cuda_res,
...
-                                                        Concretizer,
-                                                        BLOCKS_PER_SM,
-                                                        Async> p,
-               KeyIter keys_begin,
-               KeyIter keys_end,
-               ValIter vals_begin,
-               operators::greater<RAJA::detail::IterVal<KeyIter>> comp)
+         typename ValIter,
+         typename Compare>
+resources::EventProxy<resources::Cuda> unstable_pairs(
```
artv3, line 412: The documentation above says the function is stable sort but the function is called unstable pairs?

include/RAJA/policy/cuda/sort.hpp
```diff
@@ -460,198 +318,112 @@ stable_pairs(resources::Cuda cuda_res,
...
 {
-  return stable_pairs(cuda_res, p, keys_begin, keys_end, vals_begin, comp);
+  constexpr bool stable = true;
+  return detail::sort_pairs<stable>(cuda_res, p, keys_begin, keys_end,
+                                    vals_begin, comp);
 }
 
 /*!
-        \brief stable sort given range of pairs in descending order of keys
+        \brief stable sort given range of pairs in order of keys
```
artv3:
```suggestion
        \brief unstable sort given range of pairs in order of keys
```
Maybe this is what you meant?

include/RAJA/policy/cuda/sort.hpp
MrBurmark: good catch, the function name is right.

include/RAJA/policy/hip/sort.hpp
```diff
@@ -71,192 +73,151 @@ R* get_current(double_buffer<R>& d_bufs)
...
-  // Determine temporary device storage requirements
-  void* d_temp_storage      = nullptr;
-  size_t temp_storage_bytes = 0;
+  else
+  {
 #if defined(__HIPCC__)
-  CAMP_HIP_API_INVOKE_AND_CHECK(::rocprim::radix_sort_keys_desc, d_temp_storage,
-                                temp_storage_bytes, d_keys, len, begin_bit,
-                                end_bit, stream);
+    CAMP_HIP_API_INVOKE_AND_CHECK(::rocprim::merge_sort, d_temp_storage,
```
artv3: For the `__CUDACC__`  case we have a clear distinction of API calls for the stable vs not stable case. It would be nice to know what we are a getting when we call `rocprim::merge_sort` as it looks like the same API is called for the stable and unstable case.

include/RAJA/policy/hip/sort.hpp
```diff
@@ -272,215 +233,171 @@ stable(resources::Hip hip_res,
...
+                                    operators::greater<K>>)
+  {
+#if defined(__HIPCC__)
+    CAMP_HIP_API_INVOKE_AND_CHECK(::rocprim::radix_sort_pairs_desc,
+                                  d_temp_storage, temp_storage_bytes, d_keys,
+                                  d_vals, len, begin_bit, end_bit, stream);
+#elif defined(__CUDACC__)
+    CAMP_CUDA_API_INVOKE_AND_CHECK(::cub::DeviceRadixSort::SortPairsDescending,
+                                   d_temp_storage, temp_storage_bytes, d_keys,
+                                   d_vals, len, begin_bit, end_bit, stream);
```
artv3: Same question here, a comment or documentation on what we are getting when using the HIP backend, below we have clear support for stable vs not stable.

include/RAJA/policy/hip/sort.hpp
```diff
@@ -71,192 +73,151 @@ R* get_current(double_buffer<R>& d_bufs)
...
+    }
 #endif
+  }
   // Allocate temporary storage
   d_temp_storage =
       hip::device_mempool_type::getInstance().malloc<unsigned char>(
           temp_storage_bytes);
 
   // Run
+  if constexpr (std::is_arithmetic_v<R> && std::is_pointer_v<Iter> &&
```
artv3: These branching if statements look very similar to the ones above in which you allocating temp storage. Can you consider an implementation in which you choose the backend once then use the implementation in both places?

include/RAJA/policy/hip/sort.hpp
```diff
@@ -272,215 +233,171 @@ stable(resources::Hip hip_res,
...
+    }
 #endif
+  }
   // Allocate temporary storage
   d_temp_storage =
       hip::device_mempool_type::getInstance().malloc<unsigned char>(
           temp_storage_bytes);
 
   // Run
+  if constexpr (std::is_arithmetic_v<K> && std::is_pointer_v<KeyIter> &&
```
artv3: Same comment here about the similar `if constexpr` as when you are allocating the memory.

include/RAJA/policy/cuda/sort.hpp
```diff
@@ -128,88 +89,72 @@ stable(resources::Cuda cuda_res,
...
-  // Run
-  CAMP_CUDA_API_INVOKE_AND_CHECK(::cub::DeviceRadixSort::SortKeys,
-                                 d_temp_storage, temp_storage_bytes, d_keys,
-                                 len, begin_bit, end_bit, stream);
-  // Free temporary storage
-  cuda::device_mempool_type::getInstance().free(d_temp_storage);
-
-  if (d_keys.Current() == d_out)
+  if constexpr (std::is_arithmetic_v<R> && std::is_pointer_v<Iter> &&
+                std::is_same_v<std::decay_t<Compare>, operators::less<R>>)
```
artv3: This is RAJA specific right `operators::less<R>`? so an external/generic comparison won't give you the `DeviceRadixSort` call? It looks like it falls back to DeviceMergeSort::StableSortKeys , this might be worth documenting so users know what to expect

include/RAJA/policy/hip/sort.hpp
```diff
@@ -71,192 +73,151 @@ R* get_current(double_buffer<R>& d_bufs)
...
-       operators::less<RAJA::detail::IterVal<Iter>>)
-{
   hipStream_t stream = hip_res.get_stream();
 
   using R = RAJA::detail::IterVal<Iter>;
 
-  int len       = std::distance(begin, end);
-  int begin_bit = 0;
-  int end_bit   = sizeof(R) * CHAR_BIT;
+  const int len           = std::distance(begin, end);
```
artv3: `begin` and `end` are template types, if they are not int, it might make more sense to have the left hand side be `auto`

include/RAJA/policy/cuda/sort.hpp
```diff
@@ -460,198 +318,112 @@ stable_pairs(resources::Cuda cuda_res,
...
-      "sort_pairs<cuda_exec> is only implemented for arithmetic types");
-  static_assert(
-      concepts::any_of<camp::is_same<Compare, operators::less<K>>,
-                       camp::is_same<Compare, operators::greater<K>>>::value,
-      "sort_pairs<cuda_exec> is only implemented for RAJA::operators::less or "
-      "RAJA::operators::greater");
-
-  return resources::EventProxy<resources::Cuda>(cuda_res);
+  constexpr bool stable = true;
+  return detail::sort<!stable>(cuda_res, p, begin, end, comp);
```
artv3, line 379: minor nitpicking.... wouldn't be more direct if we just said `  return detail::sort<false>(cuda_res, p, begin, end, comp);` saves my brain an evaluation step...

include/RAJA/policy/cuda/sort.hpp
MrBurmark: fixed.

include/RAJA/policy/hip/sort.hpp
MrBurmark: This is done because `rocprim::merge_sort` is stable per the rocprim documentation, and cub has separate functions for stable and unstable.

MrBurmark: In all cases if there is no split between stable and unstable then there was only one stable sort implementation available, so you always get a stable sort even if you don't need it to be stable.

include/RAJA/policy/hip/sort.hpp
MrBurmark: There are multiple copies in some cases because depending on the exact path you may need or don't need extra memory. I chose to duplicate the allocation along the paths where they were needed to simplify the conditional logic. I think de-duplicating the allocations is not worth it as it would force us to have to do a lot of decisions up front that have to match how things are coded later.

include/RAJA/policy/cuda/sort.hpp
MrBurmark: Yep, I think that the current documentation states that we only support our operator less and greater with gpu backends. I'll have to update it so it says that we now support generic comparison operators and note that you get a faster implementation if you use our operators.

include/RAJA/policy/hip/sort.hpp
MrBurmark: fixed

include/RAJA/policy/cuda/sort.hpp
MrBurmark: This was written this way for readability. Writing false here makes it unclear what is false. On the other hand having a named constant lets you read this directly, `sort<stable>` is stable sort and `sort<!stable>` is unstable sort.

include/RAJA/policy/hip/sort.hpp
MrBurmark: I was thinking about what I changed the code to after you made this comment, the duplication that you mentioned has been eliminated using a lambda. However there is now some duplication in dealing with temporary output buffers, see my comment above.

include/RAJA/policy/cuda/sort.hpp
MrBurmark: fixed.

include/RAJA/policy/hip/sort.hpp
```diff
@@ -71,623 +73,453 @@ R* get_current(double_buffer<R>& d_bufs)
...
+      if constexpr (phase == 1)
+      {
+        // Copy result back via kernel as begin may not be a pointer
+        forall_impl(
+            hip_res,
+            ::RAJA::policy::hip::hip_exec<IterationMapping, IterationGetter,
+                                          Concretizer, true> {},
+            TypedRangeSegment<IndexType>(static_cast<IndexType>(0), len),
+            ::RAJA::detail::Copy1Functor {begin, d_out},
+            expt::get_empty_forall_param_pack());
```
artv3, line 195: But if this is a pointer, can't we do something like: 
```suggestion
    if constexpr (std::is_pointer_v<Iter>)
    {
      CAMP_HIP_API_INVOKE_AND_CHECK(hipMemcpyAsync, begin, d_out,
                                    len * sizeof(R), hipMemcpyDefault, stream);
    }
    else
    {
      forall_impl(
          hip_res,
          ::RAJA::policy::hip::hip_exec<IterationMapping, IterationGetter,
                                        Concretizer, true> {},
          TypedRangeSegment<IndexType>(static_cast<IndexType>(0), len),
          ::RAJA::detail::Copy1Functor {begin, d_out},
          expt::get_empty_forall_param_pack());
    }
```

include/RAJA/policy/hip/sort.hpp
```diff
@@ -71,623 +73,453 @@ R* get_current(double_buffer<R>& d_bufs)
...
+        // Copy keys and values back via kernel as iterators may not be a
+        // pointer
+        forall_impl(
+            hip_res,
+            ::RAJA::policy::hip::hip_exec<IterationMapping, IterationGetter,
+                                          Concretizer, true> {},
+            TypedRangeSegment<IndexType>(static_cast<IndexType>(0), len),
+            ::RAJA::detail::Copy2Functor {keys_begin, d_keys_out, vals_begin,
+                                          d_vals_out},
+            expt::get_empty_forall_param_pack());
```
artv3, line 390: Similar comment for when it is a pointer -- it does increase the lines of code but worth checking out?

```suggestion
    if constexpr (std::is_pointer_v<KeyIter> && std::is_pointer_v<ValIter>)
    {
      CAMP_HIP_API_INVOKE_AND_CHECK(hipMemcpyAsync, keys_begin, d_keys_out,
                                    len * sizeof(K), hipMemcpyDefault, stream);
      CAMP_HIP_API_INVOKE_AND_CHECK(hipMemcpyAsync, vals_begin, d_vals_out,
                                    len * sizeof(V), hipMemcpyDefault, stream);
    }
    else if constexpr (std::is_pointer_v<KeyIter>)
    {
      CAMP_HIP_API_INVOKE_AND_CHECK(hipMemcpyAsync, keys_begin, d_keys_out,
                                    len * sizeof(K), hipMemcpyDefault, stream);

      forall_impl(
          hip_res,
          ::RAJA::policy::hip::hip_exec<IterationMapping, IterationGetter,
                                        Concretizer, true> {},
          TypedRangeSegment<IndexType>(static_cast<IndexType>(0), len),
          ::RAJA::detail::Copy1Functor {vals_begin, d_vals_out},
          expt::get_empty_forall_param_pack());
    }
    else if constexpr (std::is_pointer_v<ValIter>)
    {
      CAMP_HIP_API_INVOKE_AND_CHECK(hipMemcpyAsync, vals_begin, d_vals_out,
                                    len * sizeof(V), hipMemcpyDefault, stream);

      forall_impl(
          hip_res,
          ::RAJA::policy::hip::hip_exec<IterationMapping, IterationGetter,
                                        Concretizer, true> {},
          TypedRangeSegment<IndexType>(static_cast<IndexType>(0), len),
          ::RAJA::detail::Copy1Functor {keys_begin, d_keys_out},
          expt::get_empty_forall_param_pack());
    }
    else
    {
      forall_impl(
          hip_res,
          ::RAJA::policy::hip::hip_exec<IterationMapping, IterationGetter,
                                        Concretizer, true> {},
          TypedRangeSegment<IndexType>(static_cast<IndexType>(0), len),
          ::RAJA::detail::Copy2Functor {keys_begin, d_keys_out, vals_begin,
                                        d_vals_out},
          expt::get_empty_forall_param_pack());
    }
```

include/RAJA/policy/cuda/sort.hpp
artv3: that makes sense, resolved

include/RAJA/policy/hip/sort.hpp
artv3: We probably have too many options here, maybe considering if both or neither are pointers? maybe others can weight in here...

MrBurmark: I chose to use a kernel more often because generally a kernel performs better, and in some of those cases we could use a single kernel instead of 2 memcpys which reduces launch overhead. The other hand is that memcpy generally has a bit less launch overhead itself and it can run in parallel with kernels in different streams.

artv3: ah good to know, I'll make this as resolved

artv3 (approved):
Looks great @MrBurmark ! Thanks for addressing my comments.

docs/sphinx/user_guide/feature/sort.rst
```diff
@@ -30,13 +30,17 @@ A few important notes:
...
 Also:
 
-.. note:: For sorts using the CUDA or HIP back-end, RAJA implementation uses
-          the NVIDIA CUB library or AMD rocPRIM library, respectively.
-          Typically, the CMake variable ``CUB_DIR`` or ``ROCPRIM_DIR`` will
-          be automatically set to the location of the CUB or rocPRIM library
-          for the CUDA or rocPRIM installation specified when either back-end
-          is enabled. More details for configuring the CUB or rocPRIM library
-          for a RAJA build can be found :ref:`getting_started_depend-label`.
+.. note:: * For sorts using the CUDA or HIP back-end, RAJA implementation uses
```
adayton1: the RAJA implementation

MrBurmark:
```suggestion
.. note:: * For sorts using the CUDA or HIP back-end, RAJA uses
```

include/RAJA/policy/cuda/sort.hpp
```diff
@@ -43,615 +44,395 @@ namespace impl
...
-      "stable_sort<cuda_exec> is only implemented for arithmetic types");
-  static_assert(concepts::any_of<
-                    camp::is_same<Compare, operators::less<iterval>>,
-                    camp::is_same<Compare, operators::greater<iterval>>>::value,
-                "stable_sort<cuda_exec> is only implemented for "
-                "RAJA::operators::less or RAJA::operators::greater");
-
-  return resources::EventProxy<resources::Cuda>(cuda_res);
-}
+  RAJA_UNUSED_VAR(Stable);
```
adayton1: Is this necessary? Isn't Stable a template parameter?

include/RAJA/policy/cuda/sort.hpp
```diff
@@ -43,615 +44,395 @@ namespace impl
...
+                                         stream);
+        }
+
+        // Free temporary output array
+        cuda::device_mempool_type::getInstance().free(d_out);
+      }
+    }
+    else
+    {
+      if (Stable)
```
adayton1: Can this be `if constexpr` since `Stable` is a template parameter?

include/RAJA/policy/cuda/sort.hpp
MrBurmark: I have seen complaints about unused template args before so this was preemptive. I really put this here to match the hip sort file where its not used based on the ifdef for AMD gpus, and to keep the diff between the hip and cuda backends to a minimum. Maybe there is a better place to put it to make this more clear?

include/RAJA/policy/cuda/sort.hpp
```diff
@@ -43,615 +44,395 @@ namespace impl
...
+      {
+        CAMP_CUDA_API_INVOKE_AND_CHECK(
+            ::cub::DeviceMergeSort::SortPairs, d_temp_storage,
+            temp_storage_bytes, keys_begin, vals_begin, len, comp, stream);
+      }
+    }
+  };
+
+  // Determine temporary device storage requirements
+  call_impl(std::integral_constant<int, 0> {});
```
adayton1, line 316: I'm not sure this lambda actually adds much. You already have to have separate phases, so why not just spell the code out?

include/RAJA/policy/cuda/sort.hpp
```diff
@@ -43,615 +44,395 @@ namespace impl
...
-      "sort_pairs<cuda_exec> is only implemented for arithmetic types");
-  static_assert(
-      concepts::any_of<camp::is_same<Compare, operators::less<K>>,
-                       camp::is_same<Compare, operators::greater<K>>>::value,
-      "sort_pairs<cuda_exec> is only implemented for RAJA::operators::less or "
-      "RAJA::operators::greater");
-
-  return resources::EventProxy<resources::Cuda>(cuda_res);
+  constexpr bool stable = true;
+  return detail::sort<!stable>(cuda_res, p, begin, end, comp);
```
adayton1: In a way, this is a little misleading. You might still get a stable sort even though you tell it not to.

include/RAJA/util/types.hpp
```diff
@@ -1053,6 +1053,55 @@ struct ScopedAssignment
...
+Copy1Functor(DestIter, SrcIter) -> Copy1Functor<DestIter, SrcIter>;
+
+/*!
+ * \brief Functor that copies src1 to dst1 and src2 to dst2.
+ */
+template<typename DestIter1,
+         typename SrcIter1,
+         typename DestIter2,
+         typename SrcIter2>
+struct Copy2Functor
```
adayton1: Maybe CopyPairFunctor?

include/RAJA/policy/cuda/sort.hpp
MrBurmark: Do you find the usage of stable here misleading, or the fact that the implementation is often stable even if you don't ask for that misleading?

include/RAJA/policy/cuda/sort.hpp
MrBurmark: Ya this is awkward and the hip case has more duplication. I decided that since the phase specific code was shorter overall and had to be somewhere either way so de-duplicating the non-phase specific code was worth it.

include/RAJA/policy/cuda/sort.hpp
adayton1: Both. Maybe the parameter should be named StableGuaranteed or StableAlways or StableRequired.

adayton1: Or RequireStable.

include/RAJA/util/types.hpp
MrBurmark: Maybe something like CopyTwoArraysFunctor would be even more descriptive.
I also thought about having a class that took a param pack of pairs of src,dst iterators so this would only be 1 class, but writing two classes ultimately seemed simpler. Well host device lambdas would be even simpler, but I don't think you can declare one in a lambda with cuda.

include/RAJA/policy/cuda/sort.hpp
artv3: I agree with @adayton1 on this one, and second the parameter recommendation

MrBurmark: I went with RequireStable.

include/RAJA/util/types.hpp
MrBurmark: I changed them to `OneCopyFunctor` and `TwoCopiesFunctor` is that more clear?

rhornung67 (approved):

include/RAJA/util/types.hpp
MrBurmark: Would `TwoCopyFunctor` be better?

MrBurmark: Went with `CopyFunctorOneRange` and `CopyFunctorTwoRanges`.

include/RAJA/policy/cuda/sort.hpp
```diff
@@ -268,8 +268,8 @@ resources::EventProxy<resources::Cuda> sort_pairs(
                   IterationMapping, IterationGetter, Concretizer, BLOCKS_PER_SM,
                   true> {},
               TypedRangeSegment<IndexType>(static_cast<IndexType>(0), len),
-              ::RAJA::detail::TwoCopiesFunctor {keys_begin, d_keys_out,
-                                                vals_begin, d_vals_out},
+              ::RAJA::detail::CopyFunctorTwoRanges {keys_begin, d_keys_out,
```
rhornung67: Would it be clearer and more consistent to change the args as follows?

`keys/vals_begin` --> `keys/vals_in_begin`
`d_keys/vals_out` --> `d_keys/vals_out_begin`

Similar changes to other places in the code.

include/RAJA/policy/hip/sort.hpp
```diff
@@ -191,7 +191,7 @@ resources::EventProxy<resources::Hip> sort(
             ::RAJA::policy::hip::hip_exec<IterationMapping, IterationGetter,
                                           Concretizer, true> {},
             TypedRangeSegment<IndexType>(static_cast<IndexType>(0), len),
-            ::RAJA::detail::OneCopyFunctor {begin, d_out},
+            ::RAJA::detail::CopyFunctorOneRange {begin, d_out},
```
rhornung67: Similar comment to the one above for two ranges.

adayton1 (approved):

include/RAJA/policy/cuda/sort.hpp
MrBurmark: I was at cppref and they use `first` for input and `d_first` for output, with the `d` standing for destination I suppose. https://en.cppreference.com/cpp/algorithm/copy

MrBurmark: Do we want to go to something like that `keys_first` `d_keys_first`, or `src_keys_first` and `dst_keys_first`, or `keys_in_first` and `keys_out_first`?

rhornung67: I think it's best if the names are clear and consistent. I think *_begin and *_out is inconsistent. I think d_* to mean output is not clear. 

Just suggesting more consideration.

MrBurmark: In this case these are all temporary variables and sometimes the result ends up in them, sometimes not, so I'm renaming them to something like `tmp_keys_begin`.

rhornung67: works for me.  also, some projects use `d_` prefix for class member variables, similar to `m_`

PR #2020, opened 2026-04-28
C++20: bump camp submodule and modify SFINAE to ensure compatibility 

johnbowen42:
This PR does more than just bump the camp submodule.  Changing the submodule required some modernization and changes to ensure compatibility.

### DIFF
--- a/include/RAJA/index/RangeSegment.hpp
+++ b/include/RAJA/index/RangeSegment.hpp
@@ -553,7 +553,7 @@ using common_type_t = typename common_type<Ts...>::type;
  * \brief Function to make a TypedRangeSegment for the interval [begin, end)
  *
  *  \return a newly constructed TypedRangeSegment where the
- *          value_type is equivilent to the common type of
+ *          value_type is equivalent to the common type of
  *          @begin and @end. If there is no common type, then
  *          a compiler error will be produced.
  */
@@ -596,23 +596,23 @@ namespace concepts
 {
 
 template<typename T, typename U>
-struct RangeConstructible
-    : DefineConcept(camp::val<RAJA::detail::common_type_t<T, U>>()) {};
+concept RangeConstructible = std::common_with<T, U>;
 
 template<typename T, typename U, typename V>
-struct RangeStrideConstructible
-    : DefineConcept(camp::val<RAJA::detail::common_type_t<T, U, V>>()) {};
+concept RangeStrideConstructible = RAJA::detail::common_type_t<T, U, V>::value;
 
 }  // namespace concepts
 
 namespace type_traits
 {
 
-DefineTypeTraitFromConcept(is_range_constructible,
-                           RAJA::concepts::RangeConstructible);
+DefineTypeTraitFromConceptTwoTypeParams(is_range_constructible,
+                                        RAJA::concepts::RangeConstructible);
 
-DefineTypeTraitFromConcept(is_range_stride_constructible,
-                           RAJA::concepts::RangeStrideConstructible);
+
+DefineTypeTraitFromConceptThreeTypeParams(
+    is_range_stride_constructible,
+    RAJA::concepts::RangeStrideConstructible);
 
 }  // namespace type_traits
 
--- a/include/RAJA/index/RangeSegment.hpp
+++ b/include/RAJA/index/RangeSegment.hpp
@@ -592,30 +592,6 @@ RAJA_HOST_DEVICE TypedRangeStrideSegment<Common> make_strided_range(
   return {begin, end, stride};
 }
 
-namespace concepts
-{
-
-template<typename T, typename U>
-concept RangeConstructible = std::common_with<T, U>;
-
-template<typename T, typename U, typename V>
-concept RangeStrideConstructible = RAJA::detail::common_type_t<T, U, V>::value;
-
-}  // namespace concepts
-
-namespace type_traits
-{
-
-DefineTypeTraitFromConceptTwoTypeParams(is_range_constructible,
-                                        RAJA::concepts::RangeConstructible);
-
-
-DefineTypeTraitFromConceptThreeTypeParams(
-    is_range_stride_constructible,
-    RAJA::concepts::RangeStrideConstructible);
-
-}  // namespace type_traits
-
 }  // namespace RAJA
 
 namespace std
--- a/include/RAJA/internal/Iterators.hpp
+++ b/include/RAJA/internal/Iterators.hpp
@@ -341,6 +341,20 @@ public:
     return *this;
   }
 
+  RAJA_HOST_DEVICE inline strided_numeric_iterator operator++(int)
+  {
+    strided_numeric_iterator tmp(*this);
+    val += stride;
+    return tmp;
+  }
+
+  RAJA_HOST_DEVICE inline strided_numeric_iterator operator--(int)
+  {
+    strided_numeric_iterator tmp(*this);
+    val -= stride;
+    return tmp;
+  }
+
   RAJA_HOST_DEVICE inline strided_numeric_iterator& operator+=(
       const difference_type& rhs)
   {
@@ -396,6 +410,16 @@ public:
     check_is_subtraction_overflow(val, rhs * stride);
 #endif
     return strided_numeric_iterator(val - rhs * stride, stride);
+  }
+
+  RAJA_HOST_DEVICE friend constexpr strided_numeric_iterator operator+(
+      difference_type lhs,
+      const strided_numeric_iterator& rhs)
+  {
+#if defined(RAJA_ENABLE_ITERATOR_OVERFLOW_DEBUG)
+    check_is_addition_overflow(rhs.val, lhs * rhs.stride);
+#endif
+    return strided_numeric_iterator(rhs.val + lhs * rhs.stride, rhs.stride);
   }
 
   // Specialized comparison to allow normal iteration to work on off-stride
--- a/include/RAJA/pattern/detail/TypeTraits.hpp
+++ b/include/RAJA/pattern/detail/TypeTraits.hpp
@@ -96,6 +96,17 @@ struct tuple_contains_Reducers<camp::tuple<Params...>>
 {};
 
 }  // namespace expt
+
+namespace type_traits
+{
+template<typename U, typename V>
+struct is_same_strip_const
+    : std::is_same<std::remove_const_t<U>, std::remove_const_t<V>>
+{};
+
+template<typename U, typename V>
+static constexpr bool is_same_strip_const_v = is_same_strip_const<U, V>::value;
+}  // namespace type_traits
 }  // namespace RAJA
 
 #endif  //  RAJA_TYPETRAITS_HPP
--- a/include/RAJA/pattern/detail/TypeTraits.hpp
+++ b/include/RAJA/pattern/detail/TypeTraits.hpp
@@ -100,13 +100,11 @@ struct tuple_contains_Reducers<camp::tuple<Params...>>
 namespace type_traits
 {
 template<typename U, typename V>
-struct is_same_remove_const
-    : std::is_same<std::remove_const_t<U>, std::remove_const_t<V>>
+struct is_same_decay : std::is_same<std::decay_t<U>, std::decay_t<V>>
 {};
 
 template<typename U, typename V>
-static constexpr bool is_same_remove_const_v =
-    is_same_remove_const<U, V>::value;
+static constexpr bool is_same_decay_v = is_same_decay<U, V>::value;
 }  // namespace type_traits
 }  // namespace RAJA
 
--- a/include/RAJA/pattern/detail/TypeTraits.hpp
+++ b/include/RAJA/pattern/detail/TypeTraits.hpp
@@ -100,12 +100,13 @@ struct tuple_contains_Reducers<camp::tuple<Params...>>
 namespace type_traits
 {
 template<typename U, typename V>
-struct is_same_strip_const
+struct is_same_remove_const
     : std::is_same<std::remove_const_t<U>, std::remove_const_t<V>>
 {};
 
 template<typename U, typename V>
-static constexpr bool is_same_strip_const_v = is_same_strip_const<U, V>::value;
+static constexpr bool is_same_remove_const_v =
+    is_same_remove_const<U, V>::value;
 }  // namespace type_traits
 }  // namespace RAJA
 
--- a/include/RAJA/pattern/scan.hpp
+++ b/include/RAJA/pattern/scan.hpp
@@ -53,7 +53,7 @@ template<concepts::ExecutionPolicy ExecPolicy,
          concepts::Resource Res,
          concepts::RandomAccessRange Container,
          typename R = RAJA::detail::ContainerVal<Container>,
-         concepts::BinaryFunction<R, R, R> Function = operators::plus<R>>
+         concepts::BinaryFunction<R> Function = operators::plus<R>>
 RAJA_INLINE resources::EventProxy<Res> inclusive_scan_inplace(
     ExecPolicy&& p,
     Res r,
@@ -71,11 +71,11 @@ RAJA_INLINE resources::EventProxy<Res> inclusive_scan_inplace(
 }
 
 ///
-template<
-    concepts::ExecutionPolicy ExecPolicy,
-    concepts::RandomAccessRange Container,
-    typename Function = operators::plus<RAJA::detail::ContainerVal<Container>>,
-    typename Res      = typename resources::get_resource<ExecPolicy>::type>
+template<concepts::ExecutionPolicy ExecPolicy,
+         concepts::RandomAccessRange Container,
+         typename T = RAJA::detail::ContainerVal<Container>,
+         concepts::BinaryFunction<T> Function = operators::plus<T>,
+         concepts::Resource Res = resources::get_resource<ExecPolicy>::type>
 RAJA_INLINE resources::EventProxy<Res> inclusive_scan_inplace(
     ExecPolicy&& p,
     Container&& c,
@@ -102,7 +102,7 @@ template<concepts::ExecutionPolicy ExecPolicy,
          concepts::Resource Res,
          concepts::RandomAccessRange Container,
          typename T = RAJA::detail::ContainerVal<Container>,
-         concepts::BinaryFunction<T, T, T> Function = operators::plus<T>>
+         concepts::BinaryFunction<T> Function = operators::plus<T>>
 RAJA_INLINE resources::EventProxy<Res> exclusive_scan_inplace(
     ExecPolicy&& p,
     Res r,
@@ -121,11 +121,12 @@ RAJA_INLINE resources::EventProxy<Res> exclusive_scan_inplace(
 }
 
 ///
-template<concepts::ExecutionPolicy ExecPolicy,
-         concepts::RandomAccessRange Container,
-         typename T        = RAJA::detail::ContainerVal<Container>,
-         typename Function = operators::plus<T>,
-         typename Res      = typename resources::get_resource<ExecPolicy>::type>
+template<
+    concepts::ExecutionPolicy ExecPolicy,
+    concepts::RandomAccessRange Container,
+    typename T = RAJA::detail::ContainerVal<Container>,
+    concepts::BinaryFunction<T> Function = operators::plus<T>,
+    concepts::Resource Res = typename resources::get_resource<ExecPolicy>::type>
 RAJA_INLINE resources::EventProxy<Res> exclusive_scan_inplace(
     ExecPolicy&& p,
     Container&& c,
@@ -159,7 +160,7 @@ template<concepts::ExecutionPolicy ExecPolicy,
          concepts::RandomAccessRange OutContainer,
          typename T = RAJA::detail::ContainerVal<InContainer>,
          typename R = RAJA::detail::ContainerVal<OutContainer>,
-         concepts::BinaryFunction<R, T, R> Function = operators::plus<T>>
+         concepts::BinaryFunction<T> Function = operators::plus<T>>
 RAJA_INLINE resources::EventProxy<Res> inclusive_scan(
     ExecPolicy&& p,
     Res r,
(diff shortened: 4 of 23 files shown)


rhornung67 (approved):

include/RAJA/internal/Iterators.hpp
```diff
@@ -341,6 +341,20 @@ class strided_numeric_iterator
...
   }
 
+  RAJA_HOST_DEVICE inline strided_numeric_iterator operator++(int)
+  {
+    strided_numeric_iterator tmp(*this);
+    val += stride;
+    return tmp;
+  }
+
+  RAJA_HOST_DEVICE inline strided_numeric_iterator operator--(int)
```
johnbowen42, line 351: These are required to ensure strided_numeric_iterator satisfies the bidirectional concept

include/RAJA/pattern/scan.hpp
```diff
@@ -361,22 +349,20 @@ exclusive_scan(Res r, Args&&... args)
...
 {
   Res r = Res::get_default();
   return ::RAJA::policy_by_value_interface::inclusive_scan<ExecPolicy>(
       ExecPolicy(), r, std::forward<Args>(args)...);
 }
 
 ///
-template<typename ExecPolicy, typename Res, typename... Args>
+template<concepts::ExecutionPolicy ExecPolicy, typename Res, typename... Args>
 RAJA_INLINE concepts::enable_if_t<resources::EventProxy<Res>,
```
johnbowen42: I will make another pass here to remove SFINAE entirely here

include/RAJA/policy/hip/policy.hpp
```diff
@@ -829,7 +829,7 @@ struct IndexGlobal
 
   template<typename IdxT     = hip_dim_member_t,
            typename IdxNDims = NonCachedIndicesAndDims>
-  RAJA_DEVICE static constexpr IdxT size(IdxNDims const& idxNDims = IdxNDims {})
+  RAJA_DEVICE static constexpr IdxT size(IdxNDims const& = IdxNDims {})
```
johnbowen42, line 832: This is actually an unrelated change--but I was getting unused variable warnings from this file while working on the HIP backend.  I can remove and put in a separate PR if reviewers would like.

include/RAJA/pattern/scan.hpp
```diff
@@ -49,13 +49,12 @@ inline namespace policy_by_value_interface
...
     typename Res,
-    typename Container,
+    concepts::Range Container,
     typename Function = operators::plus<RAJA::detail::ContainerVal<Container>>>
 RAJA_INLINE concepts::enable_if_t<
     resources::EventProxy<Res>,
-    type_traits::is_execution_policy<ExecPolicy>,
     type_traits::is_resource<Res>,
     std::is_constructible<camp::resources::Resource, Res>,
     type_traits::is_range<Container>>
```
bechols97: Can the `type_traits::is_range<Container>` be removed now since we are using the concept version in the template parameter?

include/RAJA/index/RangeSegment.hpp
```diff
@@ -596,23 +596,23 @@ namespace concepts
...
 
 template<typename T, typename U>
-struct RangeConstructible
-    : DefineConcept(camp::val<RAJA::detail::common_type_t<T, U>>()) {};
+concept RangeConstructible = std::common_with<T, U>;
 
 template<typename T, typename U, typename V>
-struct RangeStrideConstructible
-    : DefineConcept(camp::val<RAJA::detail::common_type_t<T, U, V>>()) {};
+concept RangeStrideConstructible = RAJA::detail::common_type_t<T, U, V>::value;
```
MrBurmark: Are we just trying to ask if there is a common type between T, U, and V?

include/RAJA/pattern/detail/TypeTraits.hpp
```diff
@@ -96,6 +96,17 @@ struct tuple_contains_Reducers<camp::tuple<Params...>>
 {};
 
 }  // namespace expt
+
+namespace type_traits
+{
+template<typename U, typename V>
+struct is_same_strip_const
```
MrBurmark: How about a name like `is_same_without_const`, or `is_same_removing_const`?

include/RAJA/pattern/scan.hpp
```diff
@@ -112,14 +109,13 @@ inclusive_scan_inplace(ExecPolicy&& p,
...
-         typename Container,
+         concepts::Range Container,
          typename T        = RAJA::detail::ContainerVal<Container>,
          typename Function = operators::plus<T>>
 RAJA_INLINE concepts::enable_if_t<
     resources::EventProxy<Res>,
-    type_traits::is_execution_policy<ExecPolicy>,
     type_traits::is_resource<Res>,
     std::is_constructible<camp::resources::Resource, Res>,
     type_traits::is_range<Container>>
```
MrBurmark:
```suggestion
std::is_constructible<camp::resources::Resource, Res>>
```

include/RAJA/pattern/scan.hpp
```diff
@@ -182,15 +176,14 @@ exclusive_scan_inplace(ExecPolicy&& p,
 *begin))}
 ******************************************************************************
 */
-template<typename ExecPolicy,
+template<concepts::ExecutionPolicy ExecPolicy,
          typename Res,
          typename InContainer,
          typename OutContainer,
```
MrBurmark:
```suggestion
concepts::Range InContainer,
concepts::Range OutContainer,
```

include/RAJA/pattern/scan.hpp
MrBurmark:
```suggestion
std::is_constructible<camp::resources::Resource, Res>>
```

test/include/RAJA_test-index-types.hpp
```diff
@@ -25,7 +25,9 @@
 //
 // Strongly typed indexes
 //
-RAJA_INDEX_VALUE(StrongIndexType, "StrongIndexType");
+// TODO(bowen): Did we ever support passing these types to Span?  They do not
+// model Integral
```
MrBurmark: Good question, did the tests using StrongIdxTypeList work before, but don't now?

include/RAJA/pattern/scan.hpp
johnbowen42: It is "imported" via `using namespace camp::concepts` here https://github.com/llnl/RAJA/blob/e951ec407a515227f0f0b13dd73751ce62384986/include/RAJA/util/concepts.hpp#L35

johnbowen42: But maybe we can delete it from camp?  I am not sure if there are users of RAJA/camp's concepts besides internally within RAJA

bechols97: Oh sorry, I wasn't very clear in the suggestion. I just meant remove the extra `type_traits::is_range<Container>` check from this `enable_if_t` condition. I think @MrBurmark made a similar suggestion below.

MrBurmark: Probably, as its probably not being used outside of RAJA.

include/RAJA/index/RangeSegment.hpp
artv3: If we add in , int, long, double, they won't have the ::value defined. 
```suggestion
concept RangeStrideConstructible = requires { typename RAJA::detail::common_type_t<T, U, V>; };
```

include/RAJA/util/Operators.hpp
```diff
@@ -768,35 +768,34 @@ struct safe_plus
...
+    : std::bool_constant<
+          RAJA::concepts::BinaryFunction<Function, Return, Arg1, Arg2>>
+{};
+template<class Function, class Return, class Arg1 = Return, class Arg2 = Arg1>
+inline constexpr bool is_binary_function_v =
+    is_binary_function<Function, Return, Arg1, Arg2>::value;
+
+template<class Function, class Return, class Arg = Return>
+struct is_unary_function
+    : std::bool_constant<RAJA::concepts::BinaryFunction<Function, Return, Arg>>
```
artv3:
```suggestion
    : std::bool_constant<RAJA::concepts::UnaryFunction<Function, Return, Arg>>
```
Is that what you meant?

include/RAJA/util/Operators.hpp
```diff
@@ -768,35 +768,34 @@ struct safe_plus
...
+{};
+template<class Function, class Return, class Arg1 = Return>
+inline constexpr bool is_unary_function_v =
+    is_binary_function<Function, Return, Arg1>::value;
```
artv3: Don't we want unary_function?  Maybe I'm missing something?

include/RAJA/policy/PolicyBase.hpp
```diff
@@ -187,15 +188,15 @@ namespace concepts
...
-    : DefineConcept(::RAJA::concepts::has_type<::RAJA::Policy>(
-                        camp::decay<decltype(Pol::policy)>()),
-                    ::RAJA::concepts::has_type<::RAJA::Pattern>(
-                        camp::decay<decltype(Pol::pattern)>()),
-                    ::RAJA::concepts::has_type<::RAJA::Launch>(
-                        camp::decay<decltype(Pol::launch)>()),
-                    ::RAJA::concepts::has_type<::RAJA::Platform>(
-                        camp::decay<decltype(Pol::platform)>())) {};
+concept ExecutionPolicy =
+    RAJA::type_traits::is_same_strip_const_v<decltype(Pol::policy),
```
artv3: what about using the decay semantics?  we can probably express this with features from std?

include/RAJA/util/Span.hpp
```diff
@@ -57,7 +58,7 @@ namespace RAJA
  *   compile time extents
  *
  */
-template<typename IterType, typename IndexType>
+template<concepts::RandomAccessIterator IterType, concepts::Integral IndexType>
```
artv3: Will this still work with strong index types?

include/RAJA/policy/PolicyBase.hpp
artv3: I think this may be more narrow than what we had before, this removes const while the code the previous code used decay allowing more types to go through.

artv3:
FYI: I did a short review. Once the first batch of comments are addressed, I am happy to revisit the code.

test/include/RAJA_test-index-types.hpp
johnbowen42: yes they did--I am trying to figure out why now.  Somehow we were instantiating `Span` with `StrongIndexType` before

johnbowen42: There was a bug with the old concept approach--if you replace 
`  static_assert(type_traits::is_integral<IndexType>::value, "IndexType must model Integral");` with `std::is_integral_v` inside the Span class template definition in main the same tests fail to compile

include/RAJA/index/RangeSegment.hpp
johnbowen42: @MrBurmark @bechols97 @artv3 Does anybody know if these concepts/type_traits were actually used externally?  They are not used anywhere within RAJA

johnbowen42: I removed this concept altogether.  Its unused in RAJA.  I doubt its used externally but it is possible, do you know of any cases?

include/RAJA/pattern/detail/TypeTraits.hpp
johnbowen42: renamed

test/include/RAJA_test-index-types.hpp
johnbowen42: Ok, so I changed the kernel tests (kernel uses span to create segments) to use a different list defined by this header.  Let me know what you think

include/RAJA/index/RangeSegment.hpp
johnbowen42: I deleted this in the latest commit, let me know if I should add it back

include/RAJA/util/Span.hpp
johnbowen42: Nope, and it never should have based on the static assert before. See here https://github.com/llnl/RAJA/pull/2020#discussion_r3237631793

include/RAJA/policy/PolicyBase.hpp
johnbowen42: hmm that's a good point.  I don't see any build errors as a result of this internally in our test suite but I'll go use decay instead

include/RAJA/pattern/scan.hpp
```diff
@@ -332,24 +285,24 @@ exclusive_scan(ExecPolicy&& p,
...
  */
-template<typename ExecPolicy,
+template<concepts::ExecutionPolicy ExecPolicy,
          typename... Args,
          typename Res = typename resources::get_resource<ExecPolicy>::type>
-RAJA_INLINE concepts::enable_if_t<resources::EventProxy<Res>,
-                                  type_traits::is_execution_policy<ExecPolicy>>
-exclusive_scan(Args&&... args)
+RAJA_INLINE concepts::enable_if_t<resources::EventProxy<Res>> exclusive_scan(
+    Args&&... args)
```
bechols97: This could be my lack of knowledge on how CAMP's `enable_if_t` works. But, is the `enable_if_t` here needed anymore?

```suggestion
RAJA_INLINE resources::EventProxy<Res> exclusive_scan(Args&&... args)
```

include/RAJA/index/RangeSegment.hpp
MrBurmark: I don't know of any.

test/include/RAJA_test-index-types.hpp
MrBurmark: I guess what we really want is something to convertible to integral?

MrBurmark: Is it true that the type_traits one checked that it looked like it was an integral, but is_integral actually checks that its one of the basic types?

include/RAJA/index/RangeSegment.hpp
MrBurmark: Its possible this was used at some point, or was planned to be used, but not used. We'll see if anyone notices.

include/RAJA/pattern/scan.hpp
MrBurmark: I think that was a no-op if you didn't pass in any type_traits in addition to the "return" type.

test/include/RAJA_test-index-types.hpp
johnbowen42: Do we want something convertible to an int inside Span?  Enums are convertible to ints, but I don't think that is a reasonable thing to allow.  `is_integral` checks if its any signed or unsigned integral type

johnbowen42: I also don't think that the old concept was checking if something was convertible to integer

MrBurmark: We probably want to disallow certain things can look similar or convert to ints like pointers and enums.

johnbowen42: Isn't that what our `concepts::Integral` concept is doing here?

MrBurmark: But I think that we do want to support types like our strongly typed index types that look like ints.

johnbowen42: I appreciate that we want to support that, but StrongTypeIndex did not look like an int (see https://github.com/llnl/RAJA/blob/eb00eb5fceca16f2fb522e18f406384225bfc8ad/test/include/RAJA_test-index-types.hpp#L28). Even the index types that are meant to look like ints here https://github.com/llnl/RAJA/blob/eb00eb5fceca16f2fb522e18f406384225bfc8ad/test/include/RAJA_test-index-types.hpp#L63 are commented out of the test suite.  I think our "support" for that index type through our test suite was largely an accident of the implementation of camp's concepts

johnbowen42: I think we should open an issue and revisit this aspect of the testing suite separate from this PR, because it looks like we were trying to support these types of indices but not actually supporting them in reality

johnbowen42: I'm open to making the change here as well, but I think it would expand the scope of this PR

johnbowen42: @artv3  @MrBurmark  Ok I thought about this more and I think I have a reasonable solution--see  the new `SpanIndex` concept here https://github.com/llnl/RAJA/pull/2020/changes#diff-38e454fee010a2097ab28fc9773b94e23501f140ce26ebd7acb5da2402f0e052R317 and https://github.com/llnl/RAJA/pull/2020/changes#diff-b36ff9f682b4cc0577fac7779445c486b66e2abef955db1227a7b9cca3128ff7R68.  I tried using the same trick with the other `StrongInt` and `StrongULL` types in the https://github.com/llnl/RAJA/blob/4ea83c49c874476fc0bec9500902ef527d240a95/test/include/RAJA_test-index-types.hpp, but compilation failed here, so that is still an open task

It's a bit ambiguous whether Span was intended to be used this way.  The tests seem to make it seem so, but the old static asserts seem to indicate otherwise.  In any case, this new concept fixes compilation and restores test coverage.

bechols97 (approved):

include/RAJA/pattern/scan.hpp
```diff
@@ -48,29 +49,21 @@ inline namespace policy_by_value_interface
...
+         concepts::RandomAccessRange Container,
+         typename R = RAJA::detail::ContainerVal<Container>,
+         concepts::BinaryFunction<R, R, R> Function = operators::plus<R>>
 RAJA_INLINE concepts::enable_if_t<
     resources::EventProxy<Res>,
-    type_traits::is_execution_policy<ExecPolicy>,
-    type_traits::is_resource<Res>,
-    std::is_constructible<camp::resources::Resource, Res>,
-    type_traits::is_range<Container>>
+    std::is_constructible<camp::resources::Resource, Res>>
```
MrBurmark: Is this not required for anything that satisfies `concepts::Resource`?

include/RAJA/pattern/scan.hpp
```diff
@@ -299,19 +255,16 @@ exclusive_scan(ExecPolicy&& p,
 }
 
 ///
-template<typename ExecPolicy,
-         typename InContainer,
-         typename OutContainer,
+template<concepts::ExecutionPolicy ExecPolicy,
+         concepts::Range InContainer,
+         concepts::Range OutContainer,
```
MrBurmark: Shouldn't these be RandomAccessRanges, given that this is just a wrapper around a function that requires that?

include/RAJA/policy/hip/policy.hpp
MrBurmark: If this is in hip, then it should also be in cuda.

include/RAJA/util/Span.hpp
```diff
@@ -57,7 +61,14 @@ namespace RAJA
  *   compile time extents
  *
  */
-template<typename IterType, typename IndexType>
+
+namespace concepts
+{
+template<typename T>
+concept SpanIndex = concepts::Integral<T> || concepts::IndexValued<T>;
```
MrBurmark, line 68: I imagine we're going to need this concept more generally in the future, so it should probably be in a more general location and have a more general name.

test/unit/index/test-indexvalue.cpp
```diff
@@ -280,3 +280,9 @@ TYPED_TEST(IndexValueUnitTest, IndexTypeCompare)
   ASSERT_NE(x, x_lower);
   ASSERT_NE(x, x_higher);
 }
+
+TYPED_TEST(IndexValueUnitTest, IndexValueConcept)
+{
+  Tester<Index> test1; // should compile
+  // Tester<Ty> test2; // should not compile
```
MrBurmark: `static_assert(requires...)`?

include/RAJA/util/Span.hpp
johnbowen42: In general, I have been wondering if our approach to creating concepts across many headers is the right approach.  Should we try to consolidate all our definitions into a single header?  I feel like this might be useful for understanding RAJA for new developers

artv3: I agree with that

MrBurmark: I think that would be good for most of the general concepts. I wonder how many things we'll have to move around so we actually can do this.

test/include/RAJA_test-index-types.hpp
MrBurmark: If we think of Span as being similar to View then yes, but if its just something that gives you begin and end iterators then maybe not.

include/RAJA/util/Span.hpp
johnbowen42: Yeah, that refactor might be outside the scope of this PR, but I think it could be useful.  Kokkos, for example, does this

johnbowen42: https://github.com/llnl/RAJA/issues/2026

test/include/RAJA_test-index-types.hpp
MrBurmark: Looking at `Span` a little more closely, we probably don't really want `Span::size` to return a strongly typed index, but it would with the current implementation. So I don't think `Span` was designed for strongly typed indices. I agree let's keep coverage for now and fix this more generally later.

include/RAJA/pattern/scan.hpp
johnbowen42: eliminated all sfinae here

include/RAJA/pattern/scan.hpp
johnbowen42: all scan operations require RandomAccessRange, so I changed that

include/RAJA/pattern/scan.hpp
```diff
@@ -81,19 +72,14 @@ inclusive_scan_inplace(ExecPolicy&& p,
 
 ///
 template<
-    typename ExecPolicy,
-    typename Container,
+    concepts::ExecutionPolicy ExecPolicy,
+    concepts::RandomAccessRange Container,
     typename Function = operators::plus<RAJA::detail::ContainerVal<Container>>,
     typename Res      = typename resources::get_resource<ExecPolicy>::type>
```
MrBurmark: concepts?

include/RAJA/pattern/scan.hpp
```diff
@@ -48,29 +49,19 @@ inline namespace policy_by_value_interface
...
-    type_traits::is_range<Container>>
-inclusive_scan_inplace(ExecPolicy&& p,
-                       Res r,
-                       Container&& c,
-                       Function binop = Function {})
+template<concepts::ExecutionPolicy ExecPolicy,
+         concepts::Resource Res,
+         concepts::RandomAccessRange Container,
+         typename R = RAJA::detail::ContainerVal<Container>,
+         concepts::BinaryFunction<R, R, R> Function = operators::plus<R>>
```
MrBurmark: This was called `T` below.

johnbowen42: Not sure what you mean?  It's named something else in the function below??

include/RAJA/pattern/scan.hpp
johnbowen42: I'm not sure why we would use concepts for these two--they are concrete types defined from two templated types constrained by concepts already.

MrBurmark: For `Function` the type can be deduced from the arguments. For `Res` I was just thinking about using things consistently more to consistently document intent.

johnbowen42: I went through and used concepts where it made sense.  It was not possible everywhere, for example our iterator concepts from camp all require `!Integral` (we may want to revisit the usefulness of this) so I had to keep `typename` for `ContainerVal`

MrBurmark: Nice, I think that `ContainerVal` makes sense to just typename since its whatever the user had in their container that works with the given function.

include/RAJA/pattern/scan.hpp
MrBurmark: You are right, it did use R, nevermind.

MrBurmark (approved):

PR #2021, opened 2026-04-30
RAJA::fornest : Abstraction to switch between collapse, nested loops, and tiling patterns

artv3:
#
I along with @tomstitt work an application where we have been exploring using a 1D GPU index for multi-level loops and have found that may be more performant than using hierarchical parallelism. This PR uses concepts from RAJA to create the RAJA::forall_nd convenience function. 

Cases in which this approach is more performant:
 it comes up when the threads per block are not high enough to saturate the GPU and using ``RAJA::forall + mods + divs`` allows us to increase the threads per block which ends up being more performant.


MrBurmark:
I am concerned that people will use this without knowing that its doing expensive div/mod calculations and see that it performs more poorly than a native cuda/hip 2d or 3d kernel. Would it make sense to have policies that allow mapping to 2d or 3d kernels.

artv3:
> I am concerned that people will use this without knowing that its doing expensive div/mod calculations and see that it performs more poorly than a native cuda/hip 2d or 3d kernel. Would it make sense to have policies that allow mapping to 2d or 3d kernels.

I think that can be addressed in the RAJA cookbook or examples explaining when this would be performant. In the case of @tomstitt and I, it comes up when the threads per block are not high enough to saturate the GPU and using ``RAJA::forall + mods + divs`` allows us to increase the threads per block which ends up being more performant. For 2D/3D gpu grids there are various ways to do that and perhaps we should direct developers to ``RAJA::launch or RAJA::kernel``?

tomstitt:
> > I am concerned that people will use this without knowing that its doing expensive div/mod calculations and see that it performs more poorly than a native cuda/hip 2d or 3d kernel. Would it make sense to have policies that allow mapping to 2d or 3d kernels.
> 
> 
> 
> I think that can be addressed in the RAJA cookbook or examples explaining when this would be performant. In the case of @tomstitt and I, it comes up when the threads per block are not high enough to saturate the GPU and using ``RAJA::forall + mods + divs`` allows us to increase the threads per block which ends up being more performant. For 2D/3D gpu grids there are various ways to do that and perhaps we should direct developers to ``RAJA::launch or RAJA::kernel``? 

I think my ideal is an interface where there is a choice of policy. We have 2d/3d kernels on 1d iteration spaces, using mod/div like Arturo said, to expose more parallelism. When we switch some of those to using our "true" 2d/3d grid launcher we lose performance because our  block (16x16 , 8x8x8) doesn't map well onto the grid (because we just idle threads). It's of course not always true that our div/mod approach is going to be better, like Jason said, and if we had an easy way to pick we could put both behind our abstraction and correctly dispatch

artv3:
> > > I am concerned that people will use this without knowing that its doing expensive div/mod calculations and see that it performs more poorly than a native cuda/hip 2d or 3d kernel. Would it make sense to have policies that allow mapping to 2d or 3d kernels.
> > 
> > 
> > I think that can be addressed in the RAJA cookbook or examples explaining when this would be performant. In the case of @tomstitt and I, it comes up when the threads per block are not high enough to saturate the GPU and using `RAJA::forall + mods + divs` allows us to increase the threads per block which ends up being more performant. For 2D/3D gpu grids there are various ways to do that and perhaps we should direct developers to `RAJA::launch or RAJA::kernel`?
> 
> I think my ideal is an interface where there is a choice of policy. We have 2d/3d kernels on 1d iteration spaces, using mod/div like Arturo said, to expose more parallelism. When we switch some of those to using our "true" 2d/3d grid launcher we lose performance because our block (16x16 , 8x8x8) doesn't map well onto the grid (because we just idle threads). It's of course not always true that our div/mod approach is going to be better, like Jason said, and if we had an easy way to pick we could put both behind our abstraction and correctly dispatch

@tomstitt , okay -- Let's get that imagination soaring, and cook up some ideas!  #RAJA!

examples/launch_nd.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+  auto cells = RAJA::TypedRangeSegment<int>(0, num_cells);
+  auto comps = RAJA::TypedRangeSegment<int>(0, num_comp);
+
+  // _launch_nd_call_start
+  RAJA::launch_nd(res, policy, RAJA::segments(cells, comps),
+                  [=] RAJA_HOST_DEVICE(int cell, int comp) {
+                    const int idx   = comp + num_comp * cell;
+                    values_ptr[idx] = 1000 * cell + comp;
+                  });
```
artv3: @tomstitt , @MrBurmark I invite you to take a look at this example, I think this may be what we are looking for.

MrBurmark:
> > > > I am concerned that people will use this without knowing that its doing expensive div/mod calculations and see that it performs more poorly than a native cuda/hip 2d or 3d kernel. Would it make sense to have policies that allow mapping to 2d or 3d kernels.
> > > 
> > > 
> > > I think that can be addressed in the RAJA cookbook or examples explaining when this would be performant. In the case of @tomstitt and I, it comes up when the threads per block are not high enough to saturate the GPU and using `RAJA::forall + mods + divs` allows us to increase the threads per block which ends up being more performant. For 2D/3D gpu grids there are various ways to do that and perhaps we should direct developers to `RAJA::launch or RAJA::kernel`?
> > 
> > 
> > I think my ideal is an interface where there is a choice of policy. We have 2d/3d kernels on 1d iteration spaces, using mod/div like Arturo said, to expose more parallelism. When we switch some of those to using our "true" 2d/3d grid launcher we lose performance because our block (16x16 , 8x8x8) doesn't map well onto the grid (because we just idle threads). It's of course not always true that our div/mod approach is going to be better, like Jason said, and if we had an easy way to pick we could put both behind our abstraction and correctly dispatch
> 
> @tomstitt , okay -- Let's get that imagination soaring, and cook up some ideas! #RAJA!

You certainly can use launch or teams to get some level of parallelism and then take those indices and do your own calculations with them. If we had the multiloop abstractions with a variety of policies that could take us a fair amount of the way.

artv3:
@llnl/raja-core , I received an approval from my project. Opening up the PR for review from the squad.

adrienbernede (commented):
@lc-hubcast approve

rhornung67:
@artv3 I merged in latest develop and ran clang-format so GHA CI check run. Is this ready to be merged when approved?

artv3:
> @artv3 I merged in latest develop and ran clang-format so GHA CI check run. Is this ready to be merged when approved?

Not yet, I think we want @llnl/raja-core to take another look. We had a discussed a while potentially changing the name of the abstraction

docs/sphinx/user_guide/cook_book/launch-nd.rst
```diff
@@ -0,0 +1,116 @@
...
+.. _cook-book-launch-nd-label:
+
+============================
+Cooking with RAJA::launch_nd
+============================
+
+``RAJA::launch_nd`` runs a logical 2-D or 3-D loop body through selectable
+launch-backed mappings. It is intended for kernels where the source code should
+stay written in logical multi-dimensional indices, but the best GPU mapping is
+not known from the source alone.
```
MrBurmark: The GPU is part of this, but wouldn't this also be useful for doing things like collapsed openmp loops, or reordered sequential loops?

docs/sphinx/user_guide/cook_book/launch-nd.rst
```diff
@@ -0,0 +1,116 @@
...
+can still win when the dimensions fit the block shape, when the body is very
+small and index reconstruction dominates, or when direct multi-dimensional
+mapping improves memory access or scheduling.
+
+----------------
+Mapping Policies
+----------------
+
+The example source defines backend-specific policy aliases. CUDA uses direct
+global loop policies for the true grid mapping:
```
MrBurmark: It is confusing to have direct and loop in close proximity here.

docs/sphinx/user_guide/cook_book/launch-nd.rst
```diff
@@ -0,0 +1,116 @@
...
+
+.. literalinclude:: ../../../../examples/launch_nd.cpp
+   :start-after: // _launch_nd_policy_aliases_start
+   :end-before: // _launch_nd_policy_aliases_end
+   :language: C++
+
+The flattened policy uses a forall-style execution policy such as
+``RAJA::cuda_exec<block_size_1d>`` or ``RAJA::hip_exec<block_size_1d>``. The
+grid policy uses ``RAJA::LaunchPolicy`` and direct global loop policies such as
+``RAJA::cuda_global_y_direct`` and ``RAJA::cuda_global_x_direct``.
```
MrBurmark: Could you use `RAJA::cuda_global_yx_direct`?

MrBurmark: If people want to do more complicated things like mixing direct and loop policies they could always use launch itself.

docs/sphinx/user_guide/cook_book/launch-nd.rst
```diff
@@ -0,0 +1,116 @@
...
+--------------------
+Current Capabilities
+--------------------
+
+``RAJA::launch_nd`` currently supports ``RAJA::TypedRangeSegment`` packs created
+with ``RAJA::segments``. The grid mapping supports 2-D and 3-D loops and
+requires one loop policy per segment. The flattened mapping uses
+``RAJA::layout_right`` by default, or ``RAJA::layout_left`` when that layout tag
+is supplied, to control which logical index is unit stride in the flattened
+space.
```
MrBurmark: Why not just a tuple of segments?
The layouts in the policy are a way to reorder the loops?

include/RAJA/pattern/launch_nd.hpp
```diff
@@ -0,0 +1,705 @@
...
+}
+
+#if defined(RAJA_GPU_ACTIVE)
+template<typename PolicyList>
+RAJA_INLINE auto make_launch_nd_context(RAJA::resources::Resource resource,
+                                        std::string&& kernel_name)
+{
+  return resource.get_platform() == RAJA::Platform::host
+             ? util::make_context<typename PolicyList::host_policy_t>(
+                   std::move(kernel_name))
```
rchen20: If we are checking for `RAJA::Platform::host` here, does this block need to be in `RAJA_GPU_ACTIVE`?

docs/sphinx/user_guide/cook_book/launch-nd.rst
artv3: Underneath we just map to regular nested loops in launch, I think we would have to leave that for when we do the expanded version of this. One thing to explore is transitioning between flat loops to nested loops using threads in a GPU setting. We also would have to figure out the backend on how that would look like for OpenMP

artv3: I suppose it could, I haven't thought about that yet though  actually, I think that could be a neat expansion. I think that goes beyond the targeted scope of this PR though, but something to definitely revisit as we expand capabilities. I would be interested in testing it some kernels in RAJAPerf

docs/sphinx/user_guide/cook_book/launch-nd.rst
artv3: good call, I updated the example

docs/sphinx/user_guide/cook_book/launch-nd.rst
artv3: I agree actually, that would good but we would need to first update formalize support “multi-dim per-loop” policies. Something I've played around with but have not yet formalized

docs/sphinx/user_guide/cook_book/launch-nd.rst
artv3: In some sense it already is, `struct TypedRangeSegmentPack { camp::tuple<RAJA::TypedRangeSegment<IdxTs>...> data; };` , the RAJA::Segments just enforces that we use pass in a RAJA::Segment type? Are you thinking of support for non-RAJA types? 

Second question: 
For the flatten policy it controls the linearization order (which logical index
  varies fastest / is unit-stride)

For the grid policy: 
 it’s determined by the segment + loop-
  policy order (one policy per segment)

include/RAJA/pattern/launch_nd.hpp
artv3: yes, because in GPU builds we can still dispatch to the CPU at run time.

include/RAJA/pattern/launch_nd.hpp
```diff
@@ -0,0 +1,705 @@
...
+#include "RAJA/policy/sycl.hpp"
+#endif
+#include "RAJA/util/CombiningAdapter.hpp"
+#include "RAJA/util/View.hpp"
+
+namespace RAJA
+{
+
+template<typename... IdxTs>
+struct TypedRangeSegmentPack
```
rchen20: Do we want `TypedRangeSegmentPack` to be in the outward facing RAJA API?

include/RAJA/pattern/launch_nd.hpp
```diff
@@ -0,0 +1,705 @@
...
+{
+  camp::tuple<RAJA::TypedRangeSegment<IdxTs>...> data;
+};
+
+template<typename... IdxTs>
+RAJA_INLINE auto segments(RAJA::TypedRangeSegment<IdxTs> const&... segs)
```
rchen20: Should we rename `RAJA::segments` to something else closer to its `launch_nd`-only usage? Or hide it better in a deeper namespace so that people don't accidentally use `segments`?

include/RAJA/pattern/launch_nd.hpp
```diff
@@ -0,0 +1,705 @@
...
+                                     DeviceExecPolicy,
+                                     LayoutTag>,
+    TypedRangeSegmentPack<IdxTs...> const& segs,
+    Params&&... params)
+{
+  switch (place)
+  {
+    case ExecPlace::HOST:
+      return RAJA::launch_nd(
+          resource, launch_nd_flattened_policy<HostExecPolicy, LayoutTag> {},
```
rchen20: Do we need to validate that the `resource` matches the `ExecPlace`?

include/RAJA/pattern/launch_nd.hpp
```diff
@@ -0,0 +1,705 @@
...
+ * This allows callers to select different launch_nd policy *kinds* for host and
+ * device (e.g., a grid policy on host for nested loops and a flattened policy
+ * on device for a 1D mapping).
+ */
+template<typename HostPolicy, typename DevicePolicy>
+struct launch_nd_place_policy
+{
+  HostPolicy host;
+  DevicePolicy device;
+};
```
rchen20: I know these serve two different overloads of `launch_nd()` below, but could these be combined for simplification? Maybe just have `launch_nd_flattened_place_policy`?

include/RAJA/pattern/launch_nd.hpp
artv3: good, point that should be internal, since it is specific to this API

include/RAJA/pattern/launch_nd.hpp
artv3: Agreed, renamed to  RAJA::nd_segments(...) and updated docs/tests/examples accordingly

include/RAJA/pattern/launch_nd.hpp
artv3: yes, good call, added a helper to check

include/RAJA/pattern/launch_nd.hpp
artv3: good idea, I reduced duplication by making launch_nd_flattened_place_policy just a convenience alias of the general launch_nd_place_policy, and removed the now-
  redundant launch_nd(...) overloads.

artv3:
Thanks @rchen20 , I addressed your review comments

include/RAJA/pattern/launch_nd.hpp
```diff
@@ -0,0 +1,714 @@
...
+template<typename HostPolicy,
+         typename DevicePolicy,
+         detail::typed_range_segment_pack SegmentPack,
+         typename... Params>
+void launch_nd(ExecPlace place,
+               launch_nd_place_policy<HostPolicy, DevicePolicy> const& policy,
+               SegmentPack const& segs,
+               Params&&... params)
+{
+  switch (place)
```
rchen20: We probably need `validate_launch...` here too. Would you mind checking where else we need to call it?

artv3: got it! thanks!

docs/sphinx/user_guide/cook_book/launch-nd.rst
MrBurmark: We can have a concept that each type in the tuple is a segment.

MrBurmark:
This API can be thought of as either an extension of `forall` to multiple dimensions, or a simplification of `launch`/`kernel` for perfectly nested loops.
The API currently uses `nd` but there are other options like `nested` if we want to use a whole word instead of an abbreviation. I wasn't sure what `nd` meant the first time I saw it.
The API is `launch`-like in that it takes a place to run and a dual host+device policy. We also have the `dynamic_forall` API which has an alternate way of specifying multiple policies and picking one at runtime. What kind of API for multiple policy support do we want to have?

rhornung67:
What about something like `forall_flattened`?

MrBurmark:
> What about something like `forall_flattened`?

I don't think we always want to flatten/collapse the index space.

rhornung67:
`forall_flattened_maybe`???? 😆

MrBurmark:
That's another part of the naming that can be a bit confusing here. We have a couple cases of things that we can flatten/collapse. Perhaps we should choose a consistent word for each direction.
In this case we are basically doing what `omp collapse` does, so maybe we should use `collapse`. We are taking a multi-dimensional index space and flattening it so it can be run with a single-dimensional loop.
We already have the opposite case in policies like `cuda_flatten_thread_xy_direct`, so maybe we should just use `flatten` for this case. We take a single-dimensional index space and run it via a multiple-dimensional loop by flattening the loops.
We also do something similar when we calculate single-dimensional offsets for multi-dimensional data via layouts' call operator, we can also do the opposite via `toIndices`.

MrBurmark:
We discussed the name `RAJA::fornest` as in for a nest of loops.
We also discussed having a notion of dimensionality for policies so something like `cuda_global_xy_direct` is a 2 dimensional policy. The basic exec policies like `cuda_exec<256>` would be non-dimensional and could be a way to request use of a default policy like `cuda_global_xy_direct` for 2 dimensional loops. We would decide the tile size, for example defaulting to something like 32x8 or 16x16 for 256 threads and adapting based on the size of the index segments, for example 256x1 if the second index segment is of size 1.

artv3:
@llnl/raja-core,   --- RAJA::forall runs parallel loops; RAJA::fornest runs parallel bird nests, because every loop needs somewhere to hatch its iterations.

docs/sphinx/user_guide/cook_book/fornest.rst
```diff
@@ -0,0 +1,123 @@
...
+===========================
+
+``RAJA::fornest`` runs a logical 2-D or 3-D loop body through selectable
+mappings. It is intended for kernels where the source code should stay written
+in logical multi-dimensional indices, but the best GPU mapping is not known from
+the source alone.
+
+The interface supports two mapping policy families:
+
+ * ``RAJA::fornest_flattened_policy<ExecPolicy, LayoutTag>`` maps the product
```
MrBurmark: I thought we were going to call this `collapse` instead of a `flatten` as this is a multi-dimensional index space being mapped to a 1-D exec space?

artv3: Absolutely, we can call it a collapse. Thanks for catching that, I sometimes use the application terminology out of habit.

docs/sphinx/user_guide/cook_book/fornest.rst
```diff
@@ -0,0 +1,123 @@
...
+.. _cook-book-fornest-label:
+
+===========================
+Cooking with RAJA::fornest
+===========================
+
+``RAJA::fornest`` runs a logical 2-D or 3-D loop body through selectable
+mappings. It is intended for kernels where the source code should stay written
+in logical multi-dimensional indices, but the best GPU mapping is not known from
+the source alone.
```
MrBurmark: Should this dwell on the current support (2d and 3d) or be general and then note that currently only 2d and 3d are supported. The second sentence is a bit awkward and specific to a certain GPU use case when this API is intended to be fairly general for kernels with perfectly nested loops. Maybe something like "Use this for kernels with multi-dimensional indexing that can be expressed sequentially with perfectly nested loops". The "GPU mapping" statement seems like something too specific for the intro.

docs/sphinx/user_guide/cook_book/fornest.rst
```diff
@@ -0,0 +1,123 @@
...
+The interface supports two mapping policy families:
+
+ * ``RAJA::fornest_flattened_policy<ExecPolicy, LayoutTag>`` maps the product
+   of the logical dimensions to a 1-D iteration space. The ``ExecPolicy`` is a
+   regular forall-style policy such as ``RAJA::device_exec<256>`` (CUDA/HIP) or
+   ``RAJA::seq_exec`` (host). RAJA performs the linear-to-logical index
+   reconstruction internally.
+ * ``RAJA::fornest_mapping_policy<ExecPolicy, LoopPolicies...>`` maps the logical
+   dimensions directly using one loop policy per dimension. For CUDA/HIP this
+   allows explicit mapping to global, block, or thread spaces.
```
MrBurmark: Regarding the naming and structure of these policies, are these aliases to a single underlying policy template? I am imagining something like `RAJA::fornest_policy<OuterExecPolicy, InnerPolicyList=DefaultList>` where in some cases an outer policy can have enough information for a basic case without specifying the inner policies.

include/RAJA/pattern/fornest.hpp
```diff
@@ -0,0 +1,3006 @@
...
+#if defined(RAJA_ENABLE_SYCL)
+#include "RAJA/policy/sycl.hpp"
+#endif
+
+namespace RAJA
+{
+
+/*!
+ * Policy wrapper that requests a collapsed mapping for a rank-2 or rank-3
+ * fornest.
```
MrBurmark: These should be agnostic to 2d vs 3d.

include/RAJA/pattern/fornest.hpp
```diff
@@ -0,0 +1,3006 @@
...
+ *
+ * With a collapsed policy, `RAJA::fornest` linearizes the 2D/3D iteration
+ * space and executes a single 1D `RAJA::forall` over that linear index.
+ * The `LayoutTag` controls how the linear index is mapped back to `(i0,i1)`
+ * or `(i0,i1,i2)` (e.g., `RAJA::layout_right` vs `RAJA::layout_left`).
+ *
+ * Notes:
```
MrBurmark: These notes have nothing to do with this policy.

artv3:
@MrBurmark @bechols97 , if you have some time can you take a look at the fornest-basic.cpp   and dynamic-fornest.cpp ? I want to double check I captured the desired capabilities, then we can start wrapping this up.

docs/sphinx/user_guide/cook_book/fornest.rst
```diff
@@ -0,0 +1,269 @@
...
+each logical point, regardless of whether the policy results in nested loops, a
+collapsed 1-D traversal, or a device launch.
+
+.. note::
+   ``RAJA::fornest`` is intended as a general loop-nest abstraction; the current
+   implementation supports only 2- or 3-level loop nests.
+
+The interface supports several policy families:
+
+ * ``RAJA::fornest_collapsed_policy<ExecPolicy, LayoutTag>`` maps the product
```
MrBurmark, line 32: What does the `LayoutTag` do? Allow reordering essentially? Does it default to `layout_right`?

MrBurmark: Also why `collapsed` with a d policy and not just `collapse`?

docs/sphinx/user_guide/cook_book/fornest.rst
```diff
@@ -0,0 +1,269 @@
...
+   of the logical dimensions to a 1-D iteration space. The ``ExecPolicy`` is a
+   regular forall-style policy such as ``RAJA::seq_exec`` or
+   ``RAJA::device_exec<256>``. RAJA performs the linear-to-logical index
+   reconstruction internally. On host OpenMP builds, ``RAJA::fornest_omp_collapse_policy<...>``
```
MrBurmark, line 36: The `fornest_omp_collapse_policy` policy is here to allow directly using omp collapse vs doing our own collapse with and using something like `omp_parallel_for` which can be done via `fornest_collapsed_policy<omp_parallel_for, ...>`.

docs/sphinx/user_guide/cook_book/fornest.rst
```diff
@@ -0,0 +1,269 @@
...
+   indices).
+ * ``RAJA::fornest_mapping_policy<ExecPolicy, LoopPolicies...>`` maps the logical
+   dimensions directly using one loop mapping tag per dimension. For mapping
+   policies, the ``ExecPolicy`` selects the launch backend (typically
+   ``RAJA::seq_launch_t`` on host and ``RAJA::device_launch_t<false>`` on GPU),
+   and RAJA derives ``LaunchParams(Teams, Threads)`` from the mapping tags and
+   segment extents. Device builds can use ``RAJA::device_*`` mapping tags; on
+   CUDA/HIP these map to CUDA/HIP global, block, and thread indices, and on SYCL
+   they map to the corresponding work-group and work-item indices where
+   supported.
```
MrBurmark, line 48: My wouldn't it be more consistent to use policies like `RAJA::seq_exec`, `RAJA::device_exec<256>`, etc as the `ExecPolicy`?

docs/sphinx/user_guide/cook_book/fornest.rst
```diff
@@ -0,0 +1,269 @@
...
+ * ``RAJA::fornest_tiling_policy<...>`` adds per-dimension tiling on top of a
+   loop nest. Tile sizes may be fixed, runtime-provided, or chosen by RAJA.
+ * ``RAJA::dynamic_fornest<camp::list<...>>(pol, ...)`` selects a policy from a
+   list at runtime.
```
MrBurmark, line 52:
```suggestion
 * ``RAJA::dynamic_fornest<camp::list<...>>(policy_index, ...)`` selects a policy from a
   list at runtime.
```
Would this make it more clear that `pol` is mean to be an index into the list of policies?

docs/sphinx/user_guide/cook_book/fornest.rst
```diff
@@ -0,0 +1,269 @@
...
+   dimensions directly using one loop mapping tag per dimension. For mapping
+   policies, the ``ExecPolicy`` selects the launch backend (typically
+   ``RAJA::seq_launch_t`` on host and ``RAJA::device_launch_t<false>`` on GPU),
+   and RAJA derives ``LaunchParams(Teams, Threads)`` from the mapping tags and
+   segment extents. Device builds can use ``RAJA::device_*`` mapping tags; on
+   CUDA/HIP these map to CUDA/HIP global, block, and thread indices, and on SYCL
+   they map to the corresponding work-group and work-item indices where
+   supported.
+ * ``RAJA::fornest_tiling_policy<...>`` adds per-dimension tiling on top of a
+   loop nest. Tile sizes may be fixed, runtime-provided, or chosen by RAJA.
```
MrBurmark, line 50:
```suggestion
 * ``RAJA::fornest_tiling_policy<ExecPolicy, ...>`` adds per-dimension tiling on top of a
   loop nest. Tile sizes may be fixed, runtime-provided, or chosen by RAJA.
```
Don't we still want to have an `ExecPolicy` so we can do things like apply launch bounds?

docs/sphinx/user_guide/cook_book/fornest.rst
MrBurmark: Do we still have permutation lists that we could imagine taking as the LayoutTag that could be used to reorder the loops?

include/RAJA/pattern/fornest.hpp
```diff
@@ -0,0 +1,4362 @@
...
+// Copyright (c) Lawrence Livermore National Security, LLC and other
+// RAJA Project Developers. See top-level LICENSE and COPYRIGHT
+// files for dates and other details. No copyright assignment is required
+// to contribute to RAJA.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#ifndef RAJA_pattern_fornest_HPP
+#define RAJA_pattern_fornest_HPP
```
MrBurmark, line 20: I think we want to split this file up into multiple files.

include/RAJA/pattern/fornest.hpp
```diff
@@ -0,0 +1,4362 @@
...
+    RAJA::fornest_platform_loop_policy<RAJA::seq_exec>,
+    RAJA::fornest_platform_loop_policy<RAJA::seq_exec>>;
+#endif
+#endif
+
+namespace type_traits
+{
+
+template<typename Policy>
+struct policy_dimensionality : std::integral_constant<camp::idx_t, 0>
```
MrBurmark, line 283: If we write the policy we can have a static constexpr variable to make it easy to query the number of dimensions instead of writing a separate traits class.

rhornung67:
@artv3 do you want to include this in the RAJA patch release?

artv3:
> @artv3 do you want to include this in the RAJA patch release?

Yes this wold be great

PR #2031, opened 2026-05-21
Update camp submodule gitlink to merge commit in camp

smithsg84:
# CAMP submodule gitlink updated to point to correct commit

The CAMP submodule gitlink was set to the branch commit hash, change to point to the merge commit for the branch.

- This PR is a bugfix, feature, something else)
  - Fixes CAMP gitlink to point to correct commit in main


johnbowen42 (approved):

johnbowen42:
Thanks Steve, good catch!

rhornung67 (approved):

rhornung67:
Folks, please note that all LC networks are currently down. So we will have to wait until the CZ comes back and then probably manually trigger GitLab CI there. It may require pushing an empty commit.

PR #2038, opened 2026-06-11
Add Hubcast config file and update RSCI with hubcast support

adrienbernede:
Add the `.github/hubcast.yml` file to configure hubcast.

Update Radiuss Shared CI with hubcast related tweaks.

Testing changes for RADIUSS Shared CI 2026.6.0 release.

Optimize alloc on matrix

### Note:
May be merged before lc-hubcast is turned on.

### TODO:
- [ ] activate lc-hubcast (optional)
- [x] release radiuss-shared-ci 2026.6.0

### DIFF
--- a/.gitlab/custom-variables.yml
+++ b/.gitlab/custom-variables.yml
@@ -31,7 +31,7 @@ variables:
   # Matrix (SLURM) allocation settings
   MATRIX_SHARED_ALLOC: "OFF"
   # Note: we repeat the reservation, helpful when jobs are manually re-triggered.
-  MATRIX_JOB_ALLOC: "--partition=pci --exclusive=user --time=30 --nodes=1"
+  MATRIX_JOB_ALLOC: "--partition=pci --time=30 --gpus=1"
   # Project specific variants for matrix
   PROJECT_MATRIX_VARIANTS: "~shared +cuda cuda_arch=90 +tests"
   # Project specific deps for matrix
--- a/.github/hubcast.yml
+++ b/.github/hubcast.yml
@@ -1,0 +1,14 @@
+Repo:
+  dest_org: radiuss
+
+  dest_name: raja
+
+  check_name: gitlab-ci
+
+  check_types: [pipeline, child-pipelines]
+
+  delete_closed: true
+
+  sync_drafts: false
+
+  sync_drafts_msg: true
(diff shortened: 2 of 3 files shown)


.gitlab/custom-variables.yml
smithsg84, line 1: On the change of flags... I see LC describes matrix as "gpu scheduled" but document says --nodes=1 allocates 1 GPU and 28 CPU cores.   Is there some difference  between nodes=1 and gpus=1?  I see the difference when > 1.   Just wondering if I'm missing something.

https://hpc.llnl.gov/hardware/compute-platforms/matrix

Thanks

smithsg84 (approved):

.gitlab/custom-variables.yml
adrienbernede: Hmm, interesting... I read the same statement that it was "gpu allocated" associated with the suggestion to use "--gpu".

The way I interpret the docs you mention is that at least we were not wasting ressource before, but "--gpu" removes any confusion.

Also, note that "--exclusive" does result in allocating the full node, so I removed that...

I'll double check all that when I'm back from the conference.

smithsg84: I agree that -gpu matches the suggested use in the document more closely.  I was wondering if there was something else I was missing.  There is a somewhat vague statement "If you do not specify a GPU option, your allocation may not have the resources available that you expect."   Can you get a node without the gpu??

I've been on some systems that have nodes with/without GPU so specifying GPU was required.

Think the change makes it more clear so wasn't objecting to it.

adayton1 (approved):

.gitlab/custom-variables.yml
rhornung67: @smithsg84 if you go onto matrix (or rzvector) and run `lstopo -l` on a login node, you will see that there is no GPU hardware. If you run the same command on a non-login node (e.g., a pdebug node), you will see GPU hardware. You can also see other details of the node compute and memory architecture.

smithsg84: @rhornung67 OK wasn't thinking about login nodes when I wrote that.   The options here are for the scheduler so for the non-login nodes, which I believe are all homogenous with GPUs at least on matrix/rzvector.

rhornung67: Yes, that's true. Sorry I misunderstood.

rhornung67 (approved):

PR #2042, opened 2026-06-16
Fix ambiguous unqualified partition call in util/sort.hpp

siramok:
# Summary

- This PR is a: bugfix
- It does the following:
  - Adds a `detail::` qualifier to the `partition` call within `util/sort.hpp` to avoid a naming conflict with `std::partition`. Without this change, we (over in Conduit) encounter a compilation error when leveraging RAJA for CPU-based sorting, since `std::vector` brings `std::partition` into scope which creates ambiguity.


MrBurmark:
This looks good, I'll work on bringing this in.

PR #2045, opened 2026-06-22
bugfix: Add scan unit test and fix rocprim scan regression

### LINKED ISSUE #1631: Scan Accumulation Type Consistency
Among our different back-end implementations the type used to accumulate the intermediate values of a scan differs.
This result is inconsistent scan results between different back-ends when using input and output containers with different value types.
For example scanning over an array of 16-bit int inputs and storing into a array of 32-bit int outputs could lead to truncation of the accumulated values to 16-bits despite the output being able to store 32-bits.
The behavior should be made to consistently use the output type for the accumulation type.

We default out initial values to the type from the input container.
https://github.com/LLNL/RAJA/blob/develop/include/RAJA/pattern/scan.hpp#L262
We use the type of the output container to accumulate in the sequential backend.
https://github.com/LLNL/RAJA/blob/develop/include/RAJA/policy/sequential/scan.hpp#L146
Cub uses the type of the output container to accumulate for the cuda backend.
https://github.com/dmlc/cub/blob/05eb57faa0a4cac37c2a86fdf4b4dc865a95a1a3/cub/agent/agent_scan.cuh#L110
Rocprim uses the type of the input container to accumulate for the hip backend.
https://github.com/ROCm/rocPRIM/blob/develop/rocprim/include/rocprim/device/device_scan.hpp#L653

johnbowen42:
This bug resolves #1631


artv3 (approved):
nice!

include/RAJA/policy/hip/scan.hpp
```diff
@@ -46,6 +46,13 @@ namespace impl
 namespace scan
 {
 
+template<typename InputIter, typename Function>
+using accumulator_type =
```
MrBurmark: Did we have to do this, does rocprim not use the value type of the output iterator from rocm 7 on?

test/unit/hip/CMakeLists.txt
```diff
@@ -10,3 +10,7 @@
 raja_add_test(
   NAME test-hip-synchronize
   SOURCES test-synchronize.cpp)
+
+raja_add_test(
```
MrBurmark: Why not make a test for all backends?

test/unit/hip/test-scan-char-int.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+}  // namespace
+
+GPU_TEST(ScanUnitTest, HIPCharToIntExclusive)
+{
+  std::vector<char> input(scan_len, 1);
+
+  std::vector<int> seq_output(scan_len, -1);
+  run_scan<RAJA::seq_exec>(input.data(), seq_output.data(), scan_len);
```
MrBurmark: I think I would manually scan the data rather than relying on RAJA to test RAJA.

include/RAJA/policy/hip/scan.hpp
```diff
@@ -71,9 +78,11 @@ RAJA_INLINE resources::EventProxy<resources::Hip> inclusive_inplace(
   void* d_temp_storage      = nullptr;
   size_t temp_storage_bytes = 0;
 #if defined(__HIPCC__)
-  CAMP_HIP_API_INVOKE_AND_CHECK(::rocprim::inclusive_scan, d_temp_storage,
-                                temp_storage_bytes, begin, begin, len,
-                                binary_op, stream);
+  using acc_type = accumulator_type<InputIter, Function>;
+  CAMP_HIP_API_INVOKE_AND_CHECK(
+      (::rocprim::inclusive_scan<::rocprim::default_config, InputIter,
+                                 InputIter, Function, acc_type>),
```
MrBurmark, line 83: Does this compile with rocm 6?

rhornung67: Apparently, since the CI is passing. It doesn't work with rocm 5, which is what we are testing on corona. We need to update that. We should probably move most of our rocm CI check to rocm 7.

johnbowen42: Just so I understand correctly--we should move the rocm 5 testing on corona to a higher version number instead of making this rocm 5 compatible?

rhornung67: Yes.

include/RAJA/policy/hip/scan.hpp
johnbowen42: rocprim 7 computes its internal accumulator type basically using `invoke_result<BinaryFunction, InputType, InputType>::type`, but rocprim 6 uses input type directly unless specified through the final template type parameter

include/RAJA/policy/hip/scan.hpp
```diff
@@ -46,6 +46,13 @@ namespace impl
 namespace scan
 {
 
+template<typename InputIter, typename Function>
+using accumulator_type =
+    std::remove_cv_t<std::remove_reference_t<std::invoke_result_t<
+        Function&,
+        typename std::iterator_traits<InputIter>::value_type,
+        typename std::iterator_traits<InputIter>::value_type>>>;
```
MrBurmark:
```suggestion
    std::remove_cvref_t<std::invoke_result_t<
        Function&,
        typename std::iterator_traits<InputIter>::value_type,
        typename std::iterator_traits<InputIter>::value_type>>;
```

test/unit/algorithm/tests/test-algorithm-scan.hpp
```diff
@@ -0,0 +1,109 @@
...
+  using ExecPolicy = typename camp::at<TypeParam, camp::num<0>>::type;
+  using ResType = typename camp::at<TypeParam, camp::num<1>>::type;
+
+  ResType res = ResType::get_default();
+  runScanCharToIntExclusiveTest<ExecPolicy>(res);
+}
+
+REGISTER_TYPED_TEST_SUITE_P(ScanUnitTest, CharToIntExclusive);
+
+using SequentialScanExecPolicies = camp::list<RAJA::seq_exec>;
```
MrBurmark, line 138: There aren't existing scan policy lists?

test/unit/algorithm/tests/test-algorithm-scan.hpp
```diff
@@ -0,0 +1,109 @@
...
+  char* host_in = host_res.template allocate<char>(scan_len);
+  int* host_out = host_res.template allocate<int>(scan_len);
+
+  std::fill_n(host_in, scan_len, char{1});
+  std::fill_n(host_out, scan_len, -1);
+
+  res.memcpy(work_in, host_in, scan_len * sizeof(char));
+  res.wait();
+
+  RAJA::exclusive_scan<ExecPolicy>(RAJA::make_span(static_cast<const char*>(work_in),
```
MrBurmark: Do we want to add a test for inclusive scan and the inplace versions of these?

test/unit/CMakeLists.txt
```diff
@@ -19,3 +19,7 @@ add_subdirectory(view-layout)
 add_subdirectory(algorithm)
 add_subdirectory(workgroup)
 add_subdirectory(indexing)
+
+if(RAJA_ENABLE_HIP)
```
MrBurmark: does this still exist?

johnbowen42: it does, but was an artifact from the original HIP port of RAJA (as discussed in offline meeting).  I am now deleting this subdirectory

test/unit/algorithm/tests/test-algorithm-scan.hpp
johnbowen42: there were similar policies in algorithm/unit for sort, but they rely on the `PolicySort` testing class and aren't generic.

test/unit/algorithm/tests/test-algorithm-scan.hpp
johnbowen42: I added inclusive scan.  I'm not sure in-place is relevant for this specific regression--we're specifically testing the case where the input/output storage types vary

MrBurmark (approved):

rhornung67 (approved):

rhornung67:
Thanks @johnbowen42

adayton1 (approved):

rhornung67:
@johnbowen42 will you look into why the corona ROCm 6.4.x checks are failing? It looks like Spack cannot satisfy dependencies.

johnbowen42:
> @johnbowen42 will you look into why the corona ROCm 6.4.x checks are failing? It looks like Spack cannot satisfy dependencies.

Working on it

johnbowen42:
@adrienbernede there's still an expected check for `ci/gitlab/lc.llnl.gov` that is blocking the merge of this PR--any idea what that is from?  Maybe leftover from before our hubcast conversion?

adrienbernede:
@johnbowen42 OK fixed !

PR #2047, opened 2026-06-23
bugfix: Use memcpy on device

bmhan12:
This PR:
- Fixes a downstream error due to `std::memcpy` instead of `memcpy` on device.

Error message:
```
[100%] Building HIP object axom/spin/tests/CMakeFiles/spin_uniform_grid_test.dir/spin_uniform_grid.cpp.o
In file included from /usr/WS1/han12/axom_codex/src/axom/spin/tests/spin_uniform_grid.cpp:14:
In file included from /usr/WS1/han12/axom_codex/src/axom/core/../../axom/spin/UniformGrid.hpp:11:
In file included from /usr/WS1/han12/axom_codex/src/axom/core/../../axom/core/execution/for_all.hpp:11:
In file included from /usr/WS1/han12/axom_codex/src/axom/core/../../axom/core/execution/execution_space.hpp:111:
In file included from /usr/WS1/han12/axom_codex/src/axom/core/../../axom/core/execution/internal/seq_exec.hpp:15:
In file included from /usr/WS1/han12/axom_codex/uberenv_libs/llvm-amdgpu-6.4.3/raja-develop-hbu2lep7bacs724pnfnxefjefkq64cdd/include/RAJA/RAJA.hpp:77:
In file included from /usr/WS1/han12/axom_codex/uberenv_libs/llvm-amdgpu-6.4.3/raja-develop-hbu2lep7bacs724pnfnxefjefkq64cdd/include/RAJA/policy/hip.hpp:32:
In file included from /usr/WS1/han12/axom_codex/uberenv_libs/llvm-amdgpu-6.4.3/raja-develop-hbu2lep7bacs724pnfnxefjefkq64cdd/include/RAJA/policy/hip/atomic.hpp:37:
In file included from /usr/WS1/han12/axom_codex/uberenv_libs/llvm-amdgpu-6.4.3/raja-develop-hbu2lep7bacs724pnfnxefjefkq64cdd/include/RAJA/policy/atomic_builtin.hpp:33:
/usr/WS1/han12/axom_codex/uberenv_libs/llvm-amdgpu-6.4.3/raja-develop-hbu2lep7bacs724pnfnxefjefkq64cdd/include/RAJA/util/TypeConvert.hpp:56:8: error: reference to __host__ function 'memcpy' in __host__ __device__ function
   56 |   std::memcpy(&b, &a, sizeof(A));
      |        ^
```

PR #2048, opened 2026-06-23
Use memcpy for device

rhornung67:
# Summary

- This PR is a companion PR of https://github.com/llnl/RAJA/pull/2047
- It is a bugfix.

See the original PR for explanation. Please review and approve this PR for merging.

Thanks for the fix @bmhan12

### DIFF
--- a/include/RAJA/util/TypeConvert.hpp
+++ b/include/RAJA/util/TypeConvert.hpp
@@ -53,7 +53,7 @@ RAJA_INLINE RAJA_HOST_DEVICE constexpr B reinterp_A_as_B(A const& a)
   // TODO: Consider requiring A and B to be trivially copyable
 
   B b;
-  std::memcpy(&b, &a, sizeof(A));
+  memcpy(&b, &a, sizeof(A));
   return b;
 }
 


include/RAJA/util/TypeConvert.hpp
adayton1, line 34:
```suggestion
#include <string.h>
```

adayton1: I don't think this suggestion turned out quite right. I meant to have string.h replace cstring since memcpy is the c function really from string.h. cstring probably includes string.h anyway, though. Is there a preference?

artv3: yes, I think  <cstring> for C++ code

artv3: I think I like what was there before with the std:: namespace

adayton1: It sounds like std::memcpy is broken in device code (I recently changed it from memcpy to std::memcpy in https://github.com/llnl/RAJA/pull/2005/changes and our testing didn't show any issues, but that probably just means our testing isn't thorough enough. If we have to use memcpy, should we include string.h instead?

MrBurmark: It seems that in practice the c++ version of the header makes the std namespace version and the global namespace version available, but you're only guaranteed to get the std namespace version with the c++ header. Similarly you're only guaranteed to get global namespace version with the c version of the header.

adayton1 (approved):

include/RAJA/util/TypeConvert.hpp
rhornung67: Also, I believe the C++ std header convention is to wrap the related C headers and qualifies them with `std::`; e.g., cstring includes string.h, etc.

However, the C++ standard rules govern how things are handled under the covers.

rhornung67 (commented):
I approve, but I can't since I made this PR.

MrBurmark (approved):

PR #2052, opened 2026-07-02
feature: Proteus lambda redesign RAJA-side support

johnbowen42:
We recently redesigned our lambda JIT compilation pipeline in Proteus (https://github.com/Olympus-HPC/proteus/pull/467).  As part of this work, the results of `proteus::register_lambda` is `nodiscard`, and must be passed to the kernel launch.  This makes it possible for Proteus's compile-time LLVM plugin to analyze the callsite and determine which data it needs. 

This PR removes `register_lambda` from RAJA abstractions, forcing users to opt-in and decide if they wish to JIT compile a lambda or not.  This PR also adds some integration tests which verify that lambda are specialized through RAJA abstractions, and a nice example with RAJA kernel.

### DIFF
--- a/examples/kernel-jit.cpp
+++ b/examples/kernel-jit.cpp
@@ -1,0 +1,152 @@
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+// Copyright (c) Lawrence Livermore National Security, LLC and other
+// RAJA Project Developers. See top-level LICENSE and COPYRIGHT
+// files for dates and other details. No copyright assignment is required
+// to contribute to RAJA.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+// This example mirrors forall-jit.cpp, but uses RAJA::kernel to drive the
+// outer parallel loop.
+
+#include <cstdlib>
+#include <iostream>
+
+#include "RAJA/RAJA.hpp"
+#include "RAJA/util/Timer.hpp"
+
+#if defined(RAJA_ENABLE_CUDA)
+using kernel_policy = RAJA::KernelPolicy<
+    RAJA::statement::CudaKernelFixed<256,
+        RAJA::statement::Tile<0, RAJA::tile_fixed<256>,
+                              RAJA::cuda_block_x_direct,
+            RAJA::statement::For<0, RAJA::cuda_thread_x_direct,
+                RAJA::statement::Lambda<0>>>>>;
+#elif defined(RAJA_ENABLE_HIP)
+using kernel_policy = RAJA::KernelPolicy<
+    RAJA::statement::HipKernelFixed<256,
+        RAJA::statement::Tile<0, RAJA::tile_fixed<256>,
+                              RAJA::hip_block_x_direct,
+            RAJA::statement::For<0, RAJA::hip_thread_x_direct,
+                RAJA::statement::Lambda<0>>>>>;
+#elif defined(RAJA_ENABLE_OPENMP)
+using kernel_policy = RAJA::KernelPolicy<
+    RAJA::statement::For<0, RAJA::omp_parallel_for_exec,
+        RAJA::statement::Lambda<0>>>;
+#else
+using kernel_policy = RAJA::KernelPolicy<
+    RAJA::statement::For<0, RAJA::seq_exec, RAJA::statement::Lambda<0>>>;
+#endif
+
+int main(int argc, char **argv) {
+  if (argc < 5) {
+    std::cout
+        << "Expected ./bin/kernel-jit <num rows> <num cols> <num matrices> "
+           "<boolean accum>, for example ./bin/kernel-jit 24 16 1000000 1\n";
+    return 1;
+  }
+
+  int a = std::stoi(argv[1]);
+  int b = std::stoi(argv[2]);
+  int N = std::stoi(argv[3]);
+  bool accum = std::stoi(argv[4]);
+
+#ifdef RAJA_ENABLE_CUDA
+  using resource_policy = RAJA::cuda_exec<256>;
+#elif defined(RAJA_ENABLE_HIP)
+  using resource_policy = RAJA::hip_exec<256>;
+#elif defined(RAJA_ENABLE_OPENMP)
+  using resource_policy = RAJA::omp_parallel_for_exec;
+#else
+  using resource_policy = RAJA::seq_exec;
+#endif
+
+  auto res = RAJA::resources::get_default_resource<resource_policy>();
+  double *A_ptr = res.template allocate<double>(N * a * b);
+  auto A = RAJA::make_permuted_view<RAJA::layout_right>(A_ptr, N, a, b);
+  double *B_ptr = res.template allocate<double>(N * a * b);
+  auto B = RAJA::make_permuted_view<RAJA::layout_right>(B_ptr, N, b, a);
+  double *C_ptr = res.template allocate<double>(N * a * a);
+  auto C = RAJA::make_permuted_view<RAJA::layout_right>(C_ptr, N, a, a);
+
+  auto Seg = RAJA::RangeSegment(0, N);
+
+  RAJA::kernel<kernel_policy>(RAJA::make_tuple(Seg), [=](int i) {
+    for (int row = 0; row < a; ++row) {
+      for (int col = 0; col < b; ++col) {
+        A(i, row, col) = 0;
+        B(i, row, col) = 0;
+        C(i, row, col) = 0;
+      }
+    }
+  });
+
+  RAJA::Timer aot_timer;
+  proteus::disable();
+  aot_timer.start();
+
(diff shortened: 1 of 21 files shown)


examples/kernel-jit.cpp
```diff
@@ -0,0 +1,152 @@
...
+  double *A_ptr = res.template allocate<double>(N * a * b);
+  auto A = RAJA::make_permuted_view<RAJA::layout_right>(A_ptr, N, a, b);
+  double *B_ptr = res.template allocate<double>(N * a * b);
+  auto B = RAJA::make_permuted_view<RAJA::layout_right>(B_ptr, N, b, a);
+  double *C_ptr = res.template allocate<double>(N * a * a);
+  auto C = RAJA::make_permuted_view<RAJA::layout_right>(C_ptr, N, a, a);
+
+  auto Seg = RAJA::RangeSegment(0, N);
+
+  RAJA::kernel<kernel_policy>(RAJA::make_tuple(Seg), [=](int i) {
```
MrBurmark, line 74: host device?

include/RAJA/pattern/kernel/internal/LoopData.hpp
```diff
@@ -178,6 +179,50 @@ struct LoopData
...
+                    body.get_resource(),
+                    camp::get<Is>(registered_bodies)...);
+  }
+
+  template<typename SegmentTuple,
+           typename ParamTuple,
+           typename Resource,
+           typename... Bodies>
+  inline auto register_loop_bodies(
+      LoopData<SegmentTuple, ParamTuple, Resource, Bodies...> body)
```
MrBurmark, line 213: reference?

include/RAJA/pattern/kernel/internal/LoopData.hpp
```diff
@@ -178,6 +179,50 @@ struct LoopData
...
+                    body.param_tuple,
+                    body.get_resource(),
+                    camp::get<Is>(registered_bodies)...);
+  }
+
+  template<typename SegmentTuple,
+           typename ParamTuple,
+           typename Resource,
+           typename... Bodies>
+  inline auto register_loop_bodies(
```
MrBurmark, line 212: Do we want this to always return by value or return by reference without jit enabled?

include/RAJA/policy/cuda/kernel/CudaKernel.hpp
```diff
@@ -682,15 +681,20 @@ struct StatementExecutor<
         // of the launch_dims and potential changes to shmem here that is
         // currently an unresolved issue.
         //
+        auto registered_bodies = RAJA::internal::jit::register_loop_bodies(data);
+        using registered_data_t = std::decay_t<decltype(registered_bodies)>;
+        using registered_launch_t =
+            CudaLaunchHelper<LaunchConfig, stmt_list_t, registered_data_t, Types>;
+        auto registered_function = registered_launch_t::get_func();
```
MrBurmark, line 688: This is a lot more code than it was.

johnbowen42: yeah, I am debating simplifying this but shifitng the decision to JIT the lambda to the user.  Basically, if a user wanted to JIT their lambda they could wrap it in `RAJA::register_lambda` before passing to kernel/launch/forall, and then we could get rid of all this boilerplate

johnbowen42: What do you think?

include/RAJA/pattern/kernel/internal/LoopData.hpp
johnbowen42: I deleted all these in favor of moving towards user-side lambda registration.  See the examples

include/RAJA/pattern/kernel/internal/LoopData.hpp
johnbowen42: I removed all of these because I want JIT usage to be user opt-in on a per-lambda basis.  I think we could also look at making jit variants of all the RAJA constructs as a follow up to this.  This decision is partially my own preference, but also based on code requirements

artv3:
For projects that may want to pull in an external version of Proteus, can we add a version check if we are not backwards compatible?

johnbowen42:
> For projects that may want to pull in an external version of Proteus, can we add a version check if we are not backwards compatible?

These changes are backwards compatible with all versions of Proteus

examples/launch-jit.cpp
```diff
@@ -0,0 +1,186 @@
...
+  aot_timer.stop();
+  std::cout << "aot total time = " << aot_timer.elapsed() << "\n";
+  proteus::enable();
+
+  RAJA::Timer jit_timer;
+  jit_timer.start();
+
+  RAJA::launch<launch_policy>(
+      Params,
+      [=, a = RAJA_JIT_VARIABLE(a), b = RAJA_JIT_VARIABLE(b),
```
MrBurmark: Should this lambda be registered?

test/integration/forall_jit_proteus.cpp
```diff
@@ -0,0 +1,74 @@
...
+        C(i, row, col) = 0.0;
+      }
+    }
+  });
+
+  proteus::enable();
+  RAJA::forall<policy>(
+      RAJA::RangeSegment(0, N),
+      RAJA_JIT_REGISTER_LAMBDA([=, a = RAJA_JIT_VARIABLE(a), b = RAJA_JIT_VARIABLE(b),
+       accum = RAJA_JIT_VARIABLE(accum)](int i) RAJA_JIT_COMPILE {
```
MrBurmark: It seems a bit redundant to have both RAJA_JIT_REGISTER_LAMBDA and RAJA_JIT_COMPILE

test/integration/kernel_jit_proteus.cpp
```diff
@@ -0,0 +1,63 @@
...
+  int host_result = -1;
+
+  res.memcpy(out_ptr, &zero, sizeof(int));
+  res.wait();
+
+  proteus::enable();
+  RAJA::kernel<kernel_policy>(
+      RAJA::make_tuple(RAJA::RangeSegment(0, 1)),
+      RAJA_JIT_REGISTER_LAMBDA([=, a = RAJA_JIT_VARIABLE(a), b = RAJA_JIT_VARIABLE(b),
+       accum = RAJA_JIT_VARIABLE(accum)](int i) RAJA_JIT_COMPILE RAJA_HOST_DEVICE {
```
MrBurmark: Is this the normal order for the attributes?

test/integration/launch_jit_proteus.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+  res.memcpy(out_ptr, &zero, sizeof(int));
+  res.wait();
+
+  proteus::enable();
+  RAJA::launch<launch_policy>(
+      RAJA::LaunchParams(RAJA::Teams(1), RAJA::Threads(1)),
+      RAJA_JIT_REGISTER_LAMBDA([=, a = RAJA_JIT_VARIABLE(a), b = RAJA_JIT_VARIABLE(b),
+       accum = RAJA_JIT_VARIABLE(accum)]
+      RAJA_JIT_COMPILE RAJA_HOST_DEVICE (RAJA::LaunchContext RAJA_UNUSED_ARG(ctx))RAJA_JIT_COMPILE
```
MrBurmark: Two RAJA_JIT_COMPILE s

test/integration/forall_jit_proteus.cpp
johnbowen42: I agree, unfortunately its currently a requirement for CPU compilation.  I am to eliminate the `RAJA_JIT_COMPILE` annotation requirement for registered lambdas in a feature branch soon

johnbowen42: https://github.com/Olympus-HPC/proteus/issues/529

examples/launch-jit.cpp
johnbowen42: yes--I missed this one

test/integration/kernel_jit_proteus.cpp
johnbowen42: I fixed this syntax so that the ordering it correct--we want RAJA_JIT_COMPILE to follow the parameter list

include/RAJA/util/Jit.hpp
```diff
@@ -8,13 +9,24 @@
...
 {
-namespace internal
+#if defined(RAJA_ENABLE_JIT)
+namespace detail
 {
+
+template<std::uint64_t FunctorID, typename Lambda>
+struct launch_context_type<
+    proteus::detail::LambdaFunctorWrapper<FunctorID, Lambda>>
+    : launch_context_type<Lambda>
```
MrBurmark, line 19: Is this right?

johnbowen42: We couldn't get the launch context type without this because Proteus wraps a registered lambda with a functor. Before launch would look at the lambda's argument type--but the functor has a generic `operator()(Args...)`.  This redirects the helper to the argument type of the underlying lambda

johnbowen42: @MrBurmark is that ok?

MrBurmark: Ya, that's fine. I was confused by the way it was partially specializing things as I didn't see the original declaration.

MrBurmark (approved):

johnbowen42:
I just verified again that this branch can JIT compile all MFEM kernels when MFEM uses the RAJA backend

PR #2054, opened 2026-07-09
Fixes CUDA generic atomic operation by fixing atomic load

adayton1:
# Summary

- This PR is a bugfix from @bechols97 
- It does the following:
  - Uses cuda::atomic_ref::load for atomicLoad instead of atomicOr with a value of 0
  - Uses cuda::atomic_ref::store for atomicStore instead of atomicExchange
  - The old implementation of atomicLoad in CUDA was giving incorrect results in #1832
  - Performance should be equivalent or better using atomic_ref, but even if not, correctness matters most
  - The RAJA_ENABLE_CUDA_ATOMIC_REF was never getting defined as intended, but now that we require CUDA 12, cuda::atomic_ref should always be available. The cuda/atomic header gives a very clear error message if an architecture older than sm_60 is used.

### DIFF
--- a/include/RAJA/policy/cuda/atomic.hpp
+++ b/include/RAJA/policy/cuda/atomic.hpp
@@ -28,11 +28,6 @@
 #include <type_traits>
 #include <utility>
 
-#if __CUDA__ARCH__ >= 600 && __CUDACC_VER_MAJOR__ >= 11 &&                     \
-    __CUDACC_VER_MINOR__ >= 6
-#define RAJA_ENABLE_CUDA_ATOMIC_REF
-#endif
-
 #include <cuda/atomic>
 
 #include "camp/list.hpp"
@@ -179,10 +174,8 @@ RAJA_INLINE __device__ T cuda_atomicExchange(T* acc, T value)
 }
 
 /*!
- * Atomic load and store
+ * Atomic load
  */
-#if defined(RAJA_ENABLE_CUDA_ATOMIC_REF)
-
 template<typename T>
 RAJA_INLINE __device__ T cuda_atomicLoad(T* acc)
 {
@@ -190,40 +183,15 @@ RAJA_INLINE __device__ T cuda_atomicLoad(T* acc)
       cuda::memory_order_relaxed {});
 }
 
+/*!
+ * Atomic store
+ */
 template<typename T>
 RAJA_INLINE __device__ void cuda_atomicStore(T* acc, T value)
 {
   cuda::atomic_ref<T, cuda::thread_scope_device>(*acc).store(
       value, cuda::memory_order_relaxed {});
 }
-
-#else
-
-template<typename T,
-         std::enable_if_t<cuda_useBuiltinCommon<T>::value, bool> = true>
-RAJA_INLINE __device__ T cuda_atomicLoad(T* acc)
-{
-  return cuda_atomicOr(acc, static_cast<T>(0));
-}
-
-template<typename T,
-         std::enable_if_t<cuda_useReinterpretCommon<T>::value, bool> = true>
-RAJA_INLINE __device__ T cuda_atomicLoad(T* acc)
-{
-  using R = cuda_useReinterpretCommon_t<T>;
-
-  return RAJA::util::reinterp_A_as_B<R, T>(
-      cuda_atomicLoad(reinterpret_cast<R*>(acc)));
-}
-
-template<typename T>
-RAJA_INLINE __device__ void cuda_atomicStore(T* acc, T value)
-{
-  cuda_atomicExchange(acc, value);
-}
-
-#endif
-
 
 /*!
  * Atomic compare and swap
--- a/include/RAJA/policy/cuda/atomic.hpp
+++ b/include/RAJA/policy/cuda/atomic.hpp
@@ -179,8 +179,8 @@ RAJA_INLINE __device__ T cuda_atomicExchange(T* acc, T value)
 template<typename T>
 RAJA_INLINE __device__ T cuda_atomicLoad(T* acc)
 {
-  return cuda::atomic_ref<T, cuda::thread_scope_device>(*acc).load(
-      cuda::memory_order_relaxed {});
+  return ::cuda::atomic_ref<T, ::cuda::thread_scope_device>(*acc).load(
+      ::cuda::memory_order_relaxed {});
 }
 
 /*!
@@ -189,8 +189,8 @@ RAJA_INLINE __device__ T cuda_atomicLoad(T* acc)
 template<typename T>
 RAJA_INLINE __device__ void cuda_atomicStore(T* acc, T value)
 {
-  cuda::atomic_ref<T, cuda::thread_scope_device>(*acc).store(
-      value, cuda::memory_order_relaxed {});
+  ::cuda::atomic_ref<T, ::cuda::thread_scope_device>(*acc).store(
+      value, ::cuda::memory_order_relaxed {});
 }
 
 /*!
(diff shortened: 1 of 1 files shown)


smithsg84:
Is there a reason for changing just the implementation for atomicGeneric and not using the refactored atomic_ref implementation in cuda_atomicCAS_loop?

adayton1:
> Is there a reason for changing just the implementation for atomicGeneric and not using the refactored atomic_ref implementation in cuda_atomicCAS_loop?

cuda_atomicCAS_loop has been around for much longer than atomicGeneric and is used by other operations such as atomicMin and atomicMax without any noticeable problems. This new implementation has not been thoroughly tested or benchmarked, so I'm hesitant to make changes to cuda_atomicCAS_loop directly.

artv3:
Regarding ` old implementation of atomicGeneric in CUDA is giving incorrect results ` is that a CUDA related bug? or why is that happening?

artv3 (approved):
LGTM, just to double check -- we have tests for this right?

include/RAJA/policy/cuda/atomic.hpp
```diff
@@ -844,7 +842,17 @@ RAJA_INLINE RAJA_HOST_DEVICE T atomicGeneric(cuda_atomic_explicit<host_policy>,
                                              Operation&& operation)
 {
 #ifdef __CUDA_ARCH__
-  return detail::cuda_atomicCAS_loop(acc, std::forward<Operation>(operation));
+  ::cuda::atomic_ref<T, ::cuda::thread_scope_device> ref(*acc);
+  T expected = ref.load(::cuda::std::memory_order_relaxed);
+  while (true)
+  {
+    if (ref.compare_exchange_weak(expected, operation(expected),
```
MrBurmark: Short circuit?

adayton1:
> LGTM, just to double check -- we have tests for this right?

Yes, there are tests.

adayton1:
> Regarding `old implementation of atomicGeneric in CUDA is giving incorrect results` is that a CUDA related bug? or why is that happening?

I don't understand the cause yet. Though the old implementation was using atomicOr(acc, 0) to do the load, so maybe there was something weird happening with that?

include/RAJA/policy/cuda/atomic.hpp
adayton1: I wasn't planning on supporting short-circuiting in this overload of atomicGeneric. I had considered adding another overload at some other point.

rhornung67 (approved):

MrBurmark (approved):

smithsg84 (approved):

artv3:
> > Regarding `old implementation of atomicGeneric in CUDA is giving incorrect results` is that a CUDA related bug? or why is that happening?
> 
> I don't understand the cause yet. Though the old implementation was using atomicOr(acc, 0) to do the load, so maybe there was something weird happening with that?

Maybe good to capture in an issue if we don't have one already

adayton1:
> > Regarding `old implementation of atomicGeneric in CUDA is giving incorrect results` is that a CUDA related bug? or why is that happening?
> 
> I don't understand the cause yet. Though the old implementation was using atomicOr(acc, 0) to do the load, so maybe there was something weird happening with that?

@bechols97 did some more testing and it appears that using cuda::atomic_ref::load in place of the atomicOr for the load fixed the issues. I'm guessing atomic_ref will be faster than the atomicOr, but even if not, it seems like for correctness we should be using atomic_ref for the load.

include/RAJA/policy/cuda/atomic.hpp
```diff
@@ -181,52 +174,25 @@ RAJA_INLINE __device__ T cuda_atomicExchange(T* acc, T value)
...
       value, cuda::memory_order_relaxed {});
 }
 
-#else
-
-template<typename T,
-         std::enable_if_t<cuda_useBuiltinCommon<T>::value, bool> = true>
-RAJA_INLINE __device__ T cuda_atomicLoad(T* acc)
-{
-  return cuda_atomicOr(acc, static_cast<T>(0));
```
adayton1, line 208: Using cuda::atomic_ref instead of this fixed the bug @bechols97 was seeing.

include/RAJA/policy/cuda/atomic.hpp
```diff
@@ -844,7 +810,17 @@ RAJA_INLINE RAJA_HOST_DEVICE T atomicGeneric(cuda_atomic_explicit<host_policy>,
                                              Operation&& operation)
 {
 #ifdef __CUDA_ARCH__
-  return detail::cuda_atomicCAS_loop(acc, std::forward<Operation>(operation));
+  ::cuda::atomic_ref<T, ::cuda::thread_scope_device> ref(*acc);
```
bechols97: @adayton1 Since the root cause of the issue was more the `atomicLoad` operation, do we want to revert this to use `cuda_atomicCAS_loop`?

include/RAJA/policy/cuda/atomic.hpp
```diff
@@ -181,52 +174,25 @@ RAJA_INLINE __device__ T cuda_atomicExchange(T* acc, T value)
...
 }
 
+/*!
+ * Atomic store
+ */
 template<typename T>
 RAJA_INLINE __device__ void cuda_atomicStore(T* acc, T value)
 {
   cuda::atomic_ref<T, cuda::thread_scope_device>(*acc).store(
       value, cuda::memory_order_relaxed {});
```
bechols97: To avoid having the compiler get confused with `RAJA::cuda` and `cuda` namespace, I had to prefix the cuda namespace with `::` in my testing.

```suggestion
::cuda::atomic_ref<T, ::cuda::thread_scope_device>(*acc).store(
value, ::cuda::memory_order_relaxed {});
```

include/RAJA/policy/cuda/atomic.hpp
```diff
@@ -181,52 +174,25 @@ RAJA_INLINE __device__ T cuda_atomicExchange(T* acc, T value)
...
- * Atomic load and store
+ * Atomic load
  */
-#if defined(RAJA_ENABLE_CUDA_ATOMIC_REF)
-
 template<typename T>
 RAJA_INLINE __device__ T cuda_atomicLoad(T* acc)
 {
   return cuda::atomic_ref<T, cuda::thread_scope_device>(*acc).load(
       cuda::memory_order_relaxed {});
```
bechols97: Same as the other suggestion

```suggestion
return ::cuda::atomic_ref<T, ::cuda::thread_scope_device>(*acc).load(
::cuda::memory_order_relaxed {});
```

include/RAJA/policy/cuda/atomic.hpp
adayton1: Done.

bechols97 (approved):

include/RAJA/policy/cuda/atomic.hpp
adayton1: Fixed.

include/RAJA/policy/cuda/atomic.hpp
adayton1: Fixed.

include/RAJA/policy/cuda/atomic.hpp
MrBurmark: Interesting, I'm pretty sure I added this and it was passing our tests before.

MrBurmark (approved):

include/RAJA/policy/cuda/atomic.hpp
adayton1: Yeah, I'm pretty sure we have testing for it, though not tests that use the result in other atomic operations.

PR #2055, opened 2026-07-10
Add generic atomic operation with optional early exit

adayton1:
# Summary

- This PR is a feature
- It does the following:
  - Adds a generic atomic operation taking a predicate used to determine if we can exit early from the CAS loop
  - This implementation already underpins some of our fallback implementations for min/max atomic operations for cuda and hip

### DIFF
--- a/include/RAJA/pattern/atomic.hpp
+++ b/include/RAJA/pattern/atomic.hpp
@@ -317,8 +317,7 @@ RAJA_SUPPRESS_HD_WARN
 template<typename Policy,
          typename T,
          typename Operation,
-         typename StopPredicate>
-  requires std::predicate<StopPredicate, T>
+         std::predicate<T> StopPredicate>
 RAJA_INLINE RAJA_HOST_DEVICE T atomicGeneric(T* acc,
                                              Operation&& operation,
                                              StopPredicate&& stop)
--- a/include/RAJA/pattern/atomic.hpp
+++ b/include/RAJA/pattern/atomic.hpp
@@ -22,6 +22,7 @@
 
 #include "RAJA/config.hpp"
 
+#include <concepts>
 #include <utility>
 
 #include "RAJA/policy/atomic_auto.hpp"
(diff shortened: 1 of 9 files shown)


include/RAJA/pattern/atomic.hpp
```diff
@@ -303,6 +304,29 @@ RAJA_INLINE RAJA_HOST_DEVICE T atomicGeneric(T* acc, Operation&& operation)
...
+ *                  from the current value.
+ * @param stop Predicate that returns true when no further update is needed.
+ * @return The old value observed before the successful update or early exit.
+ */
+RAJA_SUPPRESS_HD_WARN
+template<typename Policy,
+         typename T,
+         typename Operation,
+         typename StopPredicate>
+  requires std::predicate<StopPredicate, T>
```
MrBurmark: Would this be a nicer way to write this?
```suggestion
         std::predicate<T> StopPredicate>
```

MrBurmark (approved):

include/RAJA/pattern/atomic.hpp
adayton1: I'm not sure that's correct. std::predicate expects both the callable and the arguments. But I'm certainly not an expert on how to specify concepts/constraints.

bechols97 (approved):

include/RAJA/pattern/atomic.hpp
MrBurmark: When used in this way the first argument of the concept is the type of the current template parameter (kind of like `this` is implicitly the first argument of member functions). This allows using concepts without a requires clause. If the concept needs more than 1 argument the later arguments still have to be provided of course the first is still implicitly filled.
```cpp
template < std::integral T >
T foo(T);
```

adayton1: I gave it a try.

include/RAJA/policy/desul/atomic.hpp
```diff
@@ -179,6 +180,37 @@ RAJA_HOST_DEVICE RAJA_INLINE T atomicGeneric(AtomicPolicy,
...
 
+RAJA_SUPPRESS_HD_WARN
+template<typename AtomicPolicy,
+         typename T,
+         typename Operation,
+         std::predicate<T> StopPredicate>
+RAJA_HOST_DEVICE RAJA_INLINE T
+atomicGeneric(AtomicPolicy, T* acc, Operation&& operation, StopPredicate&& stop)
+{
+  static_assert(std::is_trivially_copyable_v<T>);
```
MrBurmark: Should this be a concept applied to T?

adayton1: I'm not actually sure if it is necessary or if necessary if it is sufficient. Instead of adding specific type traits just for desul, they should probably be consistent across all the overloads. I think for now I'll remove this static_assert.

MrBurmark (approved):

smithsg84 (approved):

Issue #1756, opened 2024-10-22
Runtime memory policy for multireducers


tomstitt:
**Is your feature request related to a problem? Please describe.**

The `RAJA::hip_multi_reduce_atomic` and `RAJA::cuda_multi_reduce_atomic` multireducers allocate GPU memory even when they are only used in a CPU kernel. Our application supports a runtime compute policy with GPU builds such that we can run CPU-only if desired; with multireducers this breaks because in CPU mode we now allocate GPU memory.

**Describe the solution you'd like**

We would like the GPU multireducers to dynamically choose their allocator based on the the kernel they are captured by, similar to the regular GPU RAJA reducers like `RAJA::hip_reduce` and `RAJA::cuda_reduce`.

**Describe alternatives you've considered**

We've considered templating routines where we use multireducers with a sequential dispatch for the CPU and a platform-dependent dispatch for the GPU but that requires additional boilerplate. 

**Additional context**

n/a

artv3:
@MrBurmark , would you know the status here?

MrBurmark:
Ya, this is something on my plate, but that I have not started. It will probably involve changing how loop abstrations are implemented for policies like sequential and omp in order to allow for the hip/cuda multi-reducers to allocate at the loop instead of upfront. So I have put this off multiple times due to the invasiveness of the solution. This is not a problem with the new reductions interface so it may be less intrusive to finalize that abstraction instead and have people use that.

rhornung67:
@MrBurmark can this issue be closed? I believe you addressed it in a PR before the recent release.

PR #2059, opened 2026-07-16
Draft: Runtime support for Multi-Reduce  

artv3:
# Summary

@MrBurmark @tomstitt  -- working with agents -- in support of issue https://github.com/llnl/RAJA/issues/1756 what if we consider using the RAJA::resource to choose the backing of the memory for the multi-reducers?

I acknowledge that the request asks for "dynamically choose their allocator based on the kernel they are captured by, similar to regular GPU RAJA reducer" but perhaps using a resource could work too? 

  - Added runtime-resource-aware MultiReduce constructors so multi-reducers can be initialized with the same runtime-selected resource used for execution/input allocation.
  - Updated CUDA/HIP multi-reduce tally allocation to choose host-accessible or device-accessible storage based on the resource platform.
  - Added a runtime multi-reduction example using dynamic_forall for host/device selection.

  Device Buffer Allocation Behavior

  - Previously, CUDA/HIP multi-reducers always allocated tally buffers from the device-accessible pinned/device mempool, regardless of runtime execution path.
  - Now, CUDA/HIP multi-reducers constructed with a runtime resource choose allocation backend from res.get_platform():
      - device resource: existing device-accessible tally mempool, matching previous device behavior
      - host resource: generic host mempool, avoiding device-style tally allocation for host runtime execution

  - The reducer does not allocate directly through res.allocate(); the resource is used to select the appropriate existing tally buffer backend.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -38,6 +38,10 @@ raja_add_executable(
   SOURCES forall_multi-reductions.cpp)
 
 raja_add_executable(
+  NAME runtime-multi-reductions
+  SOURCES runtime-multi-reductions.cpp)
+
+raja_add_executable(
   NAME launch-param-reductions
   SOURCES launch-param-reductions.cpp)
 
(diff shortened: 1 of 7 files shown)


tomstitt:
from an app perspective this looks great! thanks @artv3 !

MrBurmark:
Instead of passing in a whole resource would it be sufficient to pass in a platform?

MrBurmark:
@artv3 @tomstitt With platform we can differentiate between host and device, but can't differentiate between seq and omp on the host. That distinction could let us use less memory for seq policies as we use a per object copy with omp to get a per thread copy. Does that distinction seem relevant to you?

PR #2061, opened 2026-07-16
Give Runtime Policy Information to Reducers

MrBurmark:
# Summary

Add Policy enum to constructor/reset of reducers/multi-reducers to allow them to optimize their resource usage based on the backend they will actually be used with. Constructing/resetting a reducer/multi-reducer to a different policy than the reducer/multi-reducer is subsequently used with is undefined behavior. This is based on the idea from #2059 that was aimed at fixing #1756.

```cpp
//before
RAJA::ReduceSum<RAJA::cuda_reduce, double> sum(0.0);
RAJA::forall<RAJA::seq_exec>(RAJA::Range(0, N) [=](int i) { sum += i; });

// after
RAJA::ReduceSum<RAJA::cuda_reduce, double> sum(RAJA::Policy::sequential, 0.0);
RAJA::forall<RAJA::seq_exec>(RAJA::Range(0, N) [=](int i) { sum += i; });

// alternatively
RAJA::ReduceSum<RAJA::cuda_reduce, double> sum(RAJA::policy_of<RAJA::seq_exec>::value, 0.0);
RAJA::forall<RAJA::seq_exec>(RAJA::Range(0, N) [=](int i) { sum += i; });
```

TODO:
- [x] Support seq, omp, cuda, hip Reduction objects
- [x] Support omp_target, sycl Reduction objects
- [x] Support sequential exec policies with omp_target and sycl reduction objects
- [x] Test reduction correctness with other supported backends (cuda/hip supports seq, openmp, and cuda/hip)
- [x] Document these changes

## Breaking change
Some reducers like sequential and openmp when not given a default value would default construct their value, so min reductions would start with a 0 instead of the max of that type. This was changed so reducers now consistently use the identity value instead, fixing min and max reductions when constructed without explicit initial values. Note that we consider using the default constructor and not supplying an initial value to be deprecated.

  - Modifies/refactors reducer to take a Policy enum arg


MrBurmark:
I re-enabled some sycl and openmp target tests, we'll see if we have to disable them again.

rhornung67:
Unfortunately, we don't have an OpenMP target CI check at the moment. @rchen20 is working on build scripts for that. Then we can re-enable the CI check.

MrBurmark:
I discovered why the forall openmp target reduction tests were not created, they do not get correct answers. I disabled them again.

rhornung67:
> I discovered why the forall openmp target reduction tests were not created, they do not get correct answers. I disabled them again.

Was this all omp target reduction tests, or just the ones that used the "old-style" reductions (reduction object captured by value)?

docs/sphinx/user_guide/cook_book/multi-reduction.rst
```diff
@@ -74,13 +74,14 @@ number of blocks, optimized for performance with multi_reducers.::
...
-execution policy. For example, ``RAJA::seq_multi_reduce`` does a sequential multi-reduction
-and can only be used with sequential execution policies. The
-``RAJA::cuda_multi_reduce_atomic`` policy uses atomics and can only be used with
-cuda execution policies. Similarly for other RAJA execution back-ends, such as
-HIP and OpenMP. Here are example RAJA multi-reduction policies whose names are
-indicative of which execution policies they work with::
+The multi-reduction policy specifies how the multi-reduction is done and must
+support the execution policy. For example, ``RAJA::seq_multi_reduce`` does a
+sequential multi-reduction and can only be used with sequential execution
+policies. The ``RAJA::cuda_multi_reduce_atomic`` policy uses atomics and
```
artv3, line 80: The detail that cuda_reducers also supports, omp, and sequential can take a double look when first reading it. Should we just call it auto_*?

docs/sphinx/user_guide/cook_book/multi-reduction.rst
```diff
@@ -111,6 +113,25 @@ The results of these operations will yield the following values:
...
+or an OpenMP loop, construct or reset it with the matching runtime policy. For
+a sequential loop, use ``RAJA::Policy::sequential`` or
+``RAJA::policy_of<RAJA::seq_exec>::value``. For an OpenMP loop, use
+``RAJA::Policy::openmp`` or, for example,
+``RAJA::policy_of<RAJA::omp_parallel_for_exec>::value``. This tells RAJA not
+to prepare CUDA or HIP multi-reduction resources for that object.
+
+If the same multi-reducer object will later be used with CUDA or HIP loops,
+call ``reset(RAJA::Policy::cuda, ...)`` or ``reset(RAJA::Policy::hip, ...)``
+before that use. Use ``RAJA::Policy::undefined`` only when the object
```
artv3, line 131:
```suggestion
before that use. Use ``RAJA::Policy::auto`` only when the object
```
what about calling it auto?

artv3: I tend to associate undefined behavior as something that I generally don't want

include/RAJA/policy/PolicyBase.hpp
```diff
@@ -41,6 +42,123 @@ enum class Policy
   sycl
 };
 
+constexpr const char* get_policy_name(Policy p)
```
artv3, line 47: This could also be a std::string_view if you choose

artv3: and mark the function noexcept

include/RAJA/policy/PolicyBase.hpp
```diff
@@ -41,6 +42,123 @@ enum class Policy
...
+template<Policy... supported_policies>
+inline bool policy_supported_or_throw(const char* context_name,
+                                      PolicyList<supported_policies...> list,
+                                      Policy p)
+{
+  if (policy_supported(list, p))
+  {
+    return true;
+  }
+  std::string msg;
```
artv3, line 162: Don't we need to add in the string header on top?  Compilers can sometimes add it in for us though

test/include/RAJA_test-reducer-api.hpp
```diff
@@ -0,0 +1,61 @@
...
+template <typename PolicyList>
+struct ReducerApi;
+
+template <RAJA::Policy... Ps>
+struct ReducerApi<RAJA::PolicyList<Ps...>>
+{
+  template <typename REDUCER, typename... Args>
+  static REDUCER make(Args&&... args)
+  {
+    return REDUCER(Ps..., std::forward<Args>(args)...);
```
artv3, line 28: Do we need to cover the possibility of an empty parameter pack?

include/RAJA/policy/sycl/reduce.hpp
```diff
@@ -153,7 +175,13 @@ struct Reduce_Data
   //! transfers from the host to the device -- exit() is called upon failure
   RAJA_INLINE void hostToDevice(Offload_Info& RAJA_UNUSED_ARG(info))
   {
-    ::sycl::queue& q = currentResourceQueue();
```
artv3, line 156: Is this the only place where currentResourceQueue was used? it might be worth checking with the different SYCL backends to ensure the camp resourcequeue is used

docs/sphinx/user_guide/cook_book/multi-reduction.rst
MrBurmark: Well its not exactly auto if you have to tell it which Policy to support otherwise it allocates cuda memory.

docs/sphinx/user_guide/cook_book/multi-reduction.rst
MrBurmark: That was just keeping the existing naming. We may have places where we want the undefined vs the "auto" meaning, we'll have to think about that. Thinking about the name `all_supported` would be more specific than `auto` though `auto` is shorter and less awkward.

include/RAJA/policy/PolicyBase.hpp
MrBurmark: The unfortunate thing about string_view is that you can't assume that its null-terminated. I think some things are better left const char* for that reason, but it depends on how its used.

test/include/RAJA_test-reducer-api.hpp
MrBurmark: That's exactly what this is for. So we can optionally pass in policy or not, then all the other args.

include/RAJA/policy/sycl/reduce.hpp
MrBurmark: I went away from currentResourceQueue as it was never setup properly in places like the forall call so it was not better than just using get_default. The openmp target and sycl reducers are deprecated anyway, so I decided to just keep support for the default resource with them as that's all they supported before.

include/RAJA/policy/PolicyBase.hpp
MrBurmark: Added.

docs/sphinx/user_guide/cook_book/multi-reduction.rst
artv3: could `all_compat` work?

artv3: I also like: `support_all`

artv3: or inspired by raja, the `forall` policy

MrBurmark: I went with `all_supported` before I saw these suggestions.

artv3 (approved):
looks good to me!

rhornung67 (approved):

include/RAJA/pattern/detail/multi_reduce.hpp
```diff
@@ -87,7 +104,19 @@ struct BaseMultiReduce
...
                            value_type identity = MultiReduceOp::identity())
-      : data {container, identity}
+      : BaseMultiReduce {Policy::all_supported, container, identity}
+  {}
+
+  template<typename Container,
+           concepts::enable_if_t<
+               type_traits::is_range<Container>,
+               concepts::negate<std::is_convertible<Container, size_t>>,
+               concepts::negate<std::is_base_of<BaseMultiReduce, Container>>>* =
```
bechols97, line 114: (minor) Do we want this using C++20 concepts instead of the old `enable_if_t` style?

rhornung67: We are moving in  that direction. @johnbowen42 did a lot of transformation already, but I suspect that there is more to do. We can continue to push on  that after the release. IMO, this should not hold up the release.

bechols97 (approved):

rhornung67:
FYI, tuo is down for maintenance. We have the same CI checks on tioga and tuo. tioga checks are green. @MrBurmark to maintain momentum, we can merge this while tuo is down. It's your call.

MrBurmark:
> FYI, tuo is down for maintenance. We have the same CI checks on tioga and tuo. tioga checks are green. @MrBurmark to maintain momentum, we can merge this while tuo is down. It's your call.

@rhornung67  Go ahead and merge it.

rhornung67:
@MrBurmark you don't want the honors?

Issue #2067, opened 2026-07-29
Windows CI tests for Windows Visual Studio 2026


smithsg84:
The Windows CI tests are testing a single version of Windows and Visual Studio.  

CUDA 12 and 13 are better supported in Visual Studio 2022 and 2026 respectively so both should be included in tests.

This is related to #1960

smithsg84:
There are issues with moving to newer windows runners uncovered when attempting to the #2068 PR.

The use of cmake 3.29 was done due to a JSON parsing issue in the gtest module when cmake 4.4 was used.  The JSON parsing issue was intermittent.

When attempting to use cmake 3.29 with more recent window runners it was found:

- The windows-2022 runner is VS 2022 based.
- The windows-2025 updates the compiler to VS 2026.  
- cmake 3.29 does not support VS2026

So in #2068 the CI was set to use windows-2022 / VS 2022 until a more complete fix can be found for the cmake / JSON parsing issue.

Updating Window CI to VS 2026 will require resolving the observed JSON parsing issue.   Perhaps later cmake releases will address or an earlier release does not have the issue.

Believe the JSON parsing issue was only seen under windows.

smithsg84:
The Windows CI tests are overloading the runner nodes by 4X.   Parallel jobs in cmake is set to 16 and standard runners have 4 cores.

- Is 16 too large an impacting performance?  
- Is this impacting the observed bug?   The bug is non-deterministic so has some signs of a race condition.

PR #2068, opened 2026-07-29
Bugfix/smith84/specify windows cmake version

smithsg84:
- This PR is a bugfix, the windows CI jobs were failing intermittently.  Windows job was run multiple times with this patch and failure was not observed.   
  - Pin the cmake version to older 3,28,X due to issue with cmake 4.4 in the windows image
    - Previous change did not change the cmake version https://github.com/llnl/RAJA/commit/637cd1b6352f0132a2cf40a84d1a9f36c99e974f.  The variable did not change the cmake version.
  - Pin the windows image to Visual Studio 2022 
     - windows-latest was updated to Visual Studio 2026

Additional Windows testing is proposed in issue #2067 to add both Visual Studio 2022 and 2026 and CUDA.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,6 +62,7 @@ jobs:
       - uses: threeal/cmake-action@v1.3.0
         with:
           build-dir: build
+          generator: 'Visual Studio 17 2022'     # force VS generator so MSVC is used instead of NMake
           options:
             ENABLE_WARNINGS_AS_ERRORS=Off
             BLT_CXX_STD="c++20"
@@ -78,4 +79,4 @@ jobs:
 ## Run tests action
       - uses: threeal/ctest-action@v1.1.0
         with:
-          build-config: Debug
+          build-config: Release
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,7 @@ jobs:
         - args: BUILD_SHARED_LIBS=Off
           
 
-    runs-on: windows-2022
+    runs-on: windows-2025
     steps:
       - uses: actions/checkout@v2
         with:
@@ -54,7 +54,7 @@ jobs:
 
 ## Back off version of cmake with known issue
       - name: Install CMake 3.29
-        uses: lukka/get-cmake@latest
+        uses: lukka/get-cmake@e690607b2b7d42cf38fbb2ecbb6e568ebcd57291  # v4.4.0
         with:
           cmakeVersion: "~3.29.0"
 
@@ -63,7 +63,7 @@ jobs:
       - uses: threeal/cmake-action@v1.3.0
         with:
           build-dir: build
-          generator: 'Visual Studio 17 2022'     # force VS generator so MSVC is used instead of NMake
+          generator: 'Visual Studio 18 2026'     # force VS generator so MSVC is used instead of NMake
           options:
             ENABLE_WARNINGS_AS_ERRORS=Off
             BLT_CXX_STD="c++20"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -45,17 +45,25 @@ jobs:
...
+
+    runs-on: windows-2022
     steps:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
+
+## Back off version of cmake with known issue
+      - name: Install CMake 3.29
+        uses: lukka/get-cmake@latest
```
adrienbernede: I feel more comfortable with a pin-pointed version when using an external action.
Someone could argue that on the contrary "latest" ensures that we get the security patches if needed, but I tend to prefer a vetted version and no risk of surprise when latest changes.

I’ll leave this as a comment, not a required change.

adrienbernede (approved):

.github/workflows/build.yml
adrienbernede: Here is a tool that can help with the suggestion:
https://github.com/mheap/pin-github-action#how-it-works

.github/workflows/build.yml
```diff
@@ -45,17 +45,25 @@ jobs:
             CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=On
         - args: BUILD_SHARED_LIBS=Off
           
-    runs-on: windows-latest
+
+    runs-on: windows-2022
```
rhornung67: @smithsg84 I don't think we want to go back to an old version of Windows.

.github/workflows/build.yml
```diff
@@ -45,17 +45,25 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
+
+## Back off version of cmake with known issue
+      - name: Install CMake 3.29
+        uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: "~3.29.0"
```
bechols97, line 71: This might be a little outside this PR. However, did we also want to use this to specify the cmake version for the mac build as well to be consistent?

bechols97 (approved):

.github/workflows/build.yml
smithsg84: I think this is a good catch.   I think we should pin versions to avoid CI breaking due to upstream changes.

Related, Chris has mentioned there are possible security issues with using the tags.   Suggested change is to use commit hashes explicitly.    See discussion https://www.wiz.io/blog/github-actions-security-guide

.github/workflows/build.yml
rhornung67: I think we do because I recall seeing a similar issue with that.

.github/workflows/build.yml
smithsg84: I've pinned the get-cmake action.

rhornung67 (approved):

.github/workflows/build.yml
smithsg84: Moving to windows-2025 is being pushed back in order to get the CI running.

Documented updating issues in https://github.com/llnl/RAJA/issues/2067

.github/workflows/build.yml
smithsg84: Mac has been updated to same cmake 3.29 version.

adrienbernede:
@smithsg84 For the records, I used AI to audit the newly used lukka/get-cmake action.
It resulted in a security issue : https://github.com/lukka/get-cmake/issues/268,
and a PR for workflow security : https://github.com/lukka/get-cmake/pull/267
Both were addressed and merged. The next release will add sha verification for the downloaded CMake binaries.

PR #2078, opened 2026-08-11
Add a way to free internal memory

MrBurmark:
# Summary

Add a way to free memory stored in RAJA's internal memory pools. We do not free this memory as routines like `cudaFree` and `hipFree` can not be called after main.

- This PR is a feature
- It does the following:
  - Adds optional memory cleanup at the request of various people over the years

### DIFF
--- a/docs/sphinx/user_guide/memory.rst
+++ b/docs/sphinx/user_guide/memory.rst
@@ -22,9 +22,13 @@ Some RAJA backend operations, including reductions, scans, and sorts, may use
 temporary storage. RAJA keeps some of this storage for reuse in internal memory
 pools. Applications can call ``RAJA::release_unused_internal_memory()`` after
 outstanding RAJA and device work has completed to return unused internal memory
-to the backend allocator.
+to the backend allocator. Typical use is to call the function at application shutdown
+to prevent the internal memory pools from appearing in memory-leak reports. Frequent
+runtime calls are generally unnecessary because the pools are retained for reuse are
+typically not large.
 
 The function returns the number of bytes released and does not synchronize
 CUDA, HIP, or SYCL work. RAJA does not do this automatically during static
 destruction because CUDA and HIP runtime teardown can make late ``cudaFree`` or
-``hipFree`` calls fail.  As such, this function must only be called before the end of main to avoid racing with backend runtime cleanup.
+``hipFree`` calls fail. As such, this function must only be called before the
+end of main to avoid racing with backend runtime cleanup.
--- a/docs/sphinx/user_guide/memory.rst
+++ b/docs/sphinx/user_guide/memory.rst
@@ -27,4 +27,4 @@ to the backend allocator.
 The function returns the number of bytes released and does not synchronize
 CUDA, HIP, or SYCL work. RAJA does not do this automatically during static
 destruction because CUDA and HIP runtime teardown can make late ``cudaFree`` or
-``hipFree`` calls fail.
+``hipFree`` calls fail.  As such, this function must only be called before the end of main to avoid racing with backend runtime cleanup.
--- a/docs/sphinx/user_guide/memory.rst
+++ b/docs/sphinx/user_guide/memory.rst
@@ -24,7 +24,7 @@ pools. Applications can call ``RAJA::release_unused_internal_memory()`` after
 outstanding RAJA and device work has completed to return unused internal memory
 to the backend allocator. Typical use is to call the function at application shutdown
 to prevent the internal memory pools from appearing in memory-leak reports. Frequent
-runtime calls are generally unnecessary because the pools are retained for reuse are
+runtime calls are generally unnecessary because the pools retained for reuse are
 typically not large.
 
 The function returns the number of bytes released and does not synchronize
--- a/docs/sphinx/user_guide/memory.rst
+++ b/docs/sphinx/user_guide/memory.rst
@@ -1,0 +1,30 @@
+.. ##
+.. ## Copyright (c) Lawrence Livermore National Security, LLC and other
+.. ## RAJA Project Developers. See top-level LICENSE and COPYRIGHT
+.. ## files for dates and other details. No copyright assignment is required
+.. ## to contribute to RAJA.
+.. ##
+.. ## SPDX-License-Identifier: (BSD-3-Clause)
+.. ##
+
+.. _memory-label:
+
+******
+Memory
+******
+
+.. _memory-release-internal-pool-label:
+
+Releasing RAJA Internal Pool Memory
+===================================
+
+Some RAJA backend operations, including reductions, scans, and sorts, may use
+temporary storage. RAJA keeps some of this storage for reuse in internal memory
+pools. Applications can call ``RAJA::release_unused_internal_memory()`` after
+outstanding RAJA and device work has completed to return unused internal memory
+to the backend allocator.
+
+The function returns the number of bytes released and does not synchronize
+CUDA, HIP, or SYCL work. RAJA does not do this automatically during static
+destruction because CUDA and HIP runtime teardown can make late ``cudaFree`` or
+``hipFree`` calls fail.
--- a/include/RAJA/policy/sycl/MemUtils_SYCL.hpp
+++ b/include/RAJA/policy/sycl/MemUtils_SYCL.hpp
@@ -121,6 +121,15 @@ using device_zeroed_mempool_type =
     basic_mempool::MemPool<DeviceZeroedAllocator>;
 using pinned_mempool_type = basic_mempool::MemPool<PinnedAllocator>;
 
+inline size_t release_unused_internal_memory()
+{
+  size_t released = 0;
+  released += device_mempool_type::getInstance().release_unused();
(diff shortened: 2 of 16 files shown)


src/Memory.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+#if defined(RAJA_ENABLE_CUDA)
+  released += ::RAJA::cuda::release_unused_internal_memory();
+#endif
+#if defined(RAJA_ENABLE_HIP)
+  released += ::RAJA::hip::release_unused_internal_memory();
+#endif
+#if defined(RAJA_ENABLE_SYCL)
+  released += ::RAJA::sycl::release_unused_internal_memory();
+#endif
```
bechols97, line 49: Does this need a `release_unused_internal_memory` for OpenMP Target? Or do the OpenMP target policies already release internal memory?

MrBurmark: The openmp target RAJA backend allocates and frees as needed and doesn't use pools.

bechols97 (approved):

src/Memory.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+#if defined(RAJA_ENABLE_SYCL)
+#include "RAJA/policy/sycl/MemUtils_SYCL.hpp"
+#endif
+
+namespace RAJA
+{
+
+size_t release_unused_internal_memory()
```
adayton1, line 37: Does it really make sense to return the number of bytes released? Since it can consist of different memory types, I think it will be hard to interpret the return value.

test/unit/util/test-basic-mempool.cpp
```diff
@@ -0,0 +1,84 @@
...
+  ASSERT_GE(pool.release_unused(), test_arena_size);
+}
+
+TEST(BasicMemPool, release_unused_internal_memory_public_api)
+{
+  const size_t released = RAJA::release_unused_internal_memory();
+  ASSERT_EQ(released, size_t {0});
+
+#if defined(RAJA_CUDA_ACTIVE)
+  (void)RAJA::cuda::release_unused_internal_memory();
```
adayton1: Check the return value is 0?

src/Memory.cpp
MrBurmark: I agree its not very useful, but you can at least see that some bytes were released.

adayton1: I guess that's better than no information. And it makes testing easier.

adayton1 (approved):

include/RAJA/policy/sycl/MemUtils_SYCL.hpp
```diff
@@ -121,6 +121,15 @@ using device_zeroed_mempool_type =
     basic_mempool::MemPool<DeviceZeroedAllocator>;
 using pinned_mempool_type = basic_mempool::MemPool<PinnedAllocator>;
 
+inline size_t release_unused_internal_memory()
```
rhornung67, line 124: Why do the others use `RAJA_INLINE` macro and this one uses the C++ keyword `inline`?

test/unit/util/test-basic-mempool.cpp
rhornung67: I does seem that that would be a reasonable (perhaps the only) basic check here.

rhornung67 (approved):

docs/sphinx/user_guide/memory.rst
```diff
@@ -0,0 +1,30 @@
...
+Some RAJA backend operations, including reductions, scans, and sorts, may use
+temporary storage. RAJA keeps some of this storage for reuse in internal memory
+pools. Applications can call ``RAJA::release_unused_internal_memory()`` after
+outstanding RAJA and device work has completed to return unused internal memory
+to the backend allocator.
+
+The function returns the number of bytes released and does not synchronize
+CUDA, HIP, or SYCL work. RAJA does not do this automatically during static
+destruction because CUDA and HIP runtime teardown can make late ``cudaFree`` or
+``hipFree`` calls fail.
```
smithsg84: I believe the prime motivator for this is to aid with memory leak detection?   Clarify intent to make it explicit might be useful to add:

"Typical use is at application shutdown to prevent the internal memory pools from appearing in memory-leak reports. Frequent runtime calls are generally unnecessary because the pools are retained for reuse are typically not large."

smithsg84 (approved):

include/RAJA/policy/sycl/MemUtils_SYCL.hpp
MrBurmark: We are not entirely consistent about using `RAJA_INLINE`. `RAJA_INLINE` is sometimes always inline and that is more more inline than we probably want here.

docs/sphinx/user_guide/memory.rst
trws:
```suggestion
destruction because CUDA and HIP runtime teardown can make late ``cudaFree`` or
``hipFree`` calls fail.  As such, this function must only be called before the end of main to avoid racing with backend runtime cleanup.
```

docs/sphinx/user_guide/memory.rst
```diff
@@ -0,0 +1,30 @@
...
+Some RAJA backend operations, including reductions, scans, and sorts, may use
+temporary storage. RAJA keeps some of this storage for reuse in internal memory
+pools. Applications can call ``RAJA::release_unused_internal_memory()`` after
+outstanding RAJA and device work has completed to return unused internal memory
+to the backend allocator.
+
+The function returns the number of bytes released and does not synchronize
+CUDA, HIP, or SYCL work. RAJA does not do this automatically during static
+destruction because CUDA and HIP runtime teardown can make late ``cudaFree`` or
+``hipFree`` calls fail.  As such, this function must only be called before the end of main to avoid racing with backend runtime cleanup.
```
MrBurmark:
```suggestion
``hipFree`` calls fail. As such, this function must only be called before the
end of main to avoid racing with backend runtime cleanup.
```

docs/sphinx/user_guide/memory.rst
```diff
@@ -0,0 +1,30 @@
...
+.. _memory-release-internal-pool-label:
+
+Releasing RAJA Internal Pool Memory
+===================================
+
+Some RAJA backend operations, including reductions, scans, and sorts, may use
+temporary storage. RAJA keeps some of this storage for reuse in internal memory
+pools. Applications can call ``RAJA::release_unused_internal_memory()`` after
+outstanding RAJA and device work has completed to return unused internal memory
+to the backend allocator.
```
MrBurmark:
```suggestion
outstanding RAJA and device work has completed to return unused internal memory
to the backend allocator. Typical use is to call the function at application shutdown
to prevent the internal memory pools from appearing in memory-leak reports. Frequent
runtime calls are generally unnecessary because the pools are retained for reuse are
typically not large.
```

docs/sphinx/user_guide/memory.rst
MrBurmark: merged this into the docs.

src/Memory.cpp
MrBurmark: Keeping it for testing convenience. You could use the backend specific versions of the function to get more useful information.

docs/sphinx/user_guide/memory.rst
```diff
@@ -0,0 +1,34 @@
...
+Releasing RAJA Internal Pool Memory
+===================================
+
+Some RAJA backend operations, including reductions, scans, and sorts, may use
+temporary storage. RAJA keeps some of this storage for reuse in internal memory
+pools. Applications can call ``RAJA::release_unused_internal_memory()`` after
+outstanding RAJA and device work has completed to return unused internal memory
+to the backend allocator. Typical use is to call the function at application shutdown
+to prevent the internal memory pools from appearing in memory-leak reports. Frequent
+runtime calls are generally unnecessary because the pools are retained for reuse are
```
adayton1:
```suggestion
runtime calls are generally unnecessary because the pools retained for reuse are
```

test/unit/util/test-basic-mempool.cpp
```diff
@@ -0,0 +1,84 @@
...
+#endif
+#if defined(RAJA_HIP_ACTIVE)
+  (void)RAJA::hip::release_unused_internal_memory();
+#endif
+#if defined(RAJA_SYCL_ACTIVE)
+  (void)RAJA::sycl::release_unused_internal_memory();
+#endif
```
MrBurmark, line 83:
```suggestion
#if defined(RAJA_CUDA_ACTIVE)
  ASSERT_EQ(RAJA::cuda::release_unused_internal_memory(), size_t {0});
#endif
#if defined(RAJA_HIP_ACTIVE)
  ASSERT_EQ(RAJA::hip::release_unused_internal_memory(), size_t {0});
#endif
#if defined(RAJA_SYCL_ACTIVE)
  ASSERT_EQ(RAJA::sycl::release_unused_internal_memory(), size_t {0});
#endif
```

test/unit/util/test-basic-mempool.cpp
MrBurmark: now comparing to 0.

